PR #1, opened 2015-03-09
Add python-social-auth requirement

mbland:
cc: @seanherron


seanherron:
Removed in favor of using the Hub proxy.

PR #4, opened 2015-03-20
Dev

vzvenyach:
A number of changes made, including: 

1. Changes to the Projects Model.
2. Streamlining create/update for the timesheet, so that it's a single URL for both create and update.
3. Aesthetic changes to the timesheet template, and underlying CSS.


tock/hours/views.py
```diff
@@ -34,43 +36,17 @@ def get_context_data(self, **kwargs):
...
-            return self.render_to_response(self.get_context_data(form=form))
-
-class TimecardUpdateView(UpdateView):
+class TimecardView(UpdateView):
     form_class = TimecardForm
     template_name = 'hours/timecard_form.html'
 
     def get_object(self, queryset=None):
-        obj = Timecard.objects.get(reporting_period__start_date=self.kwargs['reporting_period'], user__id=self.request.user.id)
+        r = ReportingPeriod.objects.get(start_date=datetime.datetime.strptime(self.kwargs['reporting_period'], "%Y-%m-%d").date())
```
seanherron: I'm confused as to why we need to strip the timestamp in to a string and then re-encode it as a `date()`. Does referencing the reporting period on its own not work?

tock/projects/models.py
```diff
@@ -23,6 +23,8 @@ class Project(models.Model):
     name = models.CharField(max_length=200)
     iaa = models.CharField(max_length=200, blank=True)
     agency = models.ForeignKey(Agency)
+    description = models.TextField(blank=True, null=True)
+    billable = models.BooleanField(default=False)
```
seanherron: Shouldn't we be able to determine the billable status of a project via the existence (or non-existence) of `iaa`? Then we could eliminate the `billable` field.

seanherron:
Made a few inline comments, otherwise :+1:

seanherron:
Also, apart from changing the amount of extra fields, I didn't see any other style changes - am I missing something?

tock/hours/views.py
vzvenyach: This was strange to me too, to be honest. This is what I needed to do to get it to work?

tock/projects/models.py
vzvenyach: I suppose that's right. Perhaps I'm anticipating situations where it's billable through other means? But hard to know what that might be.

vzvenyach:
I guess this iteration is only the extra fields and adding the "delete" checkbox.

PR #15, opened 2015-03-23
db migrations & cf database

seanherron:


vzvenyach:
Can you change the travis build to be `python tock/manage.py test` and let it run again?

Issue #60, opened 2015-04-09
Change behavior on report view and on home page


vzvenyach:
If the hours for the timecard do not equal 100, then the report is not considered completed.

PR #63, opened 2015-04-10
adding better reporting period stats

seanherron:


seanherron:
This fixes https://github.com/18F/tock/issues/60

Issue #80, opened 2015-04-22
test efficacy of collecting hours instead of percentages


batemapf:
at an in-person meeting today, the following was decided:

1. to add provisional functionality to Tock that will allow users to report time on an hourly basis as opposed to by percentages (as is the current method), with number of hours allowed to be reported capped at 40 per five day week;

2. develop a hypothesis statement / experiment design to ensure test of provisional functionality described in 1. (above) results in actionable conclusion; and

3. launch provisional functionality within the next two weeks with launch accompanied by notice from RM / SR regarding provisional collection of hours instead of percentages.

Issue #18, opened 2015-03-24
Convert hours to percentages automatically


kaitlin:
allow the user to enter hours instead and calculate the percentage based on a 40 hour workweek automatically.

seanherron:
I was thinking of basically providing two `input` fields next to each other, like this:

`| HOUR | PERCENTAGE |`

You could enter in either a number of hours (which would then automatically fill in the relevant percentage) or fill in a percentage (which would then automatically fill in the number of hours). It may be a bit complicated looking though - any ideas for a simpler solution?

konklone:
I do think that would be a bit complicated looking. I think putting in hours is the more straightforward solution, both conceptually and UX-wise. But you could allow people who prefer entering %'s for some reason to toggle their input mode to that too, either with a user-wide default or on a per-form basis.

kaitlin:
Yeah I think I might just have a toggle button that switches between percentages and hours for all the fields.

seanherron:
This does bring up a tangentially related question - should we account for federal holidays/snow days/etc in reporting periods, or continue to have users add them in as `Out of Office`. For instance, if we have a federal holiday on a monday, should users put 20%/8 hours in "Out of Office", or should that be assumed on the backend and they would simply fill out for the remaining 4 days of the week?

jpyuda:
At the very least it would be nice to allow half-percentages. 1 hour is 2.5%.

esgoodman:
+1. A choice of hours as well as percentages would speed up timesheet completion considerably for those of us who work different hours on different projects each week.

vzvenyach:
Until direction from higher ups, wontfix.

seanherron:
I just pushed a trial for this that allows people (well, requires people) to report in hours instead of percentages! For example, here's @kaitlin's new timecard: https://hub.18f.us/tock/reports/2015-04-26/kaitlin.devine%40gsa.gov/

Stephanie is going to be sending out some guidance before the next reporting period is due around this.

PR #94, opened 2015-04-28
Allow input in hours instead of percentages

seanherron:
This PR adds a much requested feature (see https://github.com/18F/tock/issues/80 and https://github.com/18F/tock/issues/18) that switches Tock from recording time in percentages to recording time in hours. It also provides a utility to convert previously submitted timecards to the new standard.

Merging and deploying this is a multi step process:

1. After merge, this can be deployed, at which point all previous times will be converted to hours, and all future times will be recorded as hours
2. Once successful, a new PR will need to be made that removes the `time_percentage` field from the `TimecardObject` model and all related tests, as well as turn `hours_spent` in to a required field.


seanherron:
cc @vzvenyach

batemapf:
@seanherron @vzvenyach who wants to hit merge?  is the second PR teed up?

vzvenyach:
This looks pretty excellent! :100: @seanherron

PR #165, opened 2015-06-03
Stream CSV in bulk

### LINKED ISSUE #163: increase max_page_size to infinity
https://github.com/18F/tock/blob/master/tock/api/views.py#L21

^^looking for an easy way to dump all timecard data into a single CSV for offiline use. line above seems to be limiting line dump to 500.  @vzvenyach and i were thinking this is set up for something @shawnbot is doing with visualization.  maybe create a timecards_dump.csv/json endpoint that is a clone of the timecards.csv/json endpoint, minus the max_page_size limitation?

shawnbot:
This PR introduces a new API endpoint, `/api/timecards_bulk.csv`, that streams the data as CSV, and should resolve #163. I basically followed [these instructions](https://docs.djangoproject.com/en/1.8/howto/outputting-csv/#streaming-large-csv-files), and wrapped up the streaming logic in a rendering function that we can use later on.

I didn't go as far as to create a [DRF renderer](http://www.django-rest-framework.org/api-guide/renderers/) class because I couldn't figure out how to get access to the view's serializer from within the renderer's `render()` method, but this would be a good thing to do in the future if we wanted any of the other CSV views to stream. The downside of this approach is that you can't calculate the `Content-Length` HTTP header, which means that browsers can't display progress when downloading.

### DIFF
--- a/tock/api/renderers.py
+++ b/tock/api/renderers.py
@@ -30,8 +30,7 @@ def stream_csv(queryset, serializer):
     """
     rows = map(serializer.to_representation, queryset)
     fields = list(serializer.fields.keys())
-    response = StreamingHttpResponse(generate_csv(rows, fields), content_type='text/csv')
-    return response
+    return StreamingHttpResponse(generate_csv(rows, fields), content_type='text/csv')
 
 class Echo(object):
     """
--- a/tock/api/renderers.py
+++ b/tock/api/renderers.py
@@ -1,4 +1,7 @@
 from rest_framework_csv.renderers import CSVRenderer
+from django.http import StreamingHttpResponse
+
+import csv
 
 # This class is needed to extract the "results" list from paginated data.
 # see: https://github.com/mjumbewu/django-rest-framework-csv#pagination
@@ -16,3 +19,24 @@ class PaginatedCSVRenderer(CSVRenderer):
         if not self.headers and len(data) > 0:
             self.headers = data[0].keys()
         return super(PaginatedCSVRenderer, self).render(data, media_type, renderer_context)
+
+class Buffer(object):
+    """
+    A pseudo-buffer, see:
+    <https://docs.djangoproject.com/en/1.8/howto/outputting-csv/#streaming-large-csv-files>
+    """
+    def write(self, value):
+        return value
+
+def stream_csv(queryset, serializer):
+    """
+    Stream data as CSV, given an interable queryset and a DRF
+    Serializer instance with a .fields dict and .to_representation()
+    method.
+    """
+    rows = map(serializer.to_representation, queryset)
+    fieldnames = list(serializer.fields.keys())
+    writer = csv.DictWriter(Buffer(), fieldnames=fieldnames)
+    data = map(writer.writerow, rows)
+    response = StreamingHttpResponse(data, content_type='text/csv')
+    return response
--- a/tock/api/renderers.py
+++ b/tock/api/renderers.py
@@ -28,7 +28,7 @@ def stream_csv(queryset, serializer):
     Serializer instance with a .fields dict and .to_representation()
     method.
     """
-    rows = map(serializer.to_representation, queryset)
+    rows = map(serializer.to_representation, queryset.iterator())
     fields = list(serializer.fields.keys())
     return StreamingHttpResponse(generate_csv(rows, fields), content_type='text/csv')
 
--- a/tock/api/renderers.py
+++ b/tock/api/renderers.py
@@ -20,14 +20,6 @@ class PaginatedCSVRenderer(CSVRenderer):
             self.headers = data[0].keys()
         return super(PaginatedCSVRenderer, self).render(data, media_type, renderer_context)
 
-class Buffer(object):
-    """
-    A pseudo-buffer, see:
-    <https://docs.djangoproject.com/en/1.8/howto/outputting-csv/#streaming-large-csv-files>
-    """
-    def write(self, value):
-        return value
-
 def stream_csv(queryset, serializer):
     """
     Stream data as CSV, given an interable queryset and a DRF
@@ -35,8 +27,30 @@ def stream_csv(queryset, serializer):
     method.
     """
     rows = map(serializer.to_representation, queryset)
-    fieldnames = list(serializer.fields.keys())
-    writer = csv.DictWriter(Buffer(), fieldnames=fieldnames)
-    data = map(writer.writerow, rows)
-    response = StreamingHttpResponse(data, content_type='text/csv')
+    fields = list(serializer.fields.keys())
+    response = StreamingHttpResponse(generate_csv(rows, fields), content_type='text/csv')
     return response
+
+class Echo(object):
+    """
+    A pseudo-buffer, see:
+    <https://docs.djangoproject.com/en/1.8/howto/outputting-csv/#streaming-large-csv-files>
+    """
+    def write(self, value):
+        return value
+
+def generate_csv(rows, fields=None, **writer_options):
+    """
+    This generator yields text for each written row, and optionally
+    writes a header row. We're using DictWriter.writerow() instead of
+    writeheader() because the latter doesn't return a value.
+    """
+    buff = Echo()
+    if fields:
+        writer = csv.DictWriter(buff, fieldnames=fields, **writer_options)
+        header = dict((field, field) for field in fields)
+        yield writer.writerow(header)
+    else:
+        writer = csv.writer(buff, **writer_options)
+    for row in rows:
+        yield writer.writerow(row)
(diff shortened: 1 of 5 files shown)


batemapf:
@shawnbot want me to go ahead and merge? looks like all is well...  if you are looking for some other eyes on it, i know @vzvenyach has a soft place in his consulting heart for /tock

batemapf:
ideal fields would be:

Project Name
Project Code
Project Billable TRUE / FALSE
Employee Name ([NAME]@gsa.gov)
Period Begin Date
Period End Date	
Hours Logged

batemapf:
untitled

tock/api/views.py
```diff
@@ -4,44 +4,94 @@
...
+            'id',
+            'username',
+            'first_name',
+            'last_name',
+        )
 
 class TimecardSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = TimecardObject
+        fields = (
```
jmcarp: Do we need to list fields in `Meta` after declaring them as descriptors on `TimecardSerializer`? If the only reason to declare fields here is to add `hours_spent`, we could accomplish the same thing with less code by adding `hours_spent` as a field descriptor on the schema class.

tock/api/views.py
```diff
@@ -4,44 +4,94 @@
...
+            'modified_date',
+            'hours_spent',
+        )
+    project_name = serializers.CharField(source='project.name')
+    project_id = serializers.CharField(source='project.id')
+    billable = serializers.BooleanField(source='project.accounting_code.billable')
+    employee = serializers.StringRelatedField(source='timecard.user')
+    start_date = serializers.DateField(source='timecard.reporting_period.start_date')
+    end_date = serializers.DateField(source='timecard.reporting_period.end_date')
+    modified_date = serializers.DateTimeField(source='modified')
```
jmcarp: Looks like the `DateTimeField` constructor takes an optional `format` parameter, which gets passed to `datetime.strftime` during serialization. How about:

```python
modified_date = serializers.DateTimeField(source='modified', format='%Y-%m-%d %H:%M:%S')
```

Then I think we don't need to override `to_representation`.

tock/api/views.py
```diff
@@ -98,9 +148,19 @@ def map_row(row):
...
+      which the date falls within their reporting date range.
+
+    * if `user` (in either `first.last` or numeric id) is provided,
+      get rows for that user.
+
+    * if `project` is provided as a numeric id or name, get rows for
+      that project.
+    """
     if 'date' in params:
         reporting_date = params.get('date')
```
jmcarp: I know it's not part of the patch, but if we reach this line, we already know that the `"date"` key exists in `params`, so writing `params.get('date')` is exactly equivalent to `params['date']`.

tock/api/views.py
```diff
@@ -98,9 +148,19 @@ def map_row(row):
     return response
 
 def get_timecards(queryset, params={}):
```
jmcarp: Again, not part of your changeset, but using mutable values like `{}` as default arguments is dangerous in python. More at http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments.

tock/api/renderers.py
```diff
@@ -16,3 +21,38 @@ def render(self, data, media_type=None, renderer_context=None):
...
+
+def stream_csv(queryset, serializer):
+    """
+    Stream data as CSV, given an interable queryset and a DRF
+    Serializer instance with a .fields dict and .to_representation()
+    method.
+    """
+    rows = map(serializer.to_representation, queryset)
+    fields = list(serializer.fields.keys())
+    response = StreamingHttpResponse(generate_csv(rows, fields), content_type='text/csv')
```
jmcarp: No need to assign `response` to a variable--we can just `return StreamingHttpResponse(...)`.

tock/api/renderers.py
```diff
@@ -16,3 +21,38 @@ def render(self, data, media_type=None, renderer_context=None):
...
+    return response
+
+class Echo(object):
+    """
+    A pseudo-buffer, see:
+    <https://docs.djangoproject.com/en/1.8/howto/outputting-csv/#streaming-large-csv-files>
+    """
+    def write(self, value):
+        return value
```
jmcarp: This isn't a problem, but we don't *have* to return `value` here, right? AFAIK, the only purpose of this class is to present a file-like interface, but not actually to do any work. It's basically a hack to deal with the fact that the `csv` module only wants to work with file-like objects.

tock/api/tests.py
```diff
@@ -1,3 +1,120 @@
...
+            'end_date',
+            'modified_date',
+            'hours_spent',
+        ))
+        rows_read = 0
+        for row in rows:
+            self.assertEqual(set(row.keys()), expected_fields)
+            self.assertEqual(row['project_id'], '1')
+            rows_read += 1
+        self.assertTrue(rows_read > 0, 'no rows read, expecting 1 or more')
```
jmcarp: Looks good. Is it worth testing whether rows actually have the right contents?

tock/api/views.py
```diff
@@ -4,44 +4,94 @@
...
+            'billable',
+        )
     user = serializers.StringRelatedField(source='timecard.user')
     project_id = serializers.CharField(source='project.id')
     project_name = serializers.CharField(source='project.name')
     start_date = serializers.DateField(source='timecard.reporting_period.start_date')
     end_date = serializers.DateField(source='timecard.reporting_period.end_date')
     billable = serializers.BooleanField(source='project.accounting_code.billable')
+
+class BulkTimecardSerializer(serializers.ModelSerializer):
```
jmcarp: This is very similar to `TimecardSerializer`. Could these classes be merged, or share a base class?

jmcarp:
LGTM @shawnbot. I added some comments inline, all minor / stylistic.

tock/api/views.py
shawnbot: I think so. Most of the field descriptors refer to relation fields, and I *think* that if you don't specify them the serializer will expand all of them. This is particularly important for time cards, which have lots of relations, some several levels deep.

tock/api/renderers.py
shawnbot: Ah yes, this was an artifact of me twiddling response headers. I'll simplify it.

tock/api/renderers.py
shawnbot: I think it actually does have to return a value, otherwise the generator passed to the response won't write anything. It's worth a look, though.

tock/api/tests.py
shawnbot: Definitely :)

tock/api/views.py
shawnbot: I couldn't figure out how to make serializers share field descriptors. When I tried to extend TimecardSerializer and only specify differing fields, it threw errors. It's worth another look, though!

tock/api/views.py
shawnbot: Yes! This is exactly what I was looking for.

tock/api/views.py
shawnbot: Woah, I had no idea! Good call. I can just return the queryset if there are no params.

tock/api/views.py
shawnbot: :+1:

tock/api/views.py
jmcarp: So we need to list (most) fields twice, in `Meta` and on the serializer class, because `ModelSerializer` automatically serializes all model fields unless we tell it not to. The simpler approach would be to inherit from `Serializer` instead of `ModelSerializer`. Since we're already creating fields explicitly for all but one of the attributes, there isn't really any added boilerplate, and then we don't have to specify `Meta.fields` to get DRF not to serialize extra stuff.

tock/api/renderers.py
```diff
@@ -16,3 +21,38 @@ def render(self, data, media_type=None, renderer_context=None):
...
             self.headers = data[0].keys()
         return super(PaginatedCSVRenderer, self).render(data, media_type, renderer_context)
+
+def stream_csv(queryset, serializer):
+    """
+    Stream data as CSV, given an interable queryset and a DRF
+    Serializer instance with a .fields dict and .to_representation()
+    method.
+    """
+    rows = map(serializer.to_representation, queryset)
```
cmc333333: Django tends to cache results from querysets, which might blow out your system memory for huge result sets. One way to avoid that is to use `queryset.iterator()`. Kinda related - `DEBUG == True` makes django log a lot more about queries; if you ever run into trouble with memory during development, this might be why.

tock/api/views.py
cmc333333: @jmcarp good man

PR #170, opened 2015-06-04
Update timecard_form.html

batemapf:


jmcarp:
Done in https://github.com/18F/tock/pull/168.

PR #175, opened 2015-06-10
[WIP] Use Django 1.8, fix #173

### LINKED ISSUE #173: Newly added projects don't appear in the timecard selector
This means that the app needs to be redeployed whenever we add a new project. I tried to fix this in 7c967fa8d99c8fd4113300121f7f4ebb9fbc7ac0, but it looks like I didn't do this correctly. I *think* the way to do this is to define `ProjectChoiceField.choices` as a callable, like so:

```python
class ProjectChoiceField(forms.ChoiceField):
    widget = SelectWithData()
    choices = projects_as_choices
```

The [Django docs](https://docs.djangoproject.com/en/1.8/ref/models/fields/#django.db.models.Field.choices) suggest that this might not work:

> Note that choices can be any iterable object – not necessarily a list or tuple. This lets you construct choices dynamically. But if you find yourself hacking choices to be dynamic, you’re probably better off using a proper database table with a ForeignKey. choices is meant for static data that doesn’t change much, if ever.

@jmcarp or @cmc333333, any ideas?

shawnbot:
This PR should fix #173. The [Django 1.8 docs](https://docs.djangoproject.com/en/1.8/ref/forms/fields/#choicefield) for `django.forms.ChoiceField` claim that:

> If the [choices] argument is a callable, it is evaluated each time the field’s form is initialized.

So I've updated the `TimecardObjectForm` to look like this, and ditched the `ProjectChoiceField` class altogether:

```python
class TimecardObjectForm(forms.ModelForm):
    project = forms.ChoiceField(widget=SelectWithData(), choices=projects_as_choices)

    class Meta:
        model = TimecardObject
        fields = ['project', 'hours_spent']
```

This works for me. I've tested it:

1. by adding and removing projects while the app is running, then checking the timecard entry form to see if the list gets updated; and
1. by running the tests, naturally

and in both cases the issue appears to have been resolved. I could use a code review to make sure that this is a legitimate solution, though. @jmcarp, I'd love your brain on this if you have a moment.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-Django==1.7.6
+Django==1.8
 dj-database-url==0.3.0
 psycopg2==2.6
 model-mommy==1.2.3
--- a/tock/hours/forms.py
+++ b/tock/hours/forms.py
@@ -70,15 +70,8 @@ def projects_as_choices():
     return accounting_codes
 
 
-class ProjectChoiceField(forms.ChoiceField):
-    widget = SelectWithData()
-
-    def __init__(self, *args, **kwargs):
-        super(ProjectChoiceField, self).__init__(*args, **kwargs)
-        self.choices = projects_as_choices()
-
 class TimecardObjectForm(forms.ModelForm):
-    project = ProjectChoiceField()
+    project = forms.ChoiceField(widget=SelectWithData(), choices=projects_as_choices)
 
     class Meta:
         model = TimecardObject


jmcarp:
This seems like the right thing to do. The code is more straightforward than before, and it makes sense that a callable option would be called each time the form is instantiated. Plus, might as well upgrade to the latest Django.

Have you tried to write a failing test for this before applying the fix (with the test app running, add or remove a project, then assert that the project options have changed)?

jmcarp:
Actually, this made me realize that there don't seem to be any functional tests. Using webtest via django-webtest might be a good idea here--I can try that out and see if I can get the expected failure.

jmcarp:
Proposed tests in https://github.com/18F/tock/pull/176.

PR #192, opened 2015-06-26
Use DRF-compatible permission classes for auth logic.

jmcarp:
* Add a `PermissionMixin` to make Django generic views compatible with DRF-style permission classes
* Add `IsSuperUserOrSelf` permission class
* Apply said permission instead of overriding `dispatch`
* Update tests
* Delete unused code

### DIFF
--- a/tock/employees/views.py
+++ b/tock/employees/views.py
@@ -1,20 +1,12 @@
-from functools import wraps
-import csv
-import io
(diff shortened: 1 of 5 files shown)


ramirezg:
A 403 is much nicer than a `name 'PermissionDenied' is not defined` error :+1:

PR #197, opened 2015-07-06
Employee app cleanup

### LINKED ISSUE #194: Test Coverage and Refactoring: Employees App
- Make sure test cover enough code to refactor without unexpected issues. 
- Ensure PEP8 
- General simplification of code and adding doc strings.

ramirezg:
- Make sure test cover enough code to refactor without unexpected issues. 
- Ensure PEP8 
- General simplification of code and adding doc strings.  

[Resolves #194]

### DIFF
--- a/tock/employees/tests/test_views.py
+++ b/tock/employees/tests/test_views.py
@@ -1,30 +1,154 @@
-from unittest.mock import Mock, patch
-
-from django.conf import settings
 from django.test import TestCase, Client
 from django.contrib.auth import get_user_model
 from django.core.urlresolvers import reverse
 
-from employees.views import UserFormView
+import datetime
+from employees.views import parse_date, UserListView
+from employees.models import UserData
+
 
 class UserViewTests(TestCase):
 
     fixtures = ['tock/fixtures/dev_user.json']
-    
+
     def setUp(self):
-        self.regular_user = get_user_model().objects.create(username="regular.user")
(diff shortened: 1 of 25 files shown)


tock/employees/tests/test_views.py
```diff
@@ -1,30 +1,154 @@
...
+        """ Test that the parse date function returns datetime obj when
+        input is not NA """
+        date = 'NA'
+        self.assertEqual(parse_date(date=date), None)
+        date = "01/01/2013"
+        self.assertEqual(parse_date(date=date), datetime.datetime(2013, 1, 1))
+
+    def test_UserListViewPermissionforAdmin(self):
+        """ Ensure that UserListView works for admin """
+        c = Client(HTTP_X_FORWARDED_EMAIL='test.user@gsa.gov')
```
jmcarp: Prefer using WebTest over the built-in test app client (see views test for hours for examples). WebTest makes real requests against a WSGI app running in a background thread, which gives a more accurate representation of the behavior than calling views directly. Plus, we get much better API for working with forms and making assertions about responses. Instead of making brittle assertions about exact strings that need to be contained in the response, we can use BeautifulSoup-style selectors:
```python
links = response.html.find_all('a', href=reverse(...))
self.assertEqual(len(links)), 1)
```

ramirezg: @jmcarp k, I've moved the view tests to WebTest. The form submission feature is great!

PR #199, opened 2015-07-07
Start coverage before tests.

### LINKED ISSUE #198: Tests on models not covering code
Models are well tested; however, the tests aren't generating coverage.

jmcarp:
Django 1.7+ imports models before starting the testing process, so
models appear to never be tested. This patch uses a workaround described
in https://github.com/django-nose/django-nose/issues/180.

[Resolves #198]


ramirezg:
:+1: 76% coverage now.

PR #200, opened 2015-07-08
hours app cleanup and test coverage

### LINKED ISSUE #193: Test Coverage and Refactoring: Hours
- Make sure test cover enough code to refactor without unexpected issues. 
- Ensure PEP8 
- General simplification of code and adding doc strings.

ramirezg:
- Make sure test cover enough code to refactor without unexpected issues. 
- Ensure PEP8 
- General simplification of code and adding doc strings.  

[Resolves #193]

### DIFF
--- a/tock/hours/admin.py
+++ b/tock/hours/admin.py
@@ -12,9 +12,8 @@ class ReportingPeriodListFilter(admin.SimpleListFilter):
     parameter_name = 'reporting_period'
 
     def lookups(self, request, model_admin):
-        reportingperiods = set(
-            [p.start_date for p in ReportingPeriod.objects.all()])
-        return [(p, p) for p in reportingperiods]
+        data = ReportingPeriod.objects.distinct().values_list('start_date')
+        return [(p[0], p[0]) for p in data]
 
     def queryset(self, request, queryset):
         return queryset
--- a/tock/hours/admin.py
+++ b/tock/hours/admin.py
@@ -8,55 +8,53 @@ from .models import ReportingPeriod, Timecard, TimecardObject
 
 
 class ReportingPeriodListFilter(admin.SimpleListFilter):
-  title = 'Reporting Period'
-  parameter_name = 'reporting_period'
+    title = 'Reporting Period'
+    parameter_name = 'reporting_period'
 
-  def lookups(self, request, model_admin):
-    reportingperiods = set([p.start_date
-                            for p in ReportingPeriod.objects.all()])
-    return [(p, p) for p in reportingperiods]
+    def lookups(self, request, model_admin):
+        reportingperiods = set(
+            [p.start_date for p in ReportingPeriod.objects.all()])
+        return [(p, p) for p in reportingperiods]
 
-  def queryset(self, request, queryset):
-    return queryset
+    def queryset(self, request, queryset):
+        return queryset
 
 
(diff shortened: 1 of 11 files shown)


tock/hours/forms.py
```diff
@@ -101,7 +110,7 @@ def clean(self):
             # print(form)
             if form.cleaned_data:
                 # Easy way of telling if we have the right data
-                if form.cleaned_data.get('hours_spent') == None:
+                if not form.cleaned_data.get('hours_spent'):
```
jmcarp: Should we be checking for `None` or for falsy values here? If `hours_spent` is allowed to be zero, we should stick with the `is None` check, else this is good.

tock/hours/templatetags/has_submitted_timesheet.py
```diff
@@ -6,9 +6,11 @@
 
 @register.filter(name='has_submitted_timesheet')
 def has_submitted_timesheet(user, reporting_period):
-  if not Timecard.objects.filter(reporting_period=reporting_period,
-                                 time_spent__isnull=False,
-                                 user=user).exists():
-    return False
-  else:
-    return True
+    if not Timecard.objects.filter(
```
jmcarp: This can be simplified:
```python
return Timecard.objects.filter(
    reporting_period=reporting_period,
    time_spent__is_null=False,
    user=user,
).exists()
```

tock/hours/tests/test_forms.py
```diff
@@ -1,114 +1,123 @@
...
+        self.reporting_period = hours.models.ReportingPeriod.objects.create(
+            start_date=datetime.date(2015, 1, 1),
+            end_date=datetime.date(2015, 1, 7),
+            working_hours=40)
+        self.user = get_user_model().objects.get(id=1)
+        self.project_1 = projects.models.Project.objects.get(name="openFEC")
+        self.project_2 = projects.models.Project.objects.get(
+            name="Peace Corps")
+
+    def tearDown(self):
```
jmcarp: The base test class should wrap every test method in a transaction which is rolled back afterwards, so we shouldn't have to manually clear tables. I would hope this is already part of the Django testing harness, but if not, let's add it.

tock/hours/admin.py
```diff
@@ -8,55 +8,53 @@
...
-  parameter_name = 'reporting_period'
+    title = 'Reporting Period'
+    parameter_name = 'reporting_period'
 
-  def lookups(self, request, model_admin):
-    reportingperiods = set([p.start_date
-                            for p in ReportingPeriod.objects.all()])
-    return [(p, p) for p in reportingperiods]
+    def lookups(self, request, model_admin):
+        reportingperiods = set(
```
jmcarp: Nitpicky, but it's cheaper to get unique dates on `ReportingPeriod` from the database than to load all records, then filter in Python.

tock/hours/forms.py
ramirezg: I think that hours spent should not be blank or 0: neither of these values add useful information. I'll write some tests to make this more explicit.

jmcarp:
This all LGTM. @ramirezg, are you still working on anything here, or is this ready to merge?

ramirezg:
@jmcarp Yep, this and #200 are ready.

PR #208, opened 2015-07-16
Add aggregate billing views.

jmcarp:
* Add billable vs nonbillable hours by quarter.
* Add billable vs nonbillable hours by quarter by user.

Caveat: These queries are hard to express using the Django ORM, so I
implemented them using raw SQL. We may want to translate to the ORM.
Once that decision is made, I'll add unit tests.

@ramirezg: Thoughts appreciated--do you know if these queries can be expressed in a performant way using the ORM? Reservations about using plain SQL?
@mbland: Would these aggregates be useful? I'm happy to add more as needed.


mbland:
@jmcarp I think so, but can you post some sample output (with names scrubbed out, if need be)? Having a hard time imagining what it'll look like from the code alone.

Also, is there a way to parse out a time frame smaller or larger than a quarter? If not, don't sweat it.

jmcarp:
@mbland: output would look something like this:
```
[
{
  "username": "joshcarp",
  "year": 2015,
  "quarter": 2,
  "billable": 300,
  "nonbillable": 50,
  "total": 350,
},
{
  "username": "joshcarp",
  "year": 2015,
  "quarter": 3,
  "billable": 100,
  "nonbillable": 0,
  "total": 100,
},
...
]
```
It would be straightforward to aggregate by reporting period or month instead of, or in addition to, quarter.

mbland:
I'd say go ahead with it. I'd also appreciate it if you merged 18F/team-api#18 for me. ;-)

tock/api/views.py
```diff
@@ -197,3 +200,115 @@ def bulk_timecard_list(request):
...
     return stream_csv(queryset, serializer)
+
+
+from rest_framework.response import Response
+from rest_framework.decorators import api_view
+
+hours_by_quarter_query = '''
+with agg as (
+    select
+        extract(year from start_date) as year,
```
ramirezg: Can't imagine doing this faster with the Django ORM. 
I think the year is slightly off for the gov calendar. [October, November, December](https://github.com/18F/tock/blob/master/tock/hours/models.py#L32) appear to be in the following year, otherwise this ready to merge.

jmcarp:
Now updated with unit tests, and using `factory-boy` for test fixtures. Ready for review @ramirezg.

ramirezg:
:+1:

PR #210, opened 2015-07-17
[WIP] Vary billing

jmcarp:
This patch allows each project to change active status and accounting code for each reporting period. This allows projects to change from nonbillable to billable and back, while preserving the history of accounting codes over time. Projects can also be marked as active or inactive for any given reporting period. The idea is to obviate the spreadsheet that @batemapf is currently using to store this information.

As implemented, creating a new reporting period automatically copies the reporting code and active status of each project to the new period, and newly created projects default to active, with a configurable default reporting code. If the overall proposal makes sense to @batemapf and @ramirezg, then I'll flesh this out with the following:
- [ ] Unit tests
- [ ] Migration from old data model to new
- [ ] Better admin interface for updating reporting codes over time
- [ ] Hiding inactive projects from timecard interface

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ install:
 
 script:
   - cd tock
+  - python manage.py makemigrations --noinput --settings=tock.settings.test
   - python manage.py migrate --noinput --settings=tock.settings.test
   - python manage.py test --noinput --settings=tock.settings.test
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ install:
 
 script:
   - cd tock
-  - python manage.py syncdb --noinput --settings=tock.settings.test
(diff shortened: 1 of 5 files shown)


batemapf:
:+1: 

i assume ops staff will be able to access the fruits of this patch through the django control panel?

jmcarp:
@batemapf that's the idea.

ramirezg:
:+: This looks great.

batemapf:
@allalala @adelevie would love to micropurchase an update to this PR (it is very stale).  thoughts?

adelevie:
@batemapf, in what ways is it stale?

batemapf:
well i am being told there are conflicts that need to be resolved before merge, and we have done a good amount of work to the data models since july, so i simply figured it was in need of a refresh.  of course, i'd be happy to not pay for said refresh :-)

PR #217, opened 2015-07-22
Add agency name to timecards API endpoints.

jmcarp:
As requested by @batemapf.


mbland:
Hey @jmcarp, so this adds the agency behind the billing code to the bulk timecard data?

jmcarp:
Yep, this was a quick addition that @batemapf asked about in #tock. I don't know the context here--is there more to do on this?

mbland:
Don't think so! We can aggregate each timecard item and see how much has been billed to each source, throw another graph up on the financials page. Works for a MVP.

PR #219, opened 2015-07-30
Map application to 18f.gov.

jmcarp:
Since all CF requests are now redirected to HTTPS, we need to use a URL
that handles SSL. This patch maps the Tock app to tock-app.18f.gov; the
"-app" suffix is used to distinguish the route used by nginx
(tock.18f.gov) from the route used by the app.

Ping @ramirezg. We'll also need to update the staging app to live at tock-app-staging.18f.gov. See https://github.com/18F/hub/pull/358 for full configuration.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -25,10 +25,10 @@ deploy()
 	cf push $next_deployment -n $app_name-$timestamp
 	echo "Mapping $next_deployment to the Main Domain"
 	cf map-route $next_deployment 18f.gov -n $app_name
-	cf map-route $next_deployment cf.18f.us -n $app_name
+  cf map-route $next_deployment 18f.gov -n $app_name-app
 	echo "Removing $current_deployment From the Main Domain"
 	cf unmap-route $current_deployment 18f.gov -n $app_name
-	cf unmap-route $current_deployment cf.18f.us -n $app_name
+  cf unmap-route $current_deployment 18f.gov -n $app_name-app
 	read -p "Check your app. Is it functioning properly? (y/n)" -n 1 -r
 	echo    # (optional) move to a new line
 	if [[ $REPLY =~ ^[Yy]$ ]]; then


ramirezg:
@jmcarp https://tock-staging.18f.gov/ is up.

konklone:
@jmcarp @ramirezg So, `https://tock.18f.gov` is what most people have been using for long while (not the old `cf.18f.us` URL), and already had HTTPS configured properly.

@NoahKunin and I both experienced visiting `tock.18f.gov`, submitting our time cards, and having it get redirected to a 403 Forbidden error at `https://tock-app.18f.gov`. The time card did successfully get saved, so the redirect occurs after a successful operation.

Visiting https://tock-app.18f.gov/ directly also gives a 403.

jmcarp:
@konklone: Tock is deployed to Cloud Foundry but accessed through the auth proxy that lives in the Hub. Previously, the Hub nginx server was proxying requests to `http://tock.cf.18f.us`, which stopped working last week when we started redirecting to HTTPS for all CF apps. This patch deploys the internal Tock app to a URL at `18f.gov` to get HTTPS, but we can't use `tock.18f.gov`, since that's already in use in the Hub config.

Tock can't be accessed via Cloud Foundry directly (unless the user is already authenticated), since authentication happens via the auth proxy, so a 403 when hitting the CF deployment directly is the expected behavior. But getting a redirect to tock-app is a bug. Will take a look. cc @ramirezg

mbland:
Resolved. Hub PR coming soon; will link to this PR.

konklone:
Got it! Thank you for the explanation, and to @mbland for the fix.

> Previously, the Hub nginx server was proxying requests to http://tock.cf.18f.us, which stopped working last week when we started redirecting to HTTPS for all CF apps.

Yikes -- I'm glad we fixed that, bumpy as it was.

PR #224, opened 2015-09-22
Update reporting_period_list.html

batemapf:
simply deleting piece of markup that includes the past five reporting periods.  goal is to prevent users from seeing this option on the main tock page.


ramirezg:
I can see how this would be clearer.

PR #234, opened 2015-10-20
flat-rate-option

batemapf:
add flat rate option to accounting codes.

per alan d:

1. do migrations need to run for this to take effect?
2. how can we get this field to show up in the admin panel, in the 'modify accounting code' box?

### DIFF
--- a/tock/projects/migrations/0002_auto_20151020_2043.py
+++ b/tock/projects/migrations/0002_auto_20151020_2043.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('projects', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='accountingcode',
+            name='flat_rate',
+            field=models.BooleanField(default=False),
+        ),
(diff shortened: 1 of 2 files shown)


ramirezg:
@batemapf @adelevie so I made the migrations using `python manage.py makemigrations`.

It may be the case that we'll have to edit the admin.py files to get the new fields to show up how we want them to show up. I can take a look later and add some tests. 

Also, did we add `Don't make crappy names!` at some point? Looks like the migration hadn't been created.

batemapf:
no idea about  the 'Don't make crappy names!'

On Tue, Oct 20, 2015 at 4:50 PM, Gabriel Ramirez <notifications@github.com>
wrote:

> @batemapf <https://github.com/batemapf> @adelevie
> <https://github.com/adelevie> so I made the migrations using python
> manage.py makemigrations.
>
> It may be the case that we'll have to edit the admin.py files to get the
> new fields to show up how we want them to show up. I can take a look later
> and add some tests.
>
> Also, did we add Don't make crappy names! at some point? Looks like the
> migration hadn't been created.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/18F/tock/pull/234#issuecomment-149698001>.
>



-- 
Patrick F. Bateman
Operations Director (acting)
18F
U.S. General Services Administration

batemapf:
@ramirezg this OK to merge?

PR #235, opened 2015-10-23
api-flat-rate

batemapf:
adding the flat rate indicator to the timecard and bulktimecard api views

cc @ramirezg @shawnbot


batemapf:
thanks @shawnbot!

PR #247, opened 2015-11-04
Added project detail in API

vzvenyach:
One of the features that I'm hoping to eventually use is the ability
to connect Tock data with other data. Given that Tock stores some
descriptive metadata, I wanted to be able to access that data at the project
level. In other words, this PR would expose the following endpoint:

`/api/projects/:number.json`

### DIFF
--- a/tock/api/urls.py
+++ b/tock/api/urls.py
@@ -9,6 +9,7 @@ from api import views
 
 urlpatterns = patterns('',
     url(r'^projects.(?P<format>csv|json)$', views.ProjectList.as_view(), name='ProjectList'),
+    url(r'^projects/(?P<pk>\d+).json$',view=views.ProjectInstanceView.as_view(),name='ProjectInstanceView'),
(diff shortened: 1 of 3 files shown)


geramirez:
@vzvenyach is this ready for review?

vzvenyach:
Yep. That'd be great @geramirez!

tock/api/urls.py
```diff
@@ -9,6 +9,7 @@
 
 urlpatterns = patterns('',
     url(r'^projects.(?P<format>csv|json)$', views.ProjectList.as_view(), name='ProjectList'),
+    url(r'^projects/(?P<pk>\d+).json$',view=views.ProjectInstanceView.as_view(),name='ProjectInstanceView'),
```
geramirez: Everything looks good :+1:  I was wondering, though, we wanted to force the `.json` here. We could remove it to make it default to JSON output. @vzvenyach

PR #249, opened 2015-11-05
Current employee

geramirez:
Added a new field the UserData model which indicates if a user is currently employed. 
Filtered the reporting period reports to only show users marked as currently employed. 

This should make it significantly easier to determine which employees need to fill out their time cards and the first step in creating an automated system for getting users to fill out their time cards.


batemapf:
this looks good to me @geramirez, going to merge.  mind throwing it up on tock-staging to take a look before we go into prod?

geramirez:
@batemapf wasn't actually done with this one. Sorry, should have made it clearer.

batemapf:
whoops! sorry about that, get excited when i see PRs!

PR #254, opened 2015-11-12
Reporting Period Audit

geramirez:
Adding endpoint that lists users who have not completed their timecards for as given period.

The bot for consuming this api is [here](https://github.com/geramirez/tock-bot)


batemapf:
@geramirez merging!

PR #256, opened 2015-11-18
updates to prod_user for local dev purposes

batemapf:
updated `username:` attribute so local dev is automatically a superuser and can access admin panel


batemapf:
@geramirez we worked on this together earlier today.  if good to merge, go for it!

PR #274, opened 2015-12-18
Closes #236:  Update README with missing setup step

### LINKED ISSUE #236: Admin Interface username/password?
Hi, I am wondering how I can access the admin interface? It asks for a username/password but the README doesn't mention one. I followed [these](http://www.djangobook.com/en/2.0/chapter06.html) standard directions on how to create an Django "superuser" account, however when I enter the account credentials in the admin page, it does not accept it. (I checked to make sure I typed the password in correctly) What am I doing wrong?

ccostino:
This changeset addresses a missing step in the README when setting up Tock locally.  The fixture data for a user and for the projects needs to be loaded in order to access the admin site and see any project information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ This will provision an entire setup for you pretty quickly (see `provision/dev/b
 $ vagrant ssh
 $ cd /vagrant/tock
 $ pyenv activate tock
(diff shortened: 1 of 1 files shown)


ccostino:
cc: @batemapf

Issue #250, opened 2015-11-05
Add notes field to hours


cmc333333:
Sometimes the categories in Tock aren't fine grained enough. It'd be handy to have a separate notes field to document a bit more thoroughly. I would _not_ expect this to be per pay period, instead per line entry within a pay period.

Example usage:
* 6 hrs on talent, "4 hours of prep, 2 of actual interviews"
* 20 hrs on projectA, "Mostly usability studies, and demos"

To be clear, I'd expect this to be an optional field

ccostino:
This is actually addressed with #267; note (ha!) that I added support to enable this for any time entry, not just `General` time entries. :-)

PR #277, opened 2015-12-21
Closes #267:  Add support for entry notes (snippets)

### LINKED ISSUE #267: add in description for `general` time`
is the ability to add notes to a line item in Tock a desirable feature? I always wish that I could add in notes about where specifically I spent my time in each category, particularly in General.

cc @theresaanna

ccostino:
This changeset introduces the ability to add notes to a time entry. These notes can be optional or required and applied to any project. In order to enable them, go into the Django admin and find a Project that you would like to enable notes on. In the admin form you will now see to additional checkboxes to enable the display of a notes field, and wether or not it is required with time entry.

A new dependency was added to the project to support this enhancement, called bleach (https://bleach.readthedocs.org/). This was added per the Django documentation note on using the `striptags` template filter: https://docs.djangoproject.com/en/1.9/ref/templates/builtins/#striptags

A couple of tests have been added and a couple of others have been updated to support this new enhancement.  In addition, the API output has been updated to also include any notes entered by users and several Django admin features have been enabled to make it easier to work in.


ccostino:
cc @batemapf and @kaitlin

ccostino:
There's still a bug with adding multiple line items to a time card that includes notes fields, so I'll be taking a look and fixing.

ccostino:
Issue should be fixed!  Any previous value entered in the notes field of the line item being cloned was not being cleared.

ccostino:
This PR would also address #250.

PR #281, opened 2015-12-28
Update inefficient queries.

jmcarp:
Use `select_related` and `prefetch_related` as appropriate to avoid n+1
queries in a handful of views. Use the `nplusone` query profiler in test
mode to detect similar inefficient queries in the future.

Note: I'm the maintainer of `nplusone`--would be interested in any feedback folks might have about the library, bug reports, etc.

Further reading: http://use-the-index-luke.com/sql/join/nested-loops-join-n1-problem


ccostino:
Big +1 on this.  /cc @batemapf

Thanks for sharing `nplusone` too, @jmcarp, that is awesome!  It would have been extraordinarily helpful for a Django app I was working on several years ago.  I will be sure to keep it in mind for the future, especially if I find myself doing report generation across several models again. :-)

jmcarp:
I saw that a few minor merge conflicts came up after I submitted this--they're resolved now.
cc @ccostino @batemapf

requirements-dev.txt
```diff
@@ -1,6 +1 @@
-django-debug-toolbar
-django-webtest
-factory-boy
-coverage
-ipython
-ipdb
```
ccostino: I'm not opposed to leaving IPython and ipdb out of the dev dependencies (though I find them helpful!), I'm just curious if you had a specific reason for omitting them.

tock/hours/forms.py
```diff
@@ -136,7 +137,7 @@ def clean_project(self):
         return data
 
     def clean(self):
-        cleaned_data = super(TimecardObjectForm, self).clean()
+        super(TimecardObjectForm, self).clean()
```
ccostino: D'oh!  Should've caught that with my previous PR, thanks!

ccostino:
Awesome, thank you @jmcarp!  I just took a look and everything appears fine to me, just a couple of notes.  @batemapf, this would be great to have in place!

PR #294, opened 2016-01-15
add description to project list

batemapf:
adds a description to the /projects/ page, along side the name and ID of the project  cc @lanebecker


ccostino:
Thanks @batemapf!

PR #298, opened 2016-01-22
Clarify Tock Projects == Billable

thecapacity:
It seemed from the slack discussion project == billable, so I thought
it worth making explicit and obvious.

### DIFF
--- a/tock/tock/templates/projects/project_detail.html
+++ b/tock/tock/templates/projects/project_detail.html
@@ -2,7 +2,7 @@
 {% load project_tags %}
 
 {% block content %}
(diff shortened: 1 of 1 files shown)


thecapacity:
@batemapf - I have no idea if I did this correctly...

batemapf:
the projects in the tock list are both billable and non billable. if you think users would like to to filter by billable / non-billable, or have some other indication, please file an issue to this effect. as a head's up, we used to have a column on this page that had a yes/no regarding billability for each project.  we removed it last year to reduce the focus on the billable vs. nonbillable.

going to close this PR for now.

PR #306, opened 2016-02-02
changes for 18F project

markcreyn:
We believe that the superuser 18F task is complete.

### DIFF
--- a/tock/hours/forms.py
+++ b/tock/hours/forms.py
@@ -3,6 +3,7 @@ import bleach
 from django import forms
 from django.forms.models import BaseInlineFormSet
 from django.forms.models import inlineformset_factory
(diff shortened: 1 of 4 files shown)


adelevie:
Hi @markcreyn. Thanks for the contribution. The pull request is currently causing the [test suite to fail](https://travis-ci.org/18F/tock/builds/106570984#L276). In addition the code appears to not fully implement the requirements in [the acceptance criteria](https://micropurchase.18f.gov/auctions/10):

> A superuser, when creating or editing a reporting period in the admin panel, can select users who may enter between 0 and 60 hours for that reporting period.

I will run the code locally and confirm this last point.

Regardless, we cannot accept the requirement as-is. However, you have until 5PM EST to fix these issues.

Issue #286, opened 2016-01-08
Gemnasium is reporting 5 security vulnerabilities for Django


monfresh:
https://gemnasium.com/18F/tock

monfresh:
Solution: upgrade to the latest version of Django.

batemapf:
@monfresh latest non-development release of django is 1.8, which is what is tock is now using. thoughts on upgrading to the in-development version (1.9)?

cmc333333:
@batemapf 1.9 is good for production, but I think the more pressing problem is that you're locked at 1.8 -- you should really be using the latest point release in that line (1.8.9)

jgrevich:
Can this issue be closed with #312? It also looks like this duplicates #264

Issue #264, opened 2015-12-04
Gemnasium is reporting several security alerts for Django


monfresh:
https://gemnasium.com/18F/tock/alerts

monfresh:
They can be resolved by upgrading Django to version 1.9

PR #312, opened 2016-02-16
update django ver

batemapf:
re: #264 and #286 updating to django 1.8.9 cc @monfresh @cmc333333

tested locally, passes all tests.


monfresh:
LGTM :shipit:

PR #319, opened 2016-02-28
Include Client Name in Project API Results

cwarden:
Include the client name (accounting code and agency) with each project
when retrieving a project list using the Tock API to allow
reconciliation with Float.  See https://micropurchase.18f.gov/auctions/11

*Since the data reconciliation tool is a new application written in a different language than Tock, I put it in a [separate repo](https://github.com/cwarden/onena).  If you would prefer that I include the source within the Tock repo, or pull it in as submodule, let me know.*

### DIFF
--- a/data-reconciliation.md
+++ b/data-reconciliation.md
@@ -1,0 +1,16 @@
+# Data Reconciliation
+
+18F uses Tock to track past staff project allocations and uses
+[Float](https://www.float.com/) to predict future staff project allocations.
+Both systems contain very similar information but with very different temporal
+aspects: Tock contains historical data, while Float contains projected future
+data. However, the data elements are very similar.
+
+Since both applications are used for managing the business of 18F, it is
+critical that the common data elements between the two are regularly
+reconciled. For instance, analysis quickly breaks down if a project in Float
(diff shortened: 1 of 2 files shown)


adelevie:
Thanks, @cwarden. We're taking a look at this and `onena`. We'll get back to you shortly if there are any remaining issues.

cwarden:
Thanks, @adelevie.  I'm not sure how you authenticate against the production Tock API endpoint since the development instance doesn't require authentication.  Do you manually copy the cookie set by oauth2_proxy  (nee Google Auth Proxy)?  That's what I assumed in [`tock_request(url)`](https://github.com/cwarden/onena/blob/dd78600a69dfb7a38562fa8eb029571c23bdad95/lib/onena/client.rb#L110-L112)

adelevie:
@cwarden, yep that's currently how we authenticate. We could have been a bit clearer about that. Nevertheless, after testing this against some test data and some live data, we accept the requirements that you delivered in this PR and in the github.com/cwarden/onena repository. I'll contact you offline to discuss payments. 

Great work (and thanks for the early delivery)! :ship:

Issue #308, opened 2016-02-04
Micropurchase candidate: Load project/line items from previous week when opening a current week's timesheet


porta-antiporta:
From week to week, chances are the projects a user is logging to changes very little.
However, currently when a new timesheet is opened, it starts out blank.
Users have to add in each line item again; and this often leads to people asking for a search function (search for "search" in #tock).  Search can be a pretty complex feature to develop.

If we come back to the original assumption: week to week projects a user logs time to changes very little; a natural enhancement is for the current timesheet to default with the projects from last week's timesheet when opened for the first time (rather than being blank).

*_Acceptance Criteria:_*

* When the user opens a **new** timesheet the timesheet is pre-populated with the previous week's projects.
 * Only the projects are carried over, the hours and snippets (for the General project) are blank.
* When the user opens a **saved** timesheet the timesheet does not pre-populate with the previous week's projects (i.e. the saved timesheet is not overwritten with last week's projects when re-opened)
*  A user can remove line items from an **unsubmitted** (i.e a new or saved) timesheet.  (One may argue scope creep here, but without the ability to delete a line item, this feature becomes unusable).
*  A user can add line items to an **unsubmitted** timesheet.  (This is current functionality that should be preserved).

batemapf:
my concerns are over "autopilot".  do we run the risk of folks just submitting the same time over and over again without thinking critically about it? this could be tested, of course.  would like to hear others thoughts.  @ric7694 you've been spending a lot of time with the tock data, would like to hear from you as well.

porta-antiporta:
I should clarify, it just imports the same projects.  But the *hours* are left blank.  I believe tock currently throws an error if there's a project line item with blank hours.

porta-antiporta:
ping

Is there potential for this to become a micropurchase candidate?  Or should I close the issue?

NoahKunin:
@porta-antiporta I'm thinking it's a perfect candidate for micro-purchase +1!

batemapf:
let's give it a whirl.  @allalala @adelevie please add this one to the queue.

porta-antiporta:
@allalala @adelevie in case micropurchases need a business case attached.

Back of napkin ROI (using survey data with [6 responses](https://docs.google.com/a/gsa.gov/forms/d/1Hc9IBOidYlzmJwel__itY1f_HGgbxTNDLWpWsHVLSIw/viewanalytics))
- 6 out of 6 prefer preloaded timesheets
[...]

PR #323, opened 2016-03-03
Pre-populate projects for new timecards + delete functionality

brendansudol:
This PR addresses [issue #308](https://github.com/18F/tock/issues/308) and micro-purchase [auction #15](https://micropurchase.18f.gov/auctions/15) to "Reload Tock Projects in new reporting period". 

When a user opens a reporting period timesheet that has no entries, it attempts to pre-populate the form with the projects from the last submitted timecard (if there is one). While the projects are carried over, the hours are left blank, and project entries can now be deleted for unsubmitted timecards.

One thing to note: in the spec, it states that the project pre-population should occur when a new timesheet is opened. However, this could lead to a situation where you open a new timesheet to see the projects from the last timecard (as desired), but if you then reload the page the projects would go away (because the timesheet is no longer "new" as it was created on the last request. This could perhaps be confusing to the user and for that reason, I implemented it such that the project pre-population occurs for a timesheet as long as there are no saved entries for that timesheet. This way, if you reload the page after opening a new timesheet, the projects will persist (and if you then add hours and save any of those projects, the pre-population will no longer occur and instead the saved entries will be loaded). Let me know what you think about this solution. It's a very trivial change if we want to just show the pre-populated projects only once on the initial creation.

New unit tests for this functionality can be found in `tock/hours/tests/test_views.py`.

All new code is [PEP 8](https://www.python.org/dev/peps/pep-0008/) compliant.


brendansudol:
Hi @adelevie -- just wanted to check in re: this PR. Let me know if there's any additional info I can provide.

adelevie:
Hi @brendansudol,

I was drafting this very message when you just asked that.

We accept the delivery. Thanks again for participating in micro-purchase. I'll contact you offline about payment.

brendansudol:
Haha -- I guess my :ear:s were burning :smile: Thanks for organizing!

adelevie:
Needs a rebase, @brendansudol :)

brendansudol:
survived the rebase with minimal scarring @adelevie -- let's get this puppy merged!

adelevie:
@batemapf, this awaits your approval.

brendansudol:
alright, we're ready again @adelevie and @batemapf -- it's go time!

batemapf:
woohoo! thanks @brendansudol!

porta-antiporta:
:tada:

Issue #320, opened 2016-03-01
New Feature: Total Hours Billed on Reports Screen


joshuabailes:
On the Tock Projects pages, we can see hours for 5 columns. It would be a helpful feature to also see the total number of hours billed to that tock project to date on this page as well.

So, for instance on https://tock.18f.gov/projects/195/ being able to see the total hours billed all time to date would be helpful in keeping up with labor burn rates, etc.

batemapf:
closed via #331

PR #331, opened 2016-03-10
total hours billed

dsiddy:
This PR addresses #320 by adding and displaying a new `total_hours` context variable *via* the `ProjectView` class and the `project_detail.html` template, respectively.


batemapf:
@dsiddy thanks for the contribution!

Issue #317, opened 2016-02-25
Tiny change: Change "Save" button label to "Save, but not Submit"


zacharycohn:
People keep saving, thinking it's submitted. But it's not, and then they never circle back to submit. According to Ric, this has happened 20+ times since the "Save" feature was enabled.

Can we change the "Save" button label to "Save, but not Submit" (or something equally clear).

batemapf:
how about some visual distinction b/w the buttons (space, colors) and change the wording to "save only"?

zacharycohn:
Revision with @batemapf's advice: 

"how about hitting save takes you to a “thank you, but you still need to submit” page before dumping you back to tock main page?"

batemapf:
i'm game for this. @MKathrynC any interest in taking this on?

nicoleslaw:
+1 @batemapf

batemapf:
addressed via #332

PR #332, opened 2016-03-10
added a reminder for users to submit their timesheets

dsiddy:
This pull request addresses #317.

When a user saves (but not submits) a timesheet, she is presented with a flash message which both confirms that her timesheet has been saved and reminds her to submit her timesheet when she has finished her edits.

To facilitate styling, I created an `alert` mixin with an optional `$color` argument.  The flash message is presented in yellow, to contrast with the static `alert-red` message.  I then recompiled the stylesheets.


batemapf:
@dsiddy looks great! thank you for the contribution!

Issue #335, opened 2016-03-22
Make Save button less prominent 


maya:
We can make the "Save" button less prominent so people know what to click on, and don't mistakenly click on Save when they meant to Submit.

Before:

<img width="723" alt="screen shot 2016-03-22 at 10 13 40 am" src="https://cloud.githubusercontent.com/assets/5249443/13960744/ceadb8fa-f016-11e5-9a7c-c5458ac5b872.png">

After:

<img width="741" alt="screen shot 2016-03-22 at 9 57 20 am" src="https://cloud.githubusercontent.com/assets/5249443/13960750/d55ccdee-f016-11e5-80dc-a7fb215909e4.png">

```
button#save-timecard {
    color: #477dca;
    box-shadow: inset 0 0 0 2px #477dca;
    background: none;
}
```

+ I'd darken the above to the hover color on hover.

maya:
If I knew which Sass file to edit, I'd be happy to open the PR.

brendansudol:
hi @maya -- i'm happy to make this change later this afternoon

brendansudol:
hi @maya -- i created a PR with this change (https://github.com/18F/tock/pull/336)

cc: @batemapf

PR #336, opened 2016-03-22
add a btn-outline class (to help distinguish save btn)

brendansudol:
This PR addresses https://github.com/18F/tock/issues/335.

It introduces the concept of an secondary button style (btn-outline) to supplement the primary blue buttons used throughout the app.

image


batemapf:
FTW from the inside! thanks @brendansudol!

PR #342, opened 2016-03-29
clear checkbox when new line item added to page

brendansudol:
This PR addresses feedback by Jay Huie in Slack: "1) I think there’s a small “tock bug” - where I had one item (the last in the list) to be deleted then did “add new item” and it seemed to inherit that checkbox status."


batemapf:
:+1:

PR #348, opened 2016-04-23
Add org to employee model

batemapf:
- add employee unit field to employee model
- add choices for employee unit field
- update tests for employee model

purpose of this PR is to append more detailed employee information to tock data for: (1) business analytics; (2) billing information; and (3) end reliance on lower quality spreadsheet based roster information for the prior to items.


brendansudol:
lgtm :)

Issue #340, opened 2016-03-28
As a user I would like to see the date range Im entering time for displayed on the main part of my screen so I can be sure Im entering time for the correct time period.


jtag:
right now the only assurance is built into the url. 
with new feature pulling up last weeks' time categories the user needs assurance which time period they are entering is correct.

batemapf:
deployed!

jtag:
<heart>

On Fri, May 6, 2016 at 2:36 PM, Patrick Bateman <notifications@github.com>
wrote:

> Closed #340 <https://github.com/18F/tock/issues/340>.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/18F/tock/issues/340#event-653181509>
>



-- 

Jesse Taggert
Strategist | 18F |  GSA

jesse.taggert@gsa.gov
cell: 415-730-4965
18F.gsa.gov

PR #349, opened 2016-04-23
address issue #340 re period displayed on timecard entry form

batemapf:
- makes small ux and content changes, including updating 18f logo
- adds new 18f logo to assets
- addresses #340


tock/tock/static/css/style.css
```diff
@@ -204,10 +204,10 @@ picture {
   max-width: 100%; }
 
 .banner {
```
brendansudol: this css needs some formatting / linting cleanup in general -- i will do this when i get a chance soon -- but i don't think it should block this PR

brendansudol:
🚢 😄

brendansudol:
not sure why this is red -- it passed on the checks but then i accidentally clicked the "close issue" button instead of "merge" and it went red -- i think we can safely ignore

batemapf:
looks like it went red b/c it was closed before travis ci build completed. concur, safe to ignore :-)

PR #357, opened 2016-05-25
Closes #272: Adds support for alerts on projects

### LINKED ISSUE #272: ALOHA notification
Acceptance test criteria to come...but here's the user story:

> As an employee, I want tock to provide an alert to remind me to submit a leave request in ALOHA if I have put in OOO hours in the timecard, so that I can comply with leave rules and regulations.

@batemapf in thinking about this, I wonder whether we should also add an "Office Closed" category to tock to distinguish between situations where ALOHA is required and situations where it is not.

@geramirez I wonder whether the best way to handle this is to abstract some sort of post-submit behavior that is tied to the project. I'm struggling to think of another specific example at the moment, to be honest, but it seems like there may be situations when a user should get a prompt based on entering data for a specific project.

In other words, in [models.py](https://github.com/18F/tock/blob/master/tock/projects/models.py) we'd add a field to the Project class:

```
user_notification = models.CharField(max_length=200, blank=True, null=True, verbose_name="Provide an alert to a user")
```

and in (views.py)[https://github.com/18F/tock/blob/master/tock/hours/views.py], we pass the alert for any project where `user_notification is not null`.

Thoughts?

ccostino:
Given the recent audit of Tock `Out of Office` time entries against ALOHA numbers and the large amount of discrepancies, I felt it prudent to revisit this feature request in #272 in order to help ourselves in the future.  This changeset adds support for adding one or more optional alerts to any given project that will in turn be displayed on the time entry form as the user is entering their time.  If any alerts are configured for the selected project, they will appear directly below the selection menu and above the notes/snippets field (if that is set to appear).

It is a different approach then what you had initially suggested, @vzvenyach, and it might still be useful to implement a form submission response/verification that contains such reminders as well.  I opted for this route in order to provide more immediate feedback.

This feature has a few extras with it:

* You can add an optional label before the alert message.
* You can change the style (formatting) of the alert (four options are available at the moment).
* You can optionally choose to **bold** and/or _italicize_ the alert message.
* You can add an optional URL to wrap the alert in, e.g., linking to a policy document around leave.
* You can associate the same alert with more than one project.
* The alert messages are displayed on the project listing and project details page for convenience.

/cc @batemapf and @brendansudol


tock/hours/forms.py
```diff
@@ -66,11 +69,14 @@ def render_option(self, selected_choices, option_value, option_label):
             if dict.get(option_label, 'notes_required'):
                 notes_required_html = ' data-notes-required="true"'
 
+            if dict.get(option_label, 'alerts'):
+                alerts_html = 'data-alerts="%s"' % (escapejs(json.dumps(option_label['alerts'])))
```
ccostino: I'd like to clean up this line if possible.

tock/tock/templates/hours/timecard_form.html
```diff
@@ -122,6 +124,25 @@
...
   }
 
+  function displayAlerts(selectBox) {
+    var $fieldset = $(selectBox).parent().parent(),
+        $selected = $(selectBox).find(':selected'),
+        $alerts = $fieldset.find('#entry-alerts'),
+        all_alerts = $selected.data('alerts');
+
+    if (all_alerts !== undefined) {
+      all_alerts = JSON.parse(JSON.parse('"' + all_alerts + '"'));
```
ccostino: I'd also like to clean this up if possible.

batemapf:
@ccostino THANK YOU for taking this on. going to pull it locally and take a look around. 💯

ccostino:
Well I thought I fixed the n+1 errors but I guess not - something to look at when adding tests in.

ccostino:
@jmcarp:  I'm trying to address the n + 1 queries with this changeset and the couple of times I've been able to make any change that has an effect, `nplusone` changes its tune from telling me that there are potential n+1 queries to potentially unnecessary eager loads.  If you have a moment at some point, would you mind pairing with me on this or at the very least taking a look to see if you might know how to address these?  The two spots in particular that appear to be problematic based on the tests that are failing are these:

* https://github.com/18F/tock/blob/project-alerts-%23272/tock/hours/forms.py#L101
* https://github.com/18F/tock/blob/project-alerts-%23272/tock/projects/views.py#L15

Absolutely no rush on this.  Thanks in advance, I appreciate it!

brendansudol:
It looks like there may be a bug in the `nplusone` logic: if you add a `.prefetch_related('alerts')` to the queryset in the `ProjectListView`, it throws a `Potential unnecessary eager load detected` error.

From the docs, this error should occur when "your app eagerly loads related data without accessing it"; but the thing is, we *are* accessing it -- in the template (see `project_list.html`), just not in the view. If you loop through the projects and access the `alerts` in the view, the error no longer occurs, but this is not ideal.

```python
def get_queryset(self):
    projects = Project.objects.all().prefetch_related("alerts")
    for p in projects:
        for a in p.alerts.all():
            pass

    return projects
```

jmcarp:
This turned out to be a really interesting (read: frustrating) bug--nplusone hooks into iteration over querysets, but the django template renderer doesn't actually call `__iter__` on a for loop over an empty queryset. Normal python *does* call `__iter__` in this case, which is why the nplusone tests pass, and iterating in `get_queryset` works. I'll cut a new release of nplusone in a day or two and update here afterward.

ccostino:
Awesome, thank you so much for providing the extra info behind what was going on @brendansudol and for looking into this further with nplusone, @jmcarp.  I greatly appreciate the extra help from both of you!

ccostino:
Thanks again, @jmcarp!  I'll wrap this up in the next day or so to get it prepped for final review. :-)

tock/hours/forms.py
ccostino: I was not able to find a better alternative to turning the alert information into a JSON string and then escaping it on top of that in order to incorporate it into the existing page construction.  I'm open to suggestions if anyone is aware of a better way, though this does work.  However, it requires that `JSON.parse` be called on the value twice in the JavaScript side, which comes off as a code smell to me.

ccostino:
@batemapf, this is now finished and ready for full review.  Thanks!

ccostino:
@batemapf and I just took a look at this, there seems to be one issue to fix when viewing the project details, I'll investigate.

ccostino:
Capturing notes from a conversation in #django:  When checking the project details page an error like the following would be thrown:

```
Traceback:
File "/home/vagrant/.pyenv/versions/tock/lib/python3.4/site-packages/django/core/handlers/base.py" in get_response
  223.                 response = middleware_method(request, response)
File "/home/vagrant/.pyenv/versions/tock/lib/python3.4/site-packages/nplusone/ext/django/middleware.py" in process_response
  45.             listener = self.listeners[request].pop(name)
File "/home/vagrant/.pyenv/versions/3.4.1/lib/python3.4/weakref.py" in __getitem__
  345.         return self.data[ref(key)]

Exception Type: KeyError at /projects/8
Exception Value: <weakref at 0xb477f554; to 'WSGIRequest' at 0xb53325cc>
```

Which could be related to this line with nplusone:  https://github.com/jmcarp/nplusone/blob/master/nplusone/ext/django/middleware.py#L45

However, there is likely some other underlying issue taking place.  Might have to investigate in a later PR.

jmcarp:
Try updating to `nplusone==0.7.2`, which handles the edge case where the middleware runs `process_response` but not `process_request`.

ccostino:
Awesome, thanks @jmcarp!  I'll check soon and let you know.

batemapf:
per @jmcarp, upped `nplusone` to 0.7.2 in `requirements-dev.txt`. merged this branch locally with `master` to test out with docker set up (believe or not, this branch started _before_ toolness introduced us to better living through docker) and tried to invoke @ccostino error via the the projects page. seems like it did the trick! @ccostino you will probably want to test locally as well to verify.

ccostino:
@batemapf - awesome, I'm glad that worked!  I had planned on updating this as soon as the other PR is merged so it becomes merge ready again, but as you saw there we uncovered a few other things to address first. :-)

provision/dev/nginx_site_docker.conf
```diff
@@ -4,7 +4,7 @@ server {
 
   location / {
   proxy_set_header X-Forwarded-Email "aaron.snow@gsa.gov";
-    proxy_set_header Host 192.168.33.10;
+    proxy_set_header Host 192.168.99.100:8000;
```
ccostino: @toolness:  We (@batemapf and I) were running into an issue when viewing the project details pages and when saving changes to a project in the admin.  This `nginx` config line is the culprit, but do you have any idea why this is needed?  Alternatively, what had you pulled from to set these values here, the `Vagrant` config?

Also, is it possible to somehow set this dynamically?  What I'm wondering is if we could somehow set it to whatever the value is of `docker-machine ip default` when run on anyone's local machine, in case the IP address is different, and use the `EXTERNAL_PORT` environment variable defined in the `.env` file.

The best answer I could find to this was in the [nginx Docker image documentation](https://hub.docker.com/_/nginx/) (under the "using environment variables in nginx configuration" heading).  I got as far as being able to at least set an environment variable to the output of that command (e.g., `DOCKER_MACHINE_IP=$(docker-machine ip default)`), but I couldn't get the config file to be valid.

ccostino:
@batemapf:  I have one outstanding question for @toolness related to the Docker config, but I believe that will only impact local development and has no bearing in the staging or production environments.  This should finally be done!

batemapf:
woohoo! great work on this one @ccostino and 💯 for the assists from @jmcarp and @toolness on the issues that were raised here and in #370! 18F FTW!

provision/dev/nginx_site_docker.conf
toolness: Oh, good question--I did indeed get those values from the original Vagrant-specific [`provision/dev/nginx_site.conf`](https://github.com/18F/tock/blob/afd3ab719b09a7ba698e52a7126700534a2e21c3/provision/dev/nginx_site.conf).
One really annoying thing about nginx to me is that it's actually very difficult to even use environment variables in nginx configuration files.  I didn't realize that the specific `Host` value actually mattered--my impression was that it just needed to be some value considered "internal" so that the Django server would let it through, or something weird like that.

Does tock run on cloud.gov?  If nginx isn't absolutely critical for deployment, I'm wondering if we could write some simple WSGI middleware that injects these headers, instead of fronting the app with nginx during development.  I'm unfamiliar with the architecture of Tock, though, so I'm not sure if this would actually be a good solution...

toolness: Update: see #377!

PR #333, opened 2016-03-15
Add docker compose setup

milafrerichs:
## What does this PR do?

It adds docker-compose version 2 setup to create a local environment for tock. 
Make sure to have the latest version of Docker and docker-compose installed.

All commands are explained in the Readme.
This is part of the Micropurchase Experiment from 18f.

### A few comments
I opted to use shell scripts to run specific commands. Almost everything can be used in a production environment.

PR #359, opened 2016-05-31
Alternative docker-compose integration

toolness:
This cribs some things from #333 and https://github.com/18F/projects/pull/3 to add docker-compose integration with some nice developer ergonomics.

## Trying it out

1. Install docker/docker-compose. If you're on OS X/Windows, you'll also need to do everything in a Docker Quickstart Terminal, at least until [docker goes native](https://blog.docker.com/2016/03/docker-for-mac-windows-beta/).

2. `cd` into the repository's `tock` directory (that is, the same directory that `manage.py` is in).

3. Run:

  ```shell
$ cp .env.sample .env
$ docker-compose build
$ python manage.py migrate
$ python manage.py loaddata test_data/data-update.json  
$ docker-compose up
  ```

  (If the `manage.py` commands don't work, try running them with `docker-compose run app` in front of them, e.g. `docker-compose run app python manage.py migrate`.)

4. If you're on linux, visit http://localhost:8000. If you're on OS X or Windows, you will probably need to visit http://192.168.99.100:8000 instead; if that doesn't work, try visiting port 8000 of the IP returned by running `docker-machine ip default`.

At this point, you can edit any of your project's source code files (Python code, template views, SASS, etc) and everything should be rebuilt automatically. Any errors will be logged to stdout.

## To do

- [x] Add docs to readme. Will probably want to consolidate them w/ the vagrant docs since both will have common elements. [You can view the new README here.](https://github.com/18F/tock/blob/atul-docker-compose/README.md)
- [x] Make it possible to dynamically configure the port the web server is exposed at.
- [x] Add a `sass` container that continuously runs sass.
- [x] Define [`PYTHONUNBUFFERED`](https://docs.python.org/2/using/cmdline.html#envvar-PYTHONUNBUFFERED) so that the stdout from `manage.py runserver` always shows up.
- [x] Add [graceful exit to `manage.py` via signal handler](https://github.com/18F/projects/pull/3/files#diff-632b651e5579af0e928ed63482e9bc77R6).
- [x] Add convenient [`docker-compose` proxying to `manage.py`](https://github.com/18F/projects/pull/3/files#diff-632b651e5579af0e928ed63482e9bc77R45).
- [x] Consider adding some code to `manage.py` that waits for the db to be up, to obviate [race conditions like this](https://github.com/18F/tock/pull/333#issuecomment-198567190). (Note that if we do this, would be great to add it to the `18/projects` repo too!)

## Concerns

I've hacked `manage.py` to proxy all invocations through to docker-compose if its import of `django.core.management` fails. For the vast majority of cases, this means that running `python manage.py` on the host machine is just like running `docker-compose run app python manage.py`, which saves *tons* of typing.

~~However, if the developer has Django installed system-wide, then then import of `django.core.management` will succeed and `manage.py` *won't* run itself inside docker. This could potentially confuse anyone who has Django installed system-wide, so an alternative may be to write out the full `docker-compose run app python manage.py` in all installation instructions, but mention the proxied invocation as a convenience that works in most cases.~~

**Update:** I ended up just writing out the full `docker-compose run app python manage.py` in all installation instructions, but mention the proxied invocation as a convenience that works in most cases.

## Limitations

Right now this docker-compose setup is largely intended for *development* rather than deployment. For example, `docker-compose up` runs `manage.py runserver` under the hood to ensure that assets are always recompiled when they change. It shouldn't be too hard to make it work for production too, but that's out of the scope of this PR for now.


provision/dev/nginx_site_docker.conf
```diff
@@ -0,0 +1,11 @@
+server {
+  listen 80 default_server;
+  server_name _;
+
+  location / {
+  proxy_set_header X-Forwarded-Email "aaron.snow@gsa.gov";
+    proxy_set_header Host 192.168.33.10;
+  proxy_set_header X-forwarded-for $proxy_add_x_forwarded_for;
+    proxy_pass http://app:8000;
```
toolness: I'm not a big fan of this because it's just a copy of `nginx_site.conf`.

Ideally we could use [environment variables](https://github.com/docker-library/docs/tree/master/nginx#using-environment-variables-in-nginx-configuration) here, but this would require changing the vagrant setup too, and I'm unfamiliar with vagrant.

Another option would be to do some kind of janky replacement of `http://127.0.0.1` with `http://app` from `nginx_site.conf`, but that feels janky, and makes any docker-specific customizations we might want to do in the future that much harder.

It might just be easiest to violate DRY a bit here, which shouldn't be *that* bad, especially since this dev nginx configuration isn't likely to change that often. The two configs are also in the same directory and named similarly, so it shouldn't be hard to modify both at the same time (especially if we add comments to the files mentioning this).

docker-compose.yml
```diff
@@ -0,0 +1,24 @@
...
+  environment:
+    - DATABASE_URL=postgres://tock_user@db/tock
+  command: python manage.py runserver 0.0.0.0:8000
+db:
+  image: postgres:9.4
+  environment:
+    - POSTGRES_DB=tock
+    - POSTGRES_USER=tock_user
+web:
+  image: nginx
```
toolness: Um, I'm not actually sure how #333 even worked without adding this nginx proxy.  The tests run fine without it, but visiting the server without the nginx proxy just results in a `403 Forbidden` for me. It seems the dev nginx configuration adds the appropriate headers to make the app think it's always authenticated.

toolness: Oh, upon closer inspection, it seems that the `403 Forbidden` on `/` was an issue that was never resolved: https://github.com/18F/tock/pull/333#issuecomment-198567190

Well, it works now, so yay. I'm surprised that *all* routes didn't raise `403 Forbidden` without nginx's additional dev proxy headers though... Could that actually be a security bug?

Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM python:3.4
+
+RUN apt-get update && \
+  apt-get install -y ruby-full rubygems && \
+  gem install sass
+
+COPY requirements.txt requirements-dev.txt /tock/
+
+RUN pip install -r /tock/requirements.txt -r /tock/requirements-dev.txt
```
toolness: Note that this isn't being 100% faithful to the [Minimize the number of layers](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#minimize-the-number-of-layers) best practice for Dockerfiles, but I wanted to make it as easy as possible to iterate on the requirements file. Coalescing both `RUN` commands into a single one, while minimizing layers, would also mean that tinkering with requirements would also re-install ruby and sass.

docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+  volumes:
+    - ./provision/dev/nginx_site_docker.conf:/etc/nginx/conf.d/nginx_site_docker.conf
+  links:
+    - app
+  ports:
+    - "8000:80"
+sass:
+  build: .
```
toolness: Note that I could've instead used the official [ruby Docker Hub image](https://hub.docker.com/_/ruby/) but then we would've had to create a separate ruby equivalent of `docker_entrypoint.py` so that SASS wouldn't create all the compiled CSS with root ownership. So it was easier to just add ruby+sass to the app container image, and reuse that.

toolness:
@adelevie, I think this PR is good to go, with the notable exception of the documentation.

Question: I assume you still want to support vagrant as a development path?  If not, I can simplify the README a lot by only referring to what's needed for docker. Otherwise I'll have to consolidate parts of the README by saying things like "if you're using docker, do X, but if you're on vagrant, do Y instead".

adelevie:
> Question: I assume you still want to support vagrant as a development path?

I think it's OK to abandon Vagrant if the docker-compose support is good enough.

toolness:
Ok @adelevie, I removed the Vagrant stuff and documented the new instructions in the readme. I think it's good to go!

Let me know if you run into any problems.

toolness:
By the way, I was just reminded that our dev standardization guides exist, so instead of putting Docker advice on every single repository that uses Docker, I thought it might be a better idea to improve the Docker guide and link to that. Here's the PR for it if you're interested:

https://github.com/18F/dev-environment-standardization/pull/16

toolness:
**Update:** I just dog-fooded this PR by using its docker-compose environment to work on #360. Everything pretty much went without a hitch--I was able to edit files and django would auto-reload the app, and I was also able to run tests easily.

batemapf:
@toolness this is fantastic work +100. ran perfectly for me, excellent documentation. even used it to test out #365. merging!

PR #360, opened 2016-06-03
Allow autocomplete by project number.

### LINKED ISSUE #353: Allow auto-complete of tock line by tock project number
As a tocker, I want to be able to type "300" on a tock line and have tock automatically complete the project in the timecard to be "300 - BU / Acquisition Services Management" so that I can more easily find the project lines that I use.

toolness:
This fixes #353. As specified in that issue, the project labels are now prefixed with their project ID, followed by a dash:

In the above screenshot, typing `99` into the text field would cause **99 - 38r3aG Project** to be chosen. Alternatively, `38r3aG` could also be typed, and the widget would still work as it previously had (i.e., there's no regressions in the UI).

**Note:** I'm assuming here that all users have JS enabled. If they don't, and if the Chosen widget actually appears as a standard `<select>`, then some user habituation could be broken.

### DIFF
--- a/tock/hours/forms.py
+++ b/tock/hours/forms.py
@@ -88,7 +88,7 @@ def projects_as_choices():
             projects.append([
                 project.id,
                 {
-                    'label': project.name,
+                    'label': '%d - %s' % (project.id, project.name),
                     'billable': code.billable,
                     'notes_displayed': project.notes_displayed,
                     'notes_required': project.notes_required,
--- a/tock/hours/forms.py
+++ b/tock/hours/forms.py
@@ -74,6 +74,15 @@ class SelectWithData(forms.widgets.Select):
             conditional_escape(force_text(option_label)))
 
 
+def choice_label_for_project(project):
+    """
+    Returns the label for a project as it should appear in an
+    auto-completable list of choices.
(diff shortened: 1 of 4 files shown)


tock/hours/forms.py
```diff
@@ -74,6 +74,15 @@ def render_option(self, selected_choices, option_value, option_label):
             conditional_escape(force_text(option_label)))
 
 
+def choice_label_for_project(project):
+    """
+    Returns the label for a project as it should appear in an
+    auto-completable list of choices.
+    """
+
+    return '%s - %s' % (project.id, project.name)
```
toolness: Note that I could have added this as a method to the `Project` class, but I didn't want to bloat that class, since this functionality seems fairly specific to forms (and the timecard form in particular).  I could move it over to the class if that's preferred, though.

toolness: I also decided to factor `choice_label_for_project` out as a function so that tests could use it; that way if we decide to change the labeling in the future, we just change it in the function and the function's test, rather than causing regressions in other unrelated tests.

batemapf:
another excellent one! thanks @toolness!

Issue #363, opened 2016-06-07
Create API documentation


jessieay:
Hello --

a few API questions

1. Where is "Tock" in the list of projects? Viewed every page available and do not see it https://tock.18f.gov/api/projects.json but it is listed at https://tock.18f.gov/projects/

2. Are there API docs somewhere? Trying to figure out who I can up the number of records per page. Seems like there's a way but I am not fluent in python. English docs would be much appreciated!

jessieay:
ps - this is for use in Micropurchase. We are going to populate the "billable to" field in the auctions#new form with data from the Tock API. Excited to use it!

batemapf:
@jessieay `Tock` may not be showing up from api/projects.json b/c of page limits. i believe projects.json is limited to 100 lines until you have send a ?page=2 variable.

only api documentation that exist are in the `README.md` file in the root.

i wrote a little ditty to automatically pull the next page of data which dealt with the problem nicely. now i just need to find it...

batemapf:
ah, looks like that was for something completely different (the Lever API, which has similar pagination issues).

i am going to go ahead and see if i can't kill the pagination for projects. back soon.

jessieay:
@batemapf yeah I saw the page stuff and looked at every page but the Tock project was nowhere to be found!

Killing pagination would be fine. Or, allowing higher page limits would also work (default to 100, but allow up to 1000 or something)

jessieay:
Hi @batemapf - any update on this? We are trying to use the Tock API for a feature in Micropurchase so any leads on how I can help myself to use it would be awesome.

abrouilette:
@jessieay, we're spinning up a Tock team.  Is this still a need?  (I don't speak API very well.)

jessieay:
yes, this is still a need. related: https://github.com/18F/tock/issues/391#issuecomment-235320070

jessieay:
High level summary of this item: Tock API needs documentation

batemapf:
@jessieay is this one OK to close?

jessieay:
Yes, looks like docs are here https://github.com/18F/tock/tree/master/api-docs

PR #365, opened 2016-06-10
remove pagination_class from ProjectList view

batemapf:
to address pagination issue in #363


batemapf:
hrmm. tests run fine locally. another `n +1` issue, though not sure of similarity to @ccostino issue with #357. ❓

ccostino:
Ah!  No nplusone errors will be raised if it's not set to do so in the corresponding environment config.  It's not enabled in the `dev` settings, which are used locally.  It is enabled in `test`, which Travis CI uses:  https://github.com/18F/tock/blob/master/tock/tock/settings/test.py#L17

As for what's actually throwing the error, I'd have to look a bit more to see where the queryset is being accessed.  It's the same issue that I was dealing with in terms of the cause of the test failing (a many-to-many relationship between models), but this is a different part of the code.

I do wonder if merging the PR I have open and then merging those updates back into here would also fix the issue - I will try and get that last issue we saw last night either resolved or explained so we can move ahead with it, because the PR contains and update to nplusone as well.

jmcarp:
This is a fun issue. What it all boils down to is that `TimecardView.last_timecard` loads more data than it needs to: https://github.com/18F/tock/blob/master/tock/hours/views.py#L171-L178. The trick is that calling `if timecards` loads *all* results in the `timecards` queryset, even though we only want the last one. Since we're then lazy-loading relations of the last timecard, we get a lazy-load warning. This is fixed in #357, which calls `first` on the queryset instead of loading all records and returning the 0th one: https://github.com/18F/tock/blob/382e70195ff66470a03097c14bff84dd67d0bfe2/tock/hours/views.py#L171-L178.

batemapf:
all fixed. thanks again @jmcarp and @ccostino

PR #369, opened 2016-06-13
Update README.md

batemapf:
@ccostino - this should make things clearer. also added some help for an error message i have run into a few times and had to google. merge if it looks good!


ccostino:
Awesome, thank you @batemapf!

PR #370, opened 2016-06-13
Improve Django Debug Toolbar integration with Docker

ccostino:
This changeset improves the integration of `django-debug-toolbar` with the recent support added for running Tock in a Docker container.  It also addresses an underlying issue that was present where the incorrect settings file was being used for running the `manage.py collectstatic` command and when running the app in staging or production.

The following list is a brief summary of all of the changes I made:

* Remove reference to Vagrant files and add Docker files to `.cfignore`.
* Add the `DJANGO_SETTINGS_MODULE` environment variable to `docker-compose.yml`.
* Always reference the `EXTERNAL_PORT` environment variable in `docker-compose.yml`.
* Remove `django-debug-toolbar` from `requirements.txt`.
* Default to running with `tock.settings.production` in cases where `DJANGO_SETTINGS_MODULE` is not set for consistency (this also prevents debug information and the debug toolbar from loading in places it shouldn't).
* Make use of the `DEBUG_TOOLBAR_CONFIG['SHOW_TOOLBAR_CALLBACK']` dict setting for Django Debug Toolbar to account for differing container IPs on everyone's machines.
* A bunch of general cleanup and consolidation of settings and project config related files.

I believe I have accounted for everything but please let me know if anything is out of place, missing, or still not configured correctly.

One last thing I'm still not sure about here too is running the tests (typically run via `manage.py test` or `manage.py test --settings=tock.settings.test`).  As you can see, Travis ran them just fine but that's because they're not being setup and run in with the Docker container.  When trying to run them locally, the only way I was able to get them to work is to still setup a virtualenv for the Tock project and run them in that (not the Docker container); this also requires specifying that `--settings` argument.

Is there a better way of setting up and running tests when using Docker?  Regardless, the README should probably have a bit of info around this as well, so whatever the accepted solution is I'm happy to update it with instructions as a part of this.

/cc @batemapf, @toolness, @jcscottiii

### DIFF
--- a/tock/tock/settings/dev.py
+++ b/tock/tock/settings/dev.py
@@ -1,24 +1,28 @@
-from .base import *  # noqa
-
 import dj_database_url
 
+from .base import *  # noqa
+
 DEBUG = True
-TEMPLATE_DEBUG = True
+TEMPLATE_DEBUG = DEBUG
+DEBUG_TOOLBAR_PATCH_SETTINGS = False
 
-INTERNAL_IPS = ('127.0.0.1',)
-
-DATABASES = {}
 DATABASES['default'] = dj_database_url.config(
     default='postgres://tock:tock@localhost/tock'
 )
 
-INSTALLED_APPS += ('debug_toolbar',)
+INSTALLED_APPS += (
+    'debug_toolbar',
+)
 
-DEBUG_TOOLBAR_PATCH_SETTINGS = False
+MIDDLEWARE_CLASSES += (
+    'debug_toolbar.middleware.DebugToolbarMiddleware',
+)
 
-MIDDLEWARE_CLASSES += ('debug_toolbar.middleware.DebugToolbarMiddleware',)
-
-INTERNAL_IPS = ['127.0.0.1', '::1', '192.168.33.10']
+INTERNAL_IPS = [
+    '127.0.0.1',
+    '::1',
+    '192.168.33.10',
+]
 
 MEDIA_ROOT = './media/'
 MEDIA_URL = '/media/'
--- a/tock/tock/settings/dev.py
+++ b/tock/tock/settings/dev.py
@@ -6,6 +6,11 @@ DEBUG = True
 TEMPLATE_DEBUG = DEBUG
 DEBUG_TOOLBAR_PATCH_SETTINGS = False
 
+INTERNAL_IPS = [
+    '127.0.0.1',
+    '::1',
+]
+
 DATABASES['default'] = dj_database_url.config(
     default='postgres://tock:tock@localhost/tock'
 )
@@ -18,14 +23,26 @@ MIDDLEWARE_CLASSES += (
     'debug_toolbar.middleware.DebugToolbarMiddleware',
 )
 
-INTERNAL_IPS = [
-    '127.0.0.1',
-    '::1',
-    '192.168.33.10',
-]
-
 MEDIA_ROOT = './media/'
 MEDIA_URL = '/media/'
+
+# Due to the Docker configuration and Django Debug Toolbar's need to account
+# for the local machine's IP address in INTERNAL_IPS to display itself, opt
+# to show the debug toolbar with a custom callback instead. For more
+# information on this setup please take a look at these resources:
+# https://django-debug-toolbar.readthedocs.io/en/1.4/installation.html#internal-ips
+# https://django-debug-toolbar.readthedocs.io/en/1.4/configuration.html#toolbar-options (SHOW_TOOLBAR_CALLBACK)
+# http://stackoverflow.com/questions/10517765/django-debug-toolbar-not-showing-up
+# https://gist.github.com/douglasmiranda/9de51aaba14543851ca3 (code taken from here)
+def show_django_debug_toolbar(request):
+    if request.is_ajax():
+        return False
+
+    return True
+
+DEBUG_TOOLBAR_CONFIG = {
+    'SHOW_TOOLBAR_CALLBACK': 'tock.settings.dev.show_django_debug_toolbar',
+}
 
 try:
   from .local_settings import *
--- a/tock/tock/urls.py
+++ b/tock/tock/urls.py
@@ -1,9 +1,7 @@
-from django.conf.urls import patterns, include, url
-# from django.conf.urls.static import static
 from django.conf import settings
-# from django.views.generic import TemplateView
+from django.conf.urls import patterns, include, url
 
-# Uncomment the next two lines to enable the admin:
+# Enable the Django admin.
 from django.contrib import admin
 admin.autodiscover()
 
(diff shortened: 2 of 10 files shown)


ccostino:
@batemapf and I looked into this some and it looks like there's something a bit off with the Django Debug Toolbar setup: http://django-debug-toolbar.readthedocs.io/en/latest/installation.html (see the warning and note in the top there).

I am going to update Django Debug Toolbar to the latest version and pin it in the dev requirements and make sure it's setup correctly as well as a part of this PR.

ccostino:
I stand corrected, it looks like the Tock setup is okay within the settings files, but there is definitely something strange that we're both seeing during the deployment in that the debug toolbar is being uninstalled and then the app is initialized but an `ImportError` is being thrown because it the `debug_toolbar` module cannot be found.  Aside from local dev environments, it should not be looking for it in the first place based on the `DEBUG` setting.

toolness:
Oh dang, I think I know why that's happening...

So according to the log @ccostino showed me on Slack, the error is being thrown when the server is running `manage.py collectstatic`. And the [`manage.py` says](https://github.com/18F/tock/blob/afd3ab719b09a7ba698e52a7126700534a2e21c3/tock/manage.py#L72):

```py
os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'tock.settings.dev')
```

So even though the server's WSGI app is likely running with production settings on that staging server, `manage.py collectstatic` is running with *development* settings, which is triggering the load of `debug_toolbar`. :disappointed:

I can think of a few potential fixes here but they mostly rely on using 12factor's [Store config in the environment](http://12factor.net/config) to work--I'm hoping the changes wouldn't break anything about the way tock currently works, or is being deployed on cloud.gov...

toolness:
Wait, I should actually clarify an assumption I had.

I'm assuming/guessing that `DJANGO_SETTINGS_MODULE` isn't being explicitly defined on staging, but that the WSGI app is nonetheless using production settings, because there's [a line in `wsgi.py`](https://github.com/18F/tock/blob/c70c77def254c0703db5ef06d87fea9bc81dd63a/tock/tock/wsgi.py#L11) that says:

```py
os.environ.setdefault("DJANGO_SETTINGS_MODULE", "tock.settings.production")
```

This would explain why staging behaves like a production server, but is throwing the `ImportError` on `manage.py collectstatic`.

I could be wrong, though. If `DJANGO_SETTINGS_MODULE` *is* being defined on staging, then everything I've said so far is bupkis. :sweat_smile: 

If I'm right, though, I think one way to fix this on cloud.gov is just to explicitly set `DJANGO_SETTINGS_MODULE` to `tock.settings.production`.  That way the line in `manage.py` would never default it to `tock.settings.dev`, and that `ImportError` would never get thrown.

ccostino:
Ah, that discrepancy might very well be it, @toolness!  I'll double check the CF environment to verify and see if that might be the cause.

ccostino:
Okay, this turned out to be a bit more than I anticipated!  I think I have it working though and I've updated the title and description accordingly.  Tests are the only thing I'm completely unsure of.

@toolness, thanks again for taking a look before!  I'd definitely appreciate your eyes again on this set of changes, and you as well @jcscottiii if you have a moment at some point. :-)

Thank you both!

tock/tock/urls.py
```diff
@@ -13,28 +11,40 @@
...
-
-    # Uncomment the next line to enable the admin:
+    # Enable the Django admin.
     url(r'^admin/', include(admin.site.urls)),
 )
 
 
+# Enable Django Debug Toolbar only if in DEBUG mode.
 if settings.DEBUG:
     import debug_toolbar
```
toolness: This might be over-optimizing, but I'm wondering if it might be better for this conditional to branch off `if 'debug_toolbar' in settings.INSTALLED_APPS` instead of `if settings.DEBUG`?  That way just in case someone has `DEBUG` enabled but doesn't have `debug_toolbar` installed, an exception won't be thrown. (For that matter, I'm uncertain why the debug toolbar doesn't add these routes itself, actually.)

Anyways, not a big deal either way!

tock/tock/settings/dev.py
```diff
@@ -1,28 +1,49 @@
...
 DEBUG = True
-TEMPLATE_DEBUG = True
+TEMPLATE_DEBUG = DEBUG
+DEBUG_TOOLBAR_PATCH_SETTINGS = False
 
-INTERNAL_IPS = ('127.0.0.1',)
+INTERNAL_IPS = [
+    '127.0.0.1',
+    '::1',
+]
```
toolness: I'm curious what this will be used for?  For example, I often access my machine from a cell phone on my wifi network to ensure that mobile works OK, but then I'm not accessing it directly via the loopback interface and I might not have access to some functionality.

Is `::1` to support IPv6?

toolness:
Thanks @ccostino, this looks great! I've submitted a potential fix for the unit testing issue you mentioned as #375.

ccostino:
Thanks @toolness!  I just merged and tested this and it addresses the issue, however I still have a question around it now.  If you build the container and whatnot and run it, then later on decide to run the tests (`docker-compose run app python manage.py test`), it would still be using the same database as the `dev` settings, right?

I think this is the case because the next time you load up the container, you see this in the log output:

```
db_1    | ERROR:  duplicate key value violates unique constraint "hours_timecard_user_id_1c56f055733fd506_uniq"
db_1    | DETAIL:  Key (user_id, reporting_period_id)=(1, 31) already exists.
db_1    | STATEMENT:  INSERT INTO "hours_timecard" ("user_id", "reporting_period_id", "submitted", "created", "modified") VALUES (1, 31, false, '2016-06-15T13:10:41.573828+00:00'::timestamptz, '2016-06-15T13:10:41.573851+00:00'::timestamptz) RETURNING "hours_timecard"."id"
```

So what I'm wondering now is if in this new world with Docker, we should share the `dev` and `test` settings, or if there's something else we can add to the `docker-compose.yml` explicitly for running tests instead, such as a separate database configuration.

toolness:
Oh, actually, I've never worked on a django project that had separate `test` settings when running tests, regardless of whether Docker was being used... I notice in Tock's `settings/test.py` it's specifying a different database called `tock-test`, but [django actually uses a separate database for testing by default](https://docs.djangoproject.com/en/1.9/topics/testing/overview/#the-test-database).

So I definitely recommend using only `dev` settings when testing, and having a minimal amount of branches based on whether tests are being run (like #375)... We shouldn't have to configure the database differently when running tests unless we're doing something really out of the ordinary--as I understand it, at least.

toolness:
Huh, I'd never heard of [nplusone](https://github.com/jmcarp/nplusone), which is mentioned in the `test` settings, but it seems like something that'd be useful for `dev` too...

ccostino:
[nplusone](https://github.com/jmcarp/nplusone) is written by @jmcarp and is extremely handy in identifying parts of an app that rely on lazy loads via an ORM when they shouldn't.  It's very useful to have enabled when running tests but I agree, it can also be handy to have enabled in dev as well when actively trying to debug and improve query performance.

The other PR I have open accounts for this and enables it in dev to make it a bit easier/quicker for folks to work with. :-)

tock/tock/urls.py
ccostino: Ah, I see what you're saying, but in this case I think it's warranted for two reasons:

* The project requirements (the dev requirements in particularly) specify Django Debug Toolbar as a dependency and therefore I think it's reasonable to expect that a person would have it available as a part of just setting up the project for themselves.
* This is the recommended approach in the [Django Debug Toolbar documentation for the explicit configuration](https://django-debug-toolbar.readthedocs.io/en/1.4/installation.html#explicit-setup); the explicit configuration is already the recommended way to go, but in the upcoming 1.5 version that will be even more so.

That said, Django Debug Toolbar does/did do this automatically for you before with the simple configuration, but it's not an ideal setup and not meant for production.

toolness: Got it, sounds good to me! Thanks for clarifying.

tock/tock/settings/dev.py
ccostino: This was partly already there and partly due to the instructions in [Django Debug Toolbar's setup](https://django-debug-toolbar.readthedocs.io/en/1.4/installation.html#internal-ips).  Due to the nature of how this has to be run because of Docker though, we may not need to set these.

Yes, `::1` is the IPv6 version of [localhost](https://en.wikipedia.org/wiki/Localhost#Name_resolution).

batemapf:
@ccostino caught up on this today. looks like you dug deep into this one along with @toolness, which i am super grateful for. looks like it was definitely a learning experience! everything checks out, so barring any objections to the contrary, will merge in the AM.

ccostino:
Thanks @batemapf!  I think we're good to go on this for now, and once it's in I'll get back to finishing up the other PR as it should be fine at this point. :-)

batemapf:
+100!

PR #372, opened 2016-06-13
Use escape sequences for unicode JSON chars in API output

### LINKED ISSUE #371: Some HTTP clients are confused by Tock's JSON encoding
This is a long story.

Basically, Django REST Framework, which powers Tock's API, doesn't include a `charset` parameter in its `Content-Type` header because doing so [is not actually part of the JSON spec](http://lucumr.pocoo.org/2013/7/19/application-mimetypes-and-encodings/).

However, some clients (most notably Chrome's built-in JSON renderer) expect there to be a `charset`; if there isn't, then they often mis-interpret JSON data.

Ideally the solution would be to tell Django REST Framework to include a `charset` parameter to satisfy these clients, but its creator [refuses to because it's not part of the specification](https://github.com/tomchristie/django-rest-framework/issues/2891).

Because of this, I've tended to simply set `UNICODE_JSON` to `False` to avoid the issue altogether--this setting tells Django REST Framework to simply take advantage of the JSON specification's `\u` escape sequence to deliver all JSON content as ASCII while still allowing unicode to be transmitted. While this can also increase payload sizes if they contain lots of non-ASCII characters, Tock's data seems to be mostly ASCII, so it shouldn't bloat things too much.

If this sounds like a decent enough solution, I'm willing to submit a PR for it.

Related: https://github.com/18F/api-standards/issues/68

toolness:
I *think* this fixes #371.  I haven't actually tried it out though, as I've just used GitHub's in-browser text editor to make the change.

### DIFF
--- a/tock/tock/settings/base.py
+++ b/tock/tock/settings/base.py
@@ -74,6 +74,7 @@ ALLOWED_EMAIL_DOMAINS = {
 }
 
 REST_FRAMEWORK = {
+    'UNICODE_JSON': False,
     'DEFAULT_RENDERER_CLASSES': (
         'rest_framework.renderers.JSONRenderer',
         # use our CSV renderer instead of rest_framework_csv's


jessieay:
I am no pythonista so I won't merge but ✔️  from me!

batemapf:
see comment on issue https://github.com/18F/tock/issues/371#issuecomment-225765189

Issue #26, opened 2015-03-26
Basic Reporting View


vzvenyach:
Per reporting period:

1. Project
2. Total Hours
3. Calculated Billables

PR #374, opened 2016-06-14
Feature/project start and end

scottt2:
This is for the [18F micropurchase auction #26](https://micropurchase.18f.gov/auctions/26) where we add start/end date attributes to projects.

**Feature:** As a Tock Admin user, I want to mark some projects as temporary, so the system should allow me to specify start and end dates for those projects.
- [x] Scenario: Admin can specify project start and end dates for new projects
- [x] Scenario: Admin can update start and end dates for existing projects
- [x] Scenario: Admin sees project start date and project end date in data export

**Feature:** As an 18F employee recording my time, I don’t want to mistakenly report time against a project that has not yet started, or has already ended, so the system should prevent me from doing that.
- [x] Scenario: 18F employee cannot report time against a project after it’s end date
- [x] Scenario: 18F employee cannot report time against a project before its start date

**Feature:** As a consumer of the Tock API, I want to show important information about projects (including start and end dates), so if this data is available in the system, it should be returned in the API response.
- [x] Scenario: API returns project start and end dates

CR: @joshuabailes @batemapf @mtorres253


adelevie:
That was quick, @scottt2! Can you look into and fix the breaking tests? 

Also, per the requirements ("Tests adequately cover any new functionality."), could you add some tests?

We'll have @joshuabailes and/or another product owner eval the rest of this PR shortly.

scottt2:
@adelevie yeah, well given what I learned last time, I did most the work prior to the auction. It started as doing a detailed scoping of the work then I just wound up doing it in my free time.

Fixing the tests now and good catch on the `Tests adequately cover any new functionality.` Will do.

mtorres253:
That was quick @scottt2! And I understand why you wanted an early start :). After you fix/add the tests it would be great to get an idea of how much time you spent on this altogether.

patvin80:
@scottt2  the tests failed for me when I added the start_date and end_date and instead renamed the fields to project_start_date and project_end_date. The start_date is an attribute of the Reporting Period so naming them like that will add clarity imho.

scottt2:
@patvin80 Yup. I just pushed up that fix. The issue is that some previous SQL queries used `start_date` when referring to `reporting_period`. The fix there was to just disambiguate the attribute by calling `reporting_period.start_date`.

As for the other tests, the failures are due to how I implemented TimecardObjectForm. I am addressing that by refactoring away from use of globals as they tend to not work at all with method specific unit test (great TDD in this repo, btw).

I am first trying to not disturb the current schema. I have one last idea to try. If that fails, I am going to change the TCO model to have a foreignkey on Timecard so that I can directly access the ReportingPeriod like `self.timecard.reporting_period`

As it is now, the Timecard only has `time_spent` *through* `TimecardObject`. This relationship is not bidirectional and doesn't allow me to easily query into Timecard to get the reporting period. I need the reporting period to set the querySet on the project selection for the TimecardObject form. The current implementation uses a method called `projects_as_choices` that doesn't have any context on the TimecardObject it is serving. It just pulls all active projects.

So I'd love to hear thoughts on this as I wrap it up. Like I said, I am gonna finish out this attempt which is to alter the querySet on the form fields in the `get_formset` method for the `TimecardFormSet`. If that doesn't work, refactor TCO to have a direct relationship with TC.

scottt2:
So tests pass locally. It appears there is a bit of formatting in the assertions in Travis that are failing.

image

Anyone else hit this?

adelevie:
@scottt2 , I'm curious: do tests pass locally for you, and is this just an environment parity issue? Or is this just an issue about getting tests to pass, regardless of env?

scottt2:
For example:
image

The test is asserting `32 - openFEC` but everything in that array does not have the ID.

adelevie:
Ah, there's my answer!

scottt2:
@adelevie Definitely pass locally. The terminal screenshot is my local env.

adelevie:
Are you using these commands for running tests:

https://github.com/18F/tock/blob/master/.travis.yml#L21-L24 ?

scottt2:
I tried those previously and both give me `fe_sendauth: no password supplied`. I assumed there must be an env var in the Travis instance I am missing?

adelevie:
Are you using the Vagrant, Docker, or some other local setup?

scottt2:
Vagrant. I followed the README to the T to try an avoid these issues. They still found me...

scottt2:
Morning, folks!

Friendly ping @mtorres253 @adelevie @joshuabailes 

Please lemme know if you have any questions/feedback.

Thanks!

mtorres253:
@scottt2 Thanks! Will look at it today.

scottt2:
Right on! No rush. 

-Tyler

mtorres253:
Hi @scottt2 So close! But there is a problem documented below:

Steps to duplicate
1. create project where start and end dates are in the future.
2. try to enter work for current reporting period for that project—you can’t.
3. go to admin panel and edit the project such that the start + end dates include the current reporting period.
4. try to enter work for current reporting period for that project. 

Expected behavior: You should be able to see the project
Actual behavior: You can’t see the project

We would need this case taken care of before we can accept. Should be an easy fix. Let me know if you have any questions.

scottt2:
Hey @mtorres253 ! Hmm. So the logic right now is:

```
Q(active=True)
            & Q(
                Q(start_date__lte=rps)
                | Q(
                    Q(start_date__gte=rps)
                    & Q(start_date__lte=datetime.datetime.now().date())
                )
                | Q(start_date__isnull=True)
            )
            & Q(
                Q(end_date__gte=rpe)
                | Q(end_date__isnull=True)
            )
```

Did you happen to test with a project whose start date was less than or equal to today? This catch was to try and prevent users from entering time on a project whose start date has occurred **but** is after the reporting period start date.

If that is why it isn't working, the question becomes do you want it to work that way? It was *not* in the spec. I can easily remove.

scottt2:
Here's an image to explain better than I can:
image

Does that make sense, @mtorres253 ?

mtorres253:
Thanks @scottt2! I'll take a look tomorrow and get back to you.

scottt2:
Cool! I think that's it. We'll see tomorrow. 

Thanks!
-Tyler

scottt2:
@mtorres253 So after merging latest master in to resolve the conflicts, my local environment no longer functions. It appears we are deprecating Vagrant use? Can you confirm.

I will push up my latest changes to resolve the conflicts, but will hold off on redoing my local environment. I didn't account for this time, so I'll have to find some time later today to get everything running locally again.

adelevie:
Not sure when your latest pull/rebase from `master` was, @scottt2, but I can confirm that https://github.com/18F/tock/pull/359 introduced our switch from Vagrant to docker-compose.

mtorres253:
Hi @scottt2, you are right that the goal is to prevent a user from entering time on a project that has not started or has ended. What seems to be happening is that logic works after the project is saved initially, but when the start and end dates are edited so that the start date falls into the reporting period, the logic fails. I will give you actual values so that you can validate.

scottt2:
Hey @adelevie ! I literally just did it. And now it seems the Travis tests are failing.

image

Is there something on my end I need to do?

scottt2:
@mtorres253 Interesting. So the update in Admin appears to not be reflected in the app? Hmm. I can check this out in a bit after I get the local environment working again.

scottt2:
@mtorres253 I can't seem to reproduce this.

Here is what I did (lemme know if I totally missed your use case):

1. Create a reporting period for June 15th to June 30th. 
2. Create a project whose start date is July 1st and end date is July 31st.
3. Go to Tock and verify that the project is not in the drop down.
4. Edit the project in /admin to have a start date of June 1st.
5. Go to Tock and verify that the project **is** in the drop down.

mtorres253:
OK, I have clearer steps to duplicate, @scott2! 

@batemapf ran your branch locally and created this case

Case 1:
1. Create a project and set the start date to 5/1/16 and the end date to 6/7/16
2. Try to report time for the time period 6/5/16 - 6/11/16
Expected behavior: You should see the Project A in the dropdown (6/7/16 is within the reporting time period)
Actual behavior: Patrick doesn't see it as an option in the dropdown

screen shot 2016-06-17 at 10 26 12 am
screen shot 2016-06-17 at 10 26 36 am

scottt2:
@mtorres253 Awesome (well, not really--there is a definite bug)! I can repro and will get a fix shortly.

Thanks!

scottt2:
@mtorres253 So the reason for that was due to my faulty logic on the `end_date`. As I was thinking through this, it seemed the easiest fix would be to remove that logic. It really serves no purpose due to the logic on the `start_date`. I just overthought it.

Lemme know how this version works for you guys.

Thanks!

scottt2:
Pushing up one more change...

mtorres253:
Thanks so much @scottt2! We'll try again

scottt2:
@mtorres253 There we go. This is the final bit to disallow any project whose end date is prior to the start date of the reporting period.

scottt2:
image

image

mtorres253:
Looks good @scott2. @batemapf can you re-test and give the thumbs up?

batemapf:
@mtorres253 @scottt2 just ran this through the paces, including the start / end date in middle of a reporting period scenario that was giving us trouble earlier. good to go from my end!

mtorres253:
Awesome! thanks everyone @scottt2 @batemapf @joshuabailes @adelevie @patvin80

PR #375, opened 2016-06-15
Don't load debug toolbar during unit tests.

toolness:
This is actually a PR for #370 and makes `manage.py test` work without needing to explicitly pass-in a different settings configuration.


ccostino:
Awesome, thank you @toolness!

PR #377, opened 2016-06-17
Several fixes to docker nginx config

toolness:
This is an attempt to fix a bunch of [funky](https://github.com/18F/tock/commit/2c9e91b8c0653d19d8fbf3d067786dc944783283#commitcomment-17915616) [issues](https://github.com/18F/tock/pull/357/files/944772690f6f2771dc9c7c1661157853b3bac3c1#r67458243) with the nginx docker config.

These changes make it so that:

* Visiting `http://localhost:8000` on an ubuntu system works as expected.
* POST requests to the Django admin UI work as expected.
* Developers don't need to manually modify the repository's nginx configuration file.
* The email address of the currently logged-in user can be changed by overriding an environment variable in the docker configuration (e.g. via `docker-compose.override.yml`).

This is mostly done by [using environment variables in nginx configuration](https://github.com/docker-library/docs/tree/master/nginx#using-environment-variables-in-nginx-configuration).


docker-compose.yml
```diff
@@ -11,20 +11,25 @@ app:
     - DATABASE_URL=postgres://tock_user@db/tock
     - RUNNING_IN_DOCKER=yup
     - DJANGO_SETTINGS_MODULE=tock.settings.dev
-  command: "python manage.py runserver 0.0.0.0:${EXTERNAL_PORT}"
+  command: "python manage.py runserver 0.0.0.0:1234"
```
toolness: So this fixes a bug introduced in 267344707ed8275df1588d75c6084ad28eac60a0, which is my fault because I decided to make Django's runserver run on the same internal container port as the default `EXPOSED_PORT` that nginx is ultimately listening on.

To reduce the possibility of confusion, I've changed the Django runserver port to be 1234.

tock/tock/settings/dev.py
```diff
@@ -54,6 +54,8 @@ def show_django_debug_toolbar(request):
     'SHOW_TOOLBAR_CALLBACK': 'tock.settings.dev.show_django_debug_toolbar',
 }
 
+USE_X_FORWARDED_HOST = True
```
toolness: This is required for Django to think that the server it's running on is where Nginx is, rather than where Django runserver is running.

If folks are developing the site on something *other* than Docker though, we may want to put this behind an `if` statement of some sort that only gets executed when on Docker.

ccostino:
Awesome, thanks @toolness, I'll help take a look soon!

docker-compose.yml
```diff
@@ -11,20 +11,25 @@ app:
...
 db:
   image: postgres:9.4
   environment:
     - POSTGRES_DB=tock
     - POSTGRES_USER=tock_user
 web:
-  image: nginx
+  image: nginx:1.11
```
ccostino: If the version isn't specified I assume it pulls the latest image, and I take it this change is just to account for the fact that we always know what version we're running (much like any other project dependencies)?

docker-compose.yml
```diff
@@ -11,20 +11,25 @@ app:
...
   volumes:
-    - ./provision/dev/nginx_site_docker.conf:/etc/nginx/conf.d/nginx_site_docker.conf
+    - ./provision/dev/nginx_site_docker.template:/etc/nginx/conf.d/nginx_site_docker.template
+  command: /bin/bash -c "envsubst '$$EXTERNAL_PORT $$DJANGO_SERVER $$USER_EMAIL' < /etc/nginx/conf.d/nginx_site_docker.template > /etc/nginx/conf.d/nginx_site_docker.conf && exec nginx -g 'daemon off;'"
```
ccostino: Oh awesome, you got this to work!

tock/tock/settings/dev.py
ccostino: I think it's fine to move forward with this as it is.  The instructions are written so that you get yourself setup with Docker if you haven't already.

ccostino:
This is great, thank you @toolness!  The only thing other thing that comes to mind is if any README updates are needed to go along with these changes, but I don't think that's the case; I just wanted to make sure.

docker-compose.yml
toolness: Yep!  Specificaly, the `envsubst` command used to fill-in the values for the environment variables in the config template is only present on more recent builds/versions of the nginx container.

toolness:
Awesome, thanks for the review @ccostino !  Nope, no readme changes necessary.

ccostino:
@batemapf, this is good to go!  I'm not sure if you want to merge or I can. :-)

batemapf:
merging!

PR #383, opened 2016-06-22
adds a slimmer bulk timecard endpoint

batemapf:
adds a more svelte option for pull bulk tock data. ran into issues pulling the live data into google sheets off the bulk_timecards.csv endpoint. reduced the number of fields to make the load more economical. if @toolness or @ccostino have a minute or two to take a look, would much appreciate!

test coverage added, as well.

### DIFF
--- a/tock/api/tests.py
+++ b/tock/api/tests.py
@@ -159,6 +159,29 @@ class BulkTimecardsTests(TestCase):
             rows_read += 1
         self.assertNotEqual(rows_read, 0, 'no rows read, expecting 1 or more')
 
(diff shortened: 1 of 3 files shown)


ccostino:
Awesome, I'll try and take a look as soon as I can, @batemapf!

ccostino:
I just had a chance to look at this, @batemapf, and everything looks fine and works well for me!  It'd be awesome to spend a bit more time on this in the future to build a more robust solution to this problem (both in a technical sense and from a user experience sense), but this is simple and gets the job done as well. :-)

PR #386, opened 2016-06-30
Improve project info

batemapf:
~~***WIP: Tests failing locally need to be resolved.***~~

This PR adds two elements to the `projects` model: `agreement_URL` and `project_lead`.

The former is a basic `URLField` that is intended to point to the agreements folder, if available. See @leahbannon's question in #acquisition [earlier today](https://18f.slack.com/archives/acquisition/p1467245679000775) which served as inspiration.

The latter pulls in `User` info via `ForeignKey`, specifically their email address info. Might need to think about this a bit to ensure all users have values in their respective `email` fields.

Both additions to the `projects` model are surfaced via `project_detail.html` (i.e. [like here](https://tock.18f.gov/projects/120)). Also added some existing fields to the page and did some housekeeping to _hopefully_ improve the user experience.


annalee:
paired on the tests with @batemapf and it is good to go 👍

PR #388, opened 2016-06-30
4th of july reminder

MKathrynC:
Added a note to remind people to log 8 hours Out of Office - Administrative/Holiday Leave for Independence Day. @batemapf


batemapf:
thanks @MKathrynC! would you mind adding another commit to this PR that removes the existing announcements (re: the new features)? this will leave the holiday notification as the only announcement.

PR #389, opened 2016-06-30
[WIP] create float app, model and load float data

batemapf:
- create a new app in the tock project that will house task data from Float
- populate `models.py` with matching fields from [here](https://github.com/floatschedule/api/blob/master/Sections/tasks.md)


batemapf:
- add `FloatPeople` model to the Float app. In order to access the `im` field in the Float data (which contains the Tock username within Float), we will need to pull data from the /people endpoint.

batemapf:
latest two commits: 

- adds two scripts: `float_data_for_tock.py`, which pulls data from the `/tasks`Float API endpoint and parses it so it can be accepted as a Django fixture; `get_float_data.sh` which runs the first script (primarily to make sure it is run in Python 2.7). this data is dumped into `json_tock_data.json` which is dropped in the `float` fixtures dir.

- updates `float` app data model to match the output of `get_float_data.sh` (and the underlying `float_data_for_tock.py`script) + migrations

- adds a couple of files to `.gitignore`, most importantly to ignore `json_tock_data.json` to ensure that Float data is not inadvertently pushed to this public repo

cc @ccostino

float_data_for_tock.py
```diff
@@ -0,0 +1,60 @@
...
+import os
+import json
+
+api_key = os.environ.get('FLOAT_API_KEY')
+headers = {'Authorization': 'Bearer ' + api_key}
+endpoint = 'tasks'
+year = "2016"
+month = "04"
+day = "01"
+weeks = "52"
```
batemapf: static values here could be replaced by dynamic variables to generate a smaller data set.

float_data_for_tock.py
```diff
@@ -0,0 +1,60 @@
...
+output = r.content
+
+print "\nData successfully pulled from Float API. \n"
+
+fmt = 'json'
+def writeToFile(x, str_data={}, fmt=fmt):
+    with open("output." + fmt,"w") as fp:
+        fp.write(x)
+
+writeToFile(output)
```
batemapf: this extra step (writing the data to file, then reopening) is the only way i could get the api output to cooperate. assuming this is a less involved way of addressing.

batemapf:
next up: pulling in the `/people` Float endpoint to access the `im` field which aligns with the tock `username`.

float_data_for_tock.py
ccostino: We could set this up so the script or command takes in input for these but also provide sensible defaults (such as calculating and using the current Fiscal Year).

float_data_for_tock.py
ccostino: There has to be, definitely worth looking into.  If we really needed to create a temporary file for this, there's always the [`tempfile`](https://docs.python.org/3/library/tempfile.html) module.

float_data_for_tock.py
```diff
@@ -0,0 +1,60 @@
...
+endpoint = 'tasks'
+year = "2016"
+month = "04"
+day = "01"
+weeks = "52"
+timecard_date = str(year + "-" + month + "-" + day)
+payload = {'start_day': timecard_date, 'weeks': weeks}
+url = 'https://api.floatschedule.com/api/v1/' + endpoint
+r = requests.get(url, headers=headers, params=payload)
+output = r.content
```
ccostino: Actually, to fix this issue, I would think all you need to do is load the json string, e.g., `json.loads(r.content)`.  More info here:  [`json.loads`](https://docs.python.org/3/library/json.html#json.loads).

get_float_data.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+pyenv local 2.7.11
+echo "Running float_data_for_tock.py in Python 2.7.11."
```
ccostino: We need to get you into Python 3! :-)

ccostino:
This is awesome, @batemapf, thanks for taking a crack at it!  Do you mind if I hop in at some point to help out/make improvements?

batemapf:
@ccostino definitely jump in! let me know what areas you are going to tackle so we don't cross work streams. my next focus is on getting the `im` field data incorporated from the `/people` endpoint.

ccostino:
Thanks!  At the moment I'm not sure and I'm happy to let you take this as far as you'd like first, you're making solid progress and I'd hate to disrupt that. :-)

I do see a potentially different approach for running the script to pull and parse Float data that would integrate a little more seamlessly with Tock itself (and set it up for automated scheduling in the future as well), so that is likely the first place I'll be diving into.

batemapf:
**to fetch and load float data:**

1. enter cf-ssh session for app
2. pip install requests
3. set env var for FLOAT_API_KEY
4. run rmt_float_data_for_tock.py
5. python manage.py looddata [out file]

This loads current float data in the FloatTasks class within Tock and uses a `ForeignKey` to link each FloatTask object with a `User` object. That means a user's current / recent Float assignments can be called up via Tock very easily. cc @abrouilette who I know will be happy :-)

If you are interested in viewing results, and have `staff` and `superuser` status on the staging site, head over to https://tock-staging.18f.gov/admin/float/floattasks/. @abrouilette just granted you access since I know you'll want to look.

Plan is automate steps so updates come across on a regular / on-demand basis without having to execute a `cf-ssh` command and run everything manually.

ccostino:
@batemapf That is awesome, way to go!

batemapf:
Oh, and in case you are wondering, this pulls all Float assignments for the last six months and the next six months. This timeframe can be adjusted if need be.

batemapf:
added our first view of the float data within tock! this is live @ tock-staging.18f.gov! go forth and check it out!

screen shot 2016-08-25 at 11 58 57 pm

float_data_for_tock.py
```diff
@@ -0,0 +1,153 @@
...
+"""
+endpoint = 'people'
+r = requests.get(url + endpoint, headers=headers)
+people_data = json.loads(r.content.lower().strip())
+print('\nFetched people items from Float.')
+
+"""
+Clean people data for all users from Float to remove all but req'd fields.
+"""
+clean_data = list()
```
shawnbot: This might be more pythonically written as:

```py
person_keys = ['im', 'people_id']
clean_data = map(
    people_data['people'],
    lambda person: {key: person[key] for key in person_keys}
)

Comment (jmcarp) [inline float_data_for_tock.py]: Not that I get to be the arbiter of what's pythonic, but IMO comprehensions are more idiomatic than `map`, and mixing the two is weird. I would write something like:

```python
clean_data = [
    {'im': person['im'], 'people_id': person['people_id']}
    for person in people_data
]
```

shawnbot: Curse you, @jmcarp! I also just realized that I got the order of arguments to `filter()` wrong 😬

float_data_for_tock.py
```diff
@@ -0,0 +1,153 @@
...
+fixture = list()
+for item in task_list:
+    fixture.append(dict())
+y = len(task_list)
+x = y - 1
+while x >= 0:
+    fixture[x]["pk"] = x
+    fixture[x]["model"] = "float.floattasks"
+    fixture[x]["fields"] = task_list[x]
+    x = x - 1
```
cmc333333: Quick note that while loops are generally frowned on in Python code. Have you seen [`range`](https://docs.python.org/2/library/functions.html#range)?

As order doesn't matter here, the equivalent loop would be
```python
for x in range(y):
    fixture[x]["pk"] = x
    fixture[x]["model"] = "float.floattasks"
    fixture[x]["fields"] = task_list[x]
```


We can do better, though: you can simplify lines 133 through 142 to something like
```python
fixture = []
for idx, task in enumerate(task_list):
    fixture.append({'pk': idx,
                    'model': 'float.floattasks',
                    'fields': task})
```

cmc333333:
@batemapf would you be open to getting this reviewed by #python folks? Looks like there's a bit that could be abstracted/refactored to be more legible

float_data_for_tock.py
jmcarp: Or even

```python
fixture = [
    {'pk': idx, 'model': 'float.floattasks', 'fields': task}
    for idx, task in enumerate(task_list)
]
```

batemapf:
@cmc333333 totally! this is still pretty rough, put it together to test out the float / tock integration concept. my hope is we can build this more seamlessly into tock (right now, needs to be run manually via cf-ssh). that said, i suspect the basic elements will remain somewhat the same, just refactored a bit. my only ask is to be involved in your review. i am very much in learning mode and would love to hear the discussion.

batemapf:
on the topic of integrating more seamlessly into tock, how about issuing a float data request on creation of a new `ReportingPeriod`? the call could request float data for that period only, as opposed to _all_ of the data like it is doing now. even better, with it tied to existing, consistent user behavior (the weekly job of creating a new reporting period), data is fetched in a much more "automatic" fashion.

alternatively, the float request could be issued upon creation of a new `User` and `ReportingPeriod` specific `Timecard`. this could limit the request to _just_ the float data for the corresponding period and user. of course, this means the initial load of timecard_form.html for each reporting period would be delayed by however long the api call takes (right now, it takes at least five seconds to fetch and load the full data set manually).

cc @annalee @shawnbot @abrouilette

float_data_for_tock.py
```diff
@@ -0,0 +1,153 @@
+import requests
+import os
+import json
+from datetime import datetime, timedelta
+
+message = '\nResults of operation:'
+print(message)
+print(len(message) * '=')
```
adelevie: niiice

batemapf:
closing in favor of #504

PR #395, opened 2016-07-08
adding min max feature

batemapf:
The purpose of this PR is to switch Tock from requiring an exact number of hours to be reported per period to requiring a range of hours (`min_working_hours` and `max_working_hours`). This will allow for reporting of hours over the typical 40 limit. The minimum and maximums may be varied across each `ReportingPeriod`.

Specifically, this PR:

- adds `min_working_hours` and `max_working_hours` to `models.py` in the hours app;
- updates fields for the /api/reporting_period_audit/ endpoint;
- makes changes to `forms.py` in the hours app that pull the `min_working_hours` and `max_working_hours` data from the DB and load them as variables for use in subsequent logic;
- updated form validation logic to verify that total hours submitted are within the min/max range, but only when the user attempts to submit (as opposed to just save);
- adds validation into the admin panel for minimum and maximums so rules are enforced at the superuser level as well;
- made necessary updates to the hours app `views.py` file to line up with `forms.py`
- updated and expanded tests of hours app
- made some UX updates, including color changes, alert placement modifications, and notification to the user on what the min/max is for the reporting period;
- set default min to 40 hours, max to 60 hours
- maintained the existing `working_hours` as `exact_working_hours`, but can probably be excised at some point
- migrations
- other miscellaneous things like documentation enhancements


batemapf:
^^ these changes are currently live at tock-staging.18f.gov

tock/hours/forms.py
```diff
@@ -15,17 +15,20 @@
 from datetime import datetime, timedelta, time
 
 class ReportingPeriodForm(forms.ModelForm):
+    """Form for creating new reporting periods"""
 
     class Meta:
         model = ReportingPeriod
-        fields = ['start_date', 'end_date', 'working_hours', 'message']
+        fields = ['start_date', 'end_date', 'min_working_hours', 'max_working_hours',]
```
ccostino: No more need for the `message` as a part of this list?

tock/hours/models.py
```diff
@@ -10,11 +10,13 @@
...
 
 class ReportingPeriod(ValidateOnSaveMixin, models.Model):
+    """Reporting period model details"""
     start_date = models.DateField(unique=True)
-    end_date = models.DateField()
-    working_hours = models.PositiveSmallIntegerField(
-        default=40,
-        validators=[MaxValueValidator(80)])
+    end_date = models.DateField(unique=True)
+    exact_working_hours = models.PositiveSmallIntegerField(
```
ccostino: Is there a corresponding data migration with this change?  Meaning, for all records we have that were stored in the `working_hours` property, is there a migration to shift that data over into this new `exact_working_hours` field?  I'm not seeing one above so I wanted to double check.  If there's not, we need to add one so we do not lose existing data in the system.

tock/hours/forms.py
batemapf: we never completed linking the `messages` field in the `hours` model, so entries in this form don't actually show up anywhere. doesn't mean we shouldn't fix this at some point, just wanted to clear up confusion for persons creating new reporting periods.

tock/hours/migrations/0018_auto_20160708_0207.py
```diff
@@ -0,0 +1,19 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('hours', '0017_auto_20160708_0156'),
+    ]
+
+    operations = [
+        migrations.RenameField(
+            model_name='reportingperiod',
+            old_name='working_hours',
```
batemapf: @ccostino this should take care of renaming `working_hours` to `exact_working_hours`, right?

tock/hours/tests/test_forms.py
```diff
@@ -164,16 +172,25 @@ def test_timecard_inline_formset_save_only(self):
         formset.save_only = True
         self.assertTrue(formset.save_only)
 
-    def test_timecard_is_not_100(self):
-        """ Test timecard form data that doesn't total the set working
-        hours """
+    def test_timecard_is_too_big(self):
+        """ Test timecard form data that exceeds the maximum hours allowed """
```
ccostino: It looks like these hours only add up to `59` in this test.  Shouldn't it be some value over `60` hours?

tock/hours/models.py
batemapf: dropped a note [here](https://github.com/18F/tock/pull/395/files#r70126028)

tock/hours/tests/test_forms.py
batemapf: i am seeing a total of 92 actually. see ln 151 (22 hours) + ln 153 (20 hours) + ln 179 (50 hours). the `test_timecard_is_too_big` function pulls the base data from `setUp` (which holds two time card objects) and then adds an additional time card object to it, for a total of three time card objects (see ln 180). this make sense?

tock/hours/tests/test_forms.py
```diff
@@ -194,34 +211,55 @@ def test_timecard_has_0_hours_for_project(self):
...
         form_data['timecardobject_set-1-hours_spent'] = '5'
         formset = TimecardFormSet(form_data)
-        formset.set_working_hours(16)
+        formset.set_min_working_hours(16)
         self.assertFalse(formset.is_valid())
 
-    def test_reporting_period_with_less_than_40_hours_success_save_mode(self):
+    def test_reporting_period_with_more_than_max_hours_success(self):
+        """ Test the timecard form when the reporting period requires no more
+        than 60 hours to be reported and the hours entered are more than 60"""
```
ccostino: This is testing less than `60` hours, correct?

tock/hours/views.py
```diff
@@ -138,12 +138,20 @@ def get_object(self, queryset=None):
     def get_context_data(self, **kwargs):
         context = super(TimecardView, self).get_context_data(**kwargs)
 
-        working_hours = ReportingPeriod.objects.get(
```
ccostino: Right above here, we could probably set another variable, `base_reporting_period`, and set it equal to `ReportingPeriod.objects.get(start_date=self.kwargs['reporting_period'])` to cut down on the number of database calls here.  In fact, we could just set that instead and then reference it down below in lines 152 - 154, e.g.:

```
formset.set_exact_working_hours(base_reporting_period.exact_working_hours)
formset.set_max_working_hours(base_reporting_period.max_working_hours)
formset.set_min_working_hours(base_reporting_period.min_working_hours)
```

tock/hours/views.py
```diff
@@ -162,7 +170,9 @@ def get_context_data(self, **kwargs):
             formset.save_only = True
 
         context.update({
-            'working_hours': working_hours,
+            'exact_working_hours': exact_working_hours,
+            'min_working_hours': min_working_hours,
+            'max_working_hours': max_working_hours,
```
ccostino: We could reuse the variable noted above here too in retrieving these values:

```
'exact_working_hours': base_reporting_period.exact_working_hours,
'min_working_hours': base_reporting_period.min_working_hours,
'max_working_hours': base_reporting_period.max_working_hours,
```

tock/hours/tests/test_forms.py
batemapf: correct. need to change the comment :-/

tock/hours/views.py
batemapf: ah very good point, just made the edits in the next commit.

tock/hours/migrations/0018_auto_20160708_0207.py
ccostino: Ah okay, yes, presumably it should!  Would be good to confirm with a test. :-)

tock/hours/tests/test_forms.py
ccostino: Ah okay, I missed the base data in the `setUp` method, thanks for pointing that out!

tock/hours/migrations/0018_auto_20160708_0207.py
ccostino: Back in the day it used to be things like this required an extra data migration, so you'd have the initial schema migration to create the new field, a data migration to run immediately after to shift the data from the old place to the new, and then another schema migration to remove the old field.  Looks like this is improved/simplified now, which is awesome.

ccostino:
One rather big thing with this change is that the `test_data/data-update.json`... data... is now out of date with this change because it references the old `working_hours` field name instead of the new `exact_working_hours` name.  If at all possible, all of those field name references should be updated as a part of this PR (a find/replace on the file should do the trick).

batemapf:
ok @ccostino things should be updated per your line notes. did a find and replace (learned how to `sed` with help from @konklone!) on the test data to replace `working_hours` with `exact_working_hours`, as well.

thanks again for looking over this!

ccostino:
Awesome, well done with using `sed`! :-)

I just pulled this down and tried things out, I think you're good to go at this point.  You're welcome, @batemapf, always happy to help!

Issue #367, opened 2016-06-10
search not returning results from multi-word queries


zacharycohn:
When I search for "child," a bunch of entries appear for " Acquisition Services / State Child Welfare / ...".

But if I search for "child welfare" or "state child welfare", there are no results.

carodew:
This also caused me grief today.

zacharycohn:
Had this problem again when searching for "Out of office", which returns nothing in the search results.

ccostino:
All fixed as of PRs #662 and #664!

PR #398, opened 2016-07-11
match search text within dropdown options text

jseppi:
- add `search_within: true` to chosen() instantiations
- simplify calls to chosen() instantiation and event hookups

Noticed today that I was unable to just type "holiday" to find the "Administrative/Holiday Leave" project in tock. Enabling the `search_within` option in chosen selects enables searching to work within full strings instead of on space-delimited tokens.

screen shot 2016-07-11 at 8 39 12 am

### DIFF
--- a/tock/tock/templates/hours/timecard_form.html
+++ b/tock/tock/templates/hours/timecard_form.html
@@ -164,6 +164,10 @@
 
 
 $( document ).ready(function() {
+    var chosenOptions = {
+      search_contains: true
+    };
+
     $("#save-timecard").on("click", function() {
       var form = $('form'),
           save_input = '<input type="hidden" name="save_only" value="1"/>';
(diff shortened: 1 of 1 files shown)


toolness:
Just tried this out and it's great!  Not only does it work in the case @jseppi mentioned, but it also allows us to type strings with spaces in them; for example, @zacharycohn mentioned on Slack yesterday that typing "out of office" doesn't work, even though there are literally tock projects that start with the words "Out of Office". This PR fixes that issue too!

I can't speak to the event hookups changes in this PR, but the new chosen options are definitely full of win. (I'm not sure why chosen doesn't just work this way by default, but oh well.)

toolness:
Er, in other words, this fixes #367 too. :)

ccostino:
Thank you very much, @jseppi! 🎉

zacharycohn:
It seems like this issue is still occuring. I tried to search for "Mississippi" today to return "Acquisition Services / States Child Welfare / Mississippi," and it found nothing. I also tried searching for "Child" and "Welfare" with no results. I tried "Acquisition Services" and it was returning results until I started typing "Services," when it started returning no results again.

jseppi:
Hrm, not sure what is going on. I just tried searching "Mississippi" and it did bring up "Acquisition Services / States Child Welfare / Mississippi." "Child" and "Welfare" also brought up a bunch of matches. 

screen shot 2016-07-18 at 10 44 17 am

screen shot 2016-07-18 at 10 44 35 am

Could you try hard-refreshing the page (command+shift+r in Mac Chrome) and trying it again?

zacharycohn:
Did you try a multi-word search? (Like "Child Welfare")

jseppi:
@zacharycohn you're right, it is not working for a multi-word search like "child welfare"

screen shot 2016-07-18 at 10 54 36 am

PR #408, opened 2016-07-25
Remove July 4th message, reinstate previous one(s)

fureigh:
This PR:
 * Removes the Independence-Day-related language
 * Reinstates the messages that existed prior to #396


batemapf:
see #409

PR #409, opened 2016-07-28
update message

batemapf:
update message


batemapf:
updates message, also closes #408

PR #412, opened 2016-08-04
[WIP] Automatically deactivate projects based on date

batemapf:
This PR attempts to establish an automatic project deactivation function based on the project's stated `end_date` and the new `Project` model attribute of `auto_deactivate_days`, which are both used to calculate `auto_deactivate_date`. This `auto_deactivate_date` value is then evaluated against datetime.today() via a save() function within the `Project` model to determine whether the `active` attribute should be True or False.
- [x] **TODO**: Whether or not a project is `active` is only contingent on the evaluation `auto_deactivate_date` upon save(), and no other action. Therefore, in order for a project to be automatically deactivated, a user would need to save() values to the `Project` model to trigger the evaluation and resultant change of `active` from True to False. 
- [x] _Instead_, either (1) the evaluation should be triggered by another, more passive action, such as creating a new reporting period or calling project values for timecard_form.html; or (2) projects/views.py and/or hours/timecard_form.html does not display projects that are either `active` = False _or_ projects that have an `auto_deactivate_date` that is less than or equal to datetime.today() _or_ projects that have an `end_date` that is less than or equal to datetime.today().

Thinking the latter option is simplest and most economical...

cc @joshuabailes who will hopefully be excited that Tock will soon include a business rule that prevents projects from remaining active beyond their `end_date`.


batemapf:
Latest commit updates `get_projects` function (which uses Q objects to filter out projects that have "ended" or have a False `active` attribute from the list of projects that the user sees in `timecard_form.html`. The updates
- adds an filter that prevents projects that are past their `auto_deactivate_date`;
- provides some basic in-line documentation for the filter; and
- adds a test for the `auto_deactivate_date` attribute.

**TODO**:
- [ ] Add `max_hours` attribute to the `Projects` model that allows the user to establish the maximum number of hours that may be logged to a project;
- [ ] Add `max_hours_restriction` Boolean attribute that allows the user to impose (True) or not (False) the `max_hours` attribute; 
- [ ] Further revise the `get_projects` function to filter out projects where: (1) the `max_hours_restriction` attribute is True; and (2) the aggregate hours worked on the project is greater than or equal to `max_hours`.

abrouilette:
Closed without merging because of questionable legality/merit.   Will address another way.

PR #420, opened 2016-08-16
add service info for staging instance

batemapf:
ensures that the tock-staging app automatically binds to staging DB


batemapf:
deployed.

PR #326, opened 2016-03-04
Travel Form and Email Request Automation

dsiddy:
This pull request fulfills the [Travel Form and Email Request Automation](https://micropurchase.18f.gov/auctions/14) micro-purchase.  Please see the commit history for detailed information.  All new code adheres to [PEP 8](https://www.python.org/dev/peps/pep-0008/).

Please let me know if you have any questions &mdash; and thanks again for this opportunity!

PR #425, opened 2016-08-30
Travel Form and Email Request Automation - merge conflicts fixed

ccostino:
Hi folks,

I've fixed the merge conflicts with #326 but sadly had to do it as a new PR.  The old PR can be closed, and once someone has a chance to do a final review, we should be able to merge this in without an issue and move forward finally.

Thank you!

/cc @mtorres253, @adelevie, @annalee, @batemapf, @ethanheppner, @sfvictor


mtorres253:
Thanks @ccostino!

batemapf:
took a closer look at this and there are a number of things that should be addressed before merging:

1. the `tock_project_name` `requestor_name` and `tock_project_id` (at a minimum) should have ForeignKey relationships with the `projects` app (see `Project` class) and the built in `User` class. right now they call for string input that is not clearly validated against what does exist in those data tables.

2. the functionality is accessed through a hyperlink at the bottom of an individual user's "profile" page  (i.e. `/employes/patrick.bateman`), making it very hard to find. at a minimum, it should appear in the left-hand menubar in base.html.

ccostino:
@batemapf thanks for reviewing!  I or someone will need to take a closer look at the first issue, but the second one is easily fixed. :-)

batemapf:
in addition, this looks to use `EmailMultiAlternatives` to actually send mail from the application via server settings in dev.py. i suspect that in order for us to utilize the agency's email service (currently Google Mail) we would need to get authorization from our cio shop. the (low tech) solution is to just generate a text or html email in the browser which is then copied and pasted into the user's Mail app and sent manually...

abrouilette:
Closed without merging.  This seems a C2 feature...?  FYI @ccostino

ccostino:
Ah, okay, thanks for the heads up @abrouilette!

PR #431, opened 2016-09-03
Update all Tock project dependencies

### LINKED ISSUE #416: update django

ccostino:
Fixes #416

This changeset updates all of Tock's project dependencies to their latest versions.  This includes the following changes:

* Updates the `requirements.txt` and `requirements-dev.txt` files to account for all package updates
* Fixes line endings in `requirements-dev.txt`
* Updates the `Dockerfile` and `docker-compose.yml` setup to include the latest stable versions of Python, PostgreSQL, and nginx
* Addresses several deprecation warnings that appeared in the tests due to the updated packages
* Updates and fixes the Travis config

While fixing the remaining Django issues (and I was wrong, `1.10.1` is the latest version!), I noticed that Python 3.4 was still being used by the Docker container so I went through and ended up taking care of all of the remaining project dependencies

A HUGE hat tip (h/t) to @batemapf for doing all of the initial legwork with this and addressing a lot of the upgrade issues!  This only carries the work over the finish line.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,10 @@ language: python
 sudo: false
 
 python:
-  - "3.5.2"
+  - "3.5"
 
 addons:
-  postgresql: "9.5.4"
+  postgresql: "9.5"
 
 cache: pip
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,10 @@ language: python
 sudo: false
 
 python:
-  - "3.4"
+  - "3.5.2"
 
 addons:
-  postgresql: "9.4"
+  postgresql: "9.5.4"
 
 cache: pip
 
@@ -20,7 +20,7 @@ install:
 
 script:
   - cd tock
-  - python manage.py syncdb --noinput --settings=tock.settings.test
+  - python manage.py migrate --noinput --settings=tock.settings.test
   - python manage.py test --noinput --settings=tock.settings.test
 
 after_success:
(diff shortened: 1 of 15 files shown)


ccostino:
The only thing I was unable to address was one unicode warning while running the tests - I'm not sure where exactly it's coming from, but it's definitely not a show-stopper either!

.travis.yml
```diff
@@ -1,16 +1,15 @@
-language: python
+sudo: required
```
jmcarp: y sudo?

ccostino: While attempting to match the Python and PostgreSQL versions of the app in the CI test environment I realized that the default Travis environment [didn't support PostgreSQL 9.5.x](https://docs.travis-ci.com/user/database-setup/#Using-a-different-PostgreSQL-Version).  Looking into it further yielded their [Trusty Environment])https://docs.travis-ci.com/user/trusty-ci-environment/), which does support it but requires sudo apparently.

I was a bit torn on this given that this is also a beta environment according to their documentation, so if you're aware of being able to use 9.5.x differently in Travis I'd be very curious to know (if you look at the commit history you'll see I tried a few different things to get it to run correctly).  Alternatively, I'm open to other suggestions here as well if you (or anyone else!) feel there's a better way to achieve parity between the project setup and Travis.

annalee:
:shipit:

PR #437, opened 2016-09-07
Use node-sass, rename stylesheets

### LINKED ISSUE #436: Replace Ruby Sass with Node implementation
[node-sass] is markedly faster than the Ruby implementation, and the transition over to Node will make it much easier to incorporate other modern front end build tools, such as module- and browser-aware JavaScript bundling tools.

[node-sass]: https://github.com/sass/node-sass

shawnbot:
This switches the `sass` app over from the Ruby implementation to the Node one, and is a step toward replacing the collection of individual JS files with a single browser bundle.

Fixes #436.

Changes:

* Remove the Ruby installation from `Dockerfile` and simplifies the `pip install` command by adding a `WORKDIR` directive
* Add Node-specific files to the `.gitignore`
* Rename the main stylesheet (in both the Sass source and CSS output) to `main`
* Add `package.json` that declares the [node-sass] dependency and `scripts` entries for building and watching the stylesheets
* Update the `sass` app in `docker-compose.yml` to use Node v6.5 instead of the app (`.`) build and run `npm start` as the entrypoint

This could use a review from @annalee and/or @toolness and @jseppi (thanks for the pointers, y'all!)

[node-sass]: https://github.com/sass/node-sass


jseppi:
I pulled this code and got it up and running with `docker-compose` by following the README. The node sass part seems to work as intended! 👍 

I generally don't like to check-in built assets (like the `main.scss`) and instead build those files locally or in CI or upon deploy, but that preference is probably debatable and could be handled in a separate PR anyway.

I also noticed this funny path: [tock/tock/tock/tock/static/css](https://github.com/18F/tock/tree/node-sass/tock/tock/tock/tock/static/css) which is probably something that should be in a separate issue as well.

batemapf:
@shawnbot I believe we discussed in Slack --- this one is ready to go?  cc @annalee @abrouilette

Issue #424, opened 2016-08-26
Negative Hours


thecapacity:
Likely hours shouldn't be negative but I found myself in that situation (luckily I caught it before submitting). I believe I: Clicked on the page and tried to scroll (with the touchpad) - probably in clicking I clicked in an hours data-entry box - so when I tried scrolling the page up/down it started making my hours positive/negative.

batemapf:
@ccostino sounds like another form validation project!

ccostino:
Sounds like an odd client-side bug, too. :-)

thecapacity:
I believe an appropriate fix (based on the generated page I see) would be to change (whatever generates);
`<input id="id_timecardobject_set-0-hours_spent" name="timecardobject_set-0-hours_spent" step="0.01" type="number" />`

to
`<input id="id_timecardobject_set-0-hours_spent" name="timecardobject_set-0-hours_spent" step="0.01" type="number", min="0"`>`

i.e. add the HTML form specifier; `min="0"`

Unfortunately, I was not able to find where in the code these are generated.

Interesting to note this (non-active code) from: `tock/hours/forms.py`:
```
# For another day :-). Changes step increment in Hours Spent field.
#    )
#    hours_spent = forms.IntegerField(
#        widget=forms.NumberInput(attrs={'step': '0.25'})
#    )
```

Note the step is `.25`, which is different than the current `.01`

Also the fix may include updating the serializer for `DecimalField(...)` to include `max_value` and `min_value`

Sorry I couldn't submit a PR too!

PR #438, opened 2016-09-07
Require hours input to have value >= 0

shawnbot:
This is a fix for #424.

~~In my ad-hoc testing, it works great. Unfortunately, though, this fails the unit test suite in a couple of places, which you can see by clicking the ❌ icons next to the commits below. I'm out of my element with Django forms, so maybe somebody else could take a look at this? You can run the test suite on the docker container with:~~

```sh
docker-compose run app python /tock/manage.py test --noinput --settings=tock.settings.test
```

### TODO
- [x] Get the tests passing
- [x] ~~Update the input `step` attribute to `0.1` per @HuixianXu's recommendation~~


batemapf:
removed max value, i believe it makes more sense to validate that via the overall per `ReportingPeriod` hourly limit over in [hours.models.ReportingPeriod](https://github.com/18F/tock/blob/master/tock/hours/models.py#L18) via the max_hours attribute.

the failing test [here](https://github.com/18F/tock/blob/master/tock/hours/tests/test_forms.py#L104) looks like a failure of the form to be created due to `None` being non-compliant with the NumberInput widget added in this change. the test seeks to validate the the behavior of [clean_hours_spent](https://github.com/18F/tock/blob/master/tock/hours/forms.py#L180), which replaces any None types with int(0)'s. this clean method _may_ no longer be necessary (and the accompanying test) if we put our trust in the NumberInput widget to reject values that violate its attributes. if so, the clean_hours_spent method could be removed (along with the test) to tidy things up. 

side note: even if a None got successfully submitted, the DB would accept it as the hours_spent attribute of the TimecardObject model accepts null (None) values.

batemapf:
latest commit takes a different approach to solving the test failures: marking the `hours_spent` field as not required. definitely unorthodox, but the `clean_hours_spent` method should take care of any submissions that do not hold values. the commit also rolls back the form field type from `FloatField` to `IntegerField` which was something i was messing around with previously and forgot to roll back on prior commit 😨 .

if this change is acceptable, i think we are good to go. of course, there are probably additional things we can do to ensure the values submitted via the form have the correct increments (an additional clean method, maybe?). however, since this is more about preventing user error rather than getting a specific increment into the DB, probably not necessary.

batemapf:
spoke too soon 😞 . this is now creating some weird behavior in the form clean() methods. back to you @shawnbot, and feel free to roll back my last commit if needed.

abrouilette:
@shawnbot, wouldja please make sure this one gets closed out before you transition off?

abrouilette:
@shawnbot, @maya, is this going to be a change we should communicate to the team?

batemapf:
@shawnbot @maya i take it this is ready for merge + deploy? need any additional 👁 's on it?

PR #441, opened 2016-09-09
Show user report links to all users

### LINKED ISSUE #218: enable anyone to view teammates hour visualizations
Right now, I can see anyone's hours but I can't see the combined visualization (like the below) for teammates.  There's no new information available through those, but they are very useful.  

It would be useful to somehow enable everyone to see these.  Right now, the page they are on are blocked to non-admins since there are editable fields on that page, but two solutions could be: 
* make copies of them available on another page (combined or individually) 
* move them to a different page (a la https://tock.18f.gov/employees/richard.g.brooks/chart) 

screen shot 2015-07-28 at 1 56 12 pm

shawnbot:
- [x] Fix #218
- [x] Get tests passing again

### Changes
- Remove the condition in the user list template that only showed links if the the viewer is a super-user.
- Remove stray `Vagrantfile` in the Sass directory


hollyallen:
:happy dance:

PR #445, opened 2016-09-12
delinate bw saved and submitted hours in view

batemapf:
indicates which hours on project_detail.html have been submitted versus saved only and a total of both.


batemapf:
- adds custom save method TimecardObject that inherits the `submitted` value from the related Timecard

- uses inherited `submitted` attribute to total hours in projects.views and displays result in project_detail.html (in verrrrrrry basic fashion)

PR #447, opened 2016-09-13
Revert "Use node-sass, rename stylesheets"

batemapf:
Reverts 18F/tock#437


batemapf:
pushed to staging, runs fine.

PR #457, opened 2016-09-16
update nginx template to match new auth scheme

batemapf:


annalee:
Paired on this with @batemapf and it is good to go.

Issue #451, opened 2016-09-15
Staff grade included in bulk data. 


HuixianXu:
As a finance team member, I would like to request Tock bulk data include staff GS grade for each week. With this feature, labor rate can be accurately set to reflect GS level change from promotions or other personnel changes.

batemapf:
closed via #461

PR #461, opened 2016-09-19
Add grade info

batemapf:
This pull request would add employee grade information to Tock and addresses #451 as follows:

- A new `EmployeeGrade` class has been added to the employees app. Each instance of `EmployeeGrade` stores one or more sets of the following grade information for each employee:
 - grade, based on predefined choices; and
 - g_start_date, which is the date in which the grade comes into force. 
- The `TimecardObject` class has a new custom save() method that fetches the applicable grade information related to the employee submitting the timecard based on the end date of the reporting period and the most recent g_start_date. For employees with multiple `EmployeeGrade` instances, this ensures that only the most recent grade (i.e. if the employee recently got a promotion) is appended to the `TimecardObject`.
- A new API serializer, ` AdminBulkTimecardSerializer`, was created to deliver bulk timecard info to the end user with grade information attached. The view function associated with this serializer includes a `@user_passes_test` decorator which restricts access to Tock superusers only. The purpose of this is to limit the availability of employee-level grade information and protect the privacy of our employees. A new URL was created for end user access.
- Testing for all new functions (aside from string method overrides) is included.
- The project-wide fixtures file was updated to include new `EmployeeGrade` class and related changes.

TODO:
- [x] Correct API output format for grade. Currently delivers the the string representation of the related `EmployeeGrade` instance, which has more information than just a numerical grade.


batemapf:
okeydokey @annalee i tidied this one up a little. i believe it is ready for 👀!

tock/hours/models.py
```diff
@@ -108,6 +110,21 @@ def notes_list(self):
...
+        emp_grd_objects = EmployeeGrade.objects.filter(
+                Q(employee = self.timecard.user)
+                & Q(g_start_date__lte = self.timecard.reporting_period.end_date)
+                ).all().aggregate(Max('g_start_date'))['g_start_date__max']
+
+        if emp_grd_objects:
+            self.grade = EmployeeGrade.objects.filter(
+                employee=self.timecard.user,
+                g_start_date=emp_grd_objects)[0]
+
```
annalee: The code to figure out which grade applies to a given employee should live in a model method on UserData--I'd recommend setting it to default to their current grade, with optional arguments to give their grade as of a specific date. Then call that method from here to figure out what grade applies to the relevant time card.

The logic here is that it makes that code more accessible if we need it in the future, and it makes it easy to access a fact associated with the employee (their grade at a given time) from the employee's userdata object, where we're used to finding that info.

tock/hours/views.py
```diff
@@ -59,6 +60,21 @@ class SlimBulkTimecardSerializer(serializers.Serializer):
...
+    end_date = serializers.DateField(source='timecard.reporting_period.end_date')
+    hours_spent = serializers.DecimalField(max_digits=5, decimal_places=2)
+    billable = serializers.BooleanField(source='project.accounting_code.billable')
+    agency = serializers.CharField(source='project.accounting_code.agency.name')
+    flat_rate = serializers.BooleanField(source='project.accounting_code.flat_rate')
+    active = serializers.BooleanField(source='project.active')
+    mbnumber = serializers.CharField(source='project.mbnumber')
+    notes = serializers.CharField()
+    grade = serializers.CharField(source='grade.grade')
+
```
annalee: Are our bulktimecardserializer changes meant to be in this PR? or just the parts of them that add grade?

tock/employees/admin.py
```diff
@@ -1,11 +1,16 @@
...
-class EmployeeDataAdmin(admin.ModelAdmin):
+class UserDataAdmin(admin.ModelAdmin):
     list_display = ('user',)
     search_fields = ('user__username',)
 
 
-admin.site.register(UserData)
+class EmployeeGradeAdmin(admin.ModelAdmin):
+    pass
+
```
batemapf: No real need, mostly as a placeholder in case we wanted to customize the admin interface for EmployeeGrade. Happy to excise it as dead code.

tock/employees/models.py
```diff
@@ -2,6 +2,42 @@
...
+        (8, '8'),
+        (9, '9'),
+        (10, '10'),
+        (11, '11'),
+        (12, '12'),
+        (13, '13'),
+        (14, '14'),
+        (15, '15'),
+        (16, 'SES')
+    )
```
batemapf: Hehe ⛳

tock/hours/views.py
batemapf: Just the parts that add the grade.

tock/hours/models.py
batemapf: Definitely like the logic. Addressed in https://github.com/18F/tock/pull/461/commits/e4698e38888d7844476ee06106b6a4b06e81d5e2 !

tock/api/urls.py
```diff
@@ -23,5 +23,6 @@
     url(r'^hours/by_quarter_by_user.json$', views.hours_by_quarter_by_user, name='HoursByQuarterByUser'),
     url(r'^slim_timecard_bulk.csv$', views.slim_bulk_timecard_list, name='SlimBulkTimecardList'),
     url(r'^user_data.(?P<format>csv|json)$', views.UserDataView.as_view(), name='UserDataView'),
+    url(r'^admin_timecard_bulk.csv$', views.admin_bulk_timecard_list, name='AdminBulkTimecardList'),
```
batemapf: This is out in the latest commit.

tock/employees/models.py
```diff
@@ -38,6 +40,23 @@ class EmployeeGrade(models.Model):
...
+    def get_grade(end_date, user):
+        """Gets grade information based on a date and user. Original use of
+        function is to append correct grade information to each
+        TimecardObject."""
+        emp_grd_objects = EmployeeGrade.objects.filter(
+                Q(employee = user)
+                & Q(g_start_date__lte = end_date)
+                ).all().aggregate(Max('g_start_date'))['g_start_date__max']
+
+        if emp_grd_objects:
```
annalee: So this is kind of an oddball way to be going about this--we're checking if the user has grade objects, then, if they do, separately querying again to pull the one we want. It'd be more straightforward to just use get() to fetch the EmployeeGrade object we want, and set the grade to None if the object does not exist.

batemapf: Yep, good catch on the duplicate queries. Simplified it (and learned about all of the methods available for QuerySet objects!) via https://github.com/18F/tock/pull/461/commits/88892abbb8d89ffff56172691e75a31243abf187. So you don't have to go digging, here it is:

```
    def get_grade(end_date, user):
        """Gets grade information based on a date and user. Original use of
        function is to append correct grade information to each
        TimecardObject."""
        queryset = EmployeeGrade.objects.filter(
                Q(employee = user)
                & Q(g_start_date__lte = end_date)
                ).all()
        if queryset:
            return queryset.latest('g_start_date')
        else:
            return None
```

annalee:
Are we reaching out to infrastructure before merging this?

batemapf:
@annalee we got the 👍 from infra. good to merge?

PR #463, opened 2016-09-19
populates start_date attribute automatically

batemapf:
This change includes a today() value for the `start_date` attribute of the `UserData` class when a new object is instantiated. In practice, this means new users of Tock only see the reporting periods that occur _after_ they joined 18F instead of:


batemapf:
@annalee this one is ready for 👀 as well!

annalee:
We could throw a test on here to make sure that when we supply a nonexistent user, it creates the record as expected--it's not urgent but it's bot a bad idea.

annalee:
...or I could scroll down to the test which you already wrote. :shipit:

Issue #462, opened 2016-09-19
Allow aliases


batemapf:
Allow users to establish aliases that may be used in place of usernames. Use case: users who have a preferred or different name from the local part of their GSA.gov email address.

batemapf:
addressed in #464

PR #464, opened 2016-09-19
update views forms and templates to use human names instead of usernames

batemapf:
Uses a person's actual first name and last name in place of a system-generated username. Works in tandem with #463 which automatically generates first names and last names for new users. Also tidies up the UI a little bit by consolidating columns on list pages.

cc @emileighoutlaw from our chat this AM about aliases


batemapf:
see #462

Issue #458, opened 2016-09-16
Authentication setup for API


jessieay:
At the moment, there is no auth for the Tock API, eg: https://tock.18f.gov/api/projects.json

We should create and document an auth system for accessing the Tock API.

Pieces of this:
- Create a system of keys that can be created and deleted via the Django admin UI
- Keys are associated with a particular user or application (so we know who they are for)
- Keys can be used (sent in header?) to access any API endpoint
- Document how this auth works in the `README`

this will help us fix: https://github.com/18F/micropurchase/issues/951

toolness:
We should probably use `django_rest_framework`'s [`TokenAuthentication`](http://www.django-rest-framework.org/api-guide/authentication/#tokenauthentication) scheme for this.

For applications like Micropurchase, I'm wondering if it'll actually be most expedient to make a user for them that simply doesn't have an email address associated with it (so no one can "log in" as it), but that does have permissions and an API key.  Does that sound reasonable?

jessieay:
sure, that's fine. whatever works on your end works for us!

toolness:
Erm, I'm very confused now.  I thought that this app was migrating to a system where Django was handling all the auth, as per #446 (which was mysteriously closed), but it looks like the new auth system works much like the old one, where something in front of the Django app is setting an `X-Auth-User: firstname.lastname@gsa.gov` header.  As with the MyUSA-based auth, this makes it very difficult to use a completely different token-based scheme on the `/api/` endpoints unless whatever is sitting in front of the Django app is configured specifically to pass through all requests that start with `/api/`.

@batemapf any idea why this new auth scheme is in place, rather than #446?

batemapf:
ah yes, quite a mysterious close @toolness. meant to drop a comment in there referencing #456. we could certainly reopen and rework the authentication backend. however, setting up the UAA proxy to forward header info (X_AUTH_USER:...) to Tock is just simpler at this point...

...except for setting up an authenticated API :-)

Here is (roughly) what i am thinking:
1. set up a simple token-based API authentication system with a single token per user. could save this right in the employes.models.EmployeeData class and tack on a custom save() method to delete the token when the employee is `active=False`
2. require usage of that authentication system for all /api/ URLs.
3. create an alternate route to the app (say, tock-app.18f.gov/api/) that _is not_ behind the UAA proxy. the route would be limited to the /api/ path, so no other parts of the app could be accessed.
4. generate and supply tokens to users upon request.

jessieay:
[...]

PR #465, opened 2016-09-20
Add API Authentication

batemapf:
To address #458.

**TODO:**
- [x] update existing tests
- [x] add new tests for employees model method 
- [ ] remap route to tock.18f.gov/api w/o UAA proxy
- [ ] update README

### DIFF
--- a/tock/api/tests.py
+++ b/tock/api/tests.py
@@ -23,7 +23,14 @@ from rest_framework.authtoken.models import Token
 from django.test.client import Client
 from rest_framework.test import APIClient
 
+# common client for all API tests
 
+def client(self):
+    self.user = User.objects.get(username='aaron.snow')
+    self.token = Token.objects.get_or_create(user=self.user)[0].key
+    self.client = APIClient()
+    self.client.credentials(HTTP_AUTHORIZATION='Token ' + self.token)
+    return self.client
 
 # common fixtures for all API tests
 FIXTURES = [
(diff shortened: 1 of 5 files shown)


batemapf:
new tests not needed b/c only new code is calling a third party library (`rest_framework.authentication.TokenAuthentication`) which we don't need to test. cc @annalee 

existing tests have been updated to work with new auth system via https://github.com/18F/tock/pull/465/commits/6f085269a2d675940669a2039d6d11007114f39f

tock/employees/models.py
```diff
@@ -38,3 +39,13 @@ class Meta:
 
     def __str__(self):
         return '%s' % (self.user)
+
+    def save(self, *args, **kwargs):
+        if self.current_employee is False:
+            try:
+                token = Token.objects.get(user=self.user)
+                token.delete()
+            except:
```
annalee: In general, it's better to do specific exceptions like `except.DoesNotExist(): pass`, so that it's not silently failing on exceptions other than the ones we're expecting. When working on this, were you mostly seeing DoesNotExist, or was it throwing other exceptions as well?

tock/api/tests.py
```diff
@@ -39,11 +46,17 @@ def test_projects_csv(self):
 class ProjectInstanceAPITests(WebTest):
     fixtures = FIXTURES
 
+    def setUp(self):
+        self.user = User.objects.get(username='aaron.snow')
+        self.token = Token.objects.get_or_create(user=self.user)[0].key
+        self.client = APIClient()
+        self.client.credentials(HTTP_AUTHORIZATION='Token ' + self.token)
+
```
annalee: We're duplicating this setUp code here in several places. It probably makes sense to create a function that the various setUp functions all call, so that if we need to update one of these setup steps, we only have to update it in one place.

tock/employees/models.py
batemapf: Nope, just does not exist. Will update with the specific exception!

tock/api/tests.py
batemapf: On it!

batemapf: addressed via https://github.com/18F/tock/pull/465/commits/462a01c564e669a9375ed298f7279e3bbb441ac7 and https://github.com/18F/tock/pull/465/commits/c4b53fbba479ba3a3299cc008bec22eabaa68c0d

tock/employees/models.py
batemapf: addressed via https://github.com/18F/tock/pull/465/commits/27998c2c1341ce2229e0f6daf56af5f8b40b521f

PR #466, opened 2016-09-20
Align UserData and User models with save method

batemapf:
The UserData class extends the built-in User class. This change better aligns the two models when the `current_employee` attribute is updated in an instance of the UserData class.


annalee:
This is stale, but still useful, so we should clean up the conflicts on this and get it merged.

batemapf:
@annalee caught up, plus a little tidying. ready for review and merge!

PR #470, opened 2016-09-20
add active field to projects endpoint

batemapf:
@adelevie's delight. adding the `active` attribute to the project serializer.


adelevie:
looks friggin' sweet to me

annalee:
:shipit:

PR #475, opened 2016-09-21
make csv views available w/o token to auth'd user

batemapf:
This change is a complement to the new token-based authentication in #465. It provides logged in (via the UAA proxy) users with access to two common CSV reports w/o requiring a token. h/t to @annalee for working through it with me!

### DIFF
--- a/tock/api/views.py
+++ b/tock/api/views.py
@@ -1,6 +1,8 @@
 import collections
 import datetime
 
+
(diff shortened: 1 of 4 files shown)


batemapf:
doh! that's why you test locally before pushing to remote 😞

batemapf:
closing in favor of #476

PR #476, opened 2016-09-22
CSV download issue

batemapf:
This change is complementary to #465 and  moves CSV reports (views, serializers, tests, and URL confs) to the /reports/ route which is behind the cloud.gov UAA authentication system. Users who consume these reports can do so after authenticating into Tock as usual.

Users getting JSON data can hit the /api/ route with a token.

Tip of the hat to @annalee for encouraging me to do this the _right_ way :-).

### DIFF
--- a/tock/api/tests.py
+++ b/tock/api/tests.py
@@ -24,7 +24,6 @@ from django.test.client import Client
 from rest_framework.test import APIClient
 
 # common client for all API tests
-
 def client(self):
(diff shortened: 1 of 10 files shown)


annalee:
We'll want to email our ops users to let them know the links are changing, in case any of them have the csv links bookmarked directly.

batemapf:
👍

Issue #474, opened 2016-09-21
Document the API


adelevie:
cc @rememberlenny

jessieay:
see https://github.com/18F/tock/issues/458 as well :)

adelevie:
oh nice! lotta good info there!

batemapf:
updated README in latest [commit](https://github.com/18F/tock/commit/30dce2c000606d77d43fa888ce37e6f6811003a0#diff-04c6e90faac2675aa89e2176d2eec7d8)

adelevie:
Looking better, but the HTTP request methods and response schema/samples should also be here ... loving the progress!

PR #481, opened 2016-09-23
Api docs

batemapf:
@adelevie mind taking a look at these and letting me know if this is sufficient to address #474?


adelevie:
looks sweet! squash the commits!!!

PR #482, opened 2016-09-23
Re-do: replace ruby sass with node-sass

### LINKED ISSUE #436: Replace Ruby Sass with Node implementation

### LINKED ISSUE #439: Remove built assets (CSS) from git; build before deploying
This is a follow-up to #437. I confirmed with the Chrome dev tools that the CSS was being built and served fresh by the app whenever I changed the source files. But, as @jseppi points out, it's not ideal to have built CSS files (or their source maps) in git, because that can quickly lead to merge conflicts if more than one person modifies the source files in their respective branches. So, we need to do the following:

- [ ] Remove the built [CSS file(s)](https://github.com/18F/tock/tree/master/tock/tock/static/css) from git
- [ ] Ensure somehow that the CSS files are built before (or as part of) deployment
- [ ] Maybe move deployment over to Travis so that we can script the build and always deploy from a fresh environment with passing tests?

That second one and third are not currently in my wheelhouse because I haven't deployed the app in a very long time and don't know if it's possible to add build steps to either our [deploy script](https://github.com/18F/tock/blob/master/deploy.sh) (which is a year old and doesn't appear to be in use anymore) or the push to CF. A little help here, @annalee @batemapf?

batemapf:
This PR does a couple of things:

1. Fix #436 by removing the ruby sass stuff from the main docker build and replacing the `sass` container with a Node image that does the building
1. Fix #439 by removing the built CSS and source maps from git. They're now built by the docker container _and_ by the deploy scripts
1. Build CSS on Travis, because why not?


batemapf:
looks good on staging! merging :shipit:

PR #488, opened 2016-09-26
[WIP] Hours health status

batemapf:


batemapf:
for now, displays:

screen shot 2016-09-26 at 5 55 31 pm

annalee:
I think we want to track this information on a separate page, somewhere in the utilization space, rather than on time cards. On time cards, it's going to read as pressure to get people to tock a certain way, which isn't what we want.

PR #477, opened 2016-09-22
Upgrade CSV reporting and API tweaks

batemapf:
The commits on this PR make the following changes to the way data comes out of Tock:

1. Updates CSV reports;
2. Adds links to updated reports in templates;
3. Updates associated testing, including creation of fixture file for the UserData model; and 
4. Removes pagination per #484 for all API / report views.

Issue #487, opened 2016-09-26
My employee report graph is empty


hollyallen:
When I view my individual tock report, the graph is empty:

![screen shot 2016-09-26 at 2 38 52 pm](https://cloud.githubusercontent.com/assets/3476681/18852827/1cf5c8da-83f7-11e6-9939-c84734db7b22.png)

batemapf:
Please check it out now. Should be fixed!

hollyallen:
Confirmed!

PR #489, opened 2016-09-26
Fix missing chart data

batemapf:
Updating template to comply with changes to where CSV's live via #477.


batemapf:
to address #487

Issue #486, opened 2016-09-26
Still cannot see other employee reports


hollyallen:
Although #218 is marked as closed, I receive a 403 error when viewing another user's report.

batemapf:
Strange. Would you mind dropping in your target URL that is returning the 403?

hollyallen:
for example: https://tock.18f.gov/employees/patrick.bateman/

batemapf:
being addressed in #492 :-)

batemapf:
ok #492 is merged and live. enjoy!

hollyallen:
Yay, I can see other people's tock data now! Excited as only a manager can be.

batemapf:
If there are additional slices / views of employee data that you would be interested in seeing as a manager, don't hesitate to file an issue 😄

PR #492, opened 2016-09-27
Allow access to user detail

batemapf:
To address #218 and #486. 

Earlier resolution of #218 failed due to the retention of view-level permission (requiring either the requesting user be a superuser or requesting their own information). This PR addresses this issue as follows:

1. Creates a new employee detail view from the generic DetailView class that is accessible to all authenticated users;
2. Delivers this view via a new template, user_detail.html, which contains the same information from user_form.html (via UserFormView), but in static / non-editable form. [1]
3. The pre-existing UserFormView retains its permission requirement (superuser or self).
4. Testing added for new view.
5. Added new URL conf for user_detail.html and related view.
6. Updated logic on user_list.html to display one set of URLs (`/employees/f_name.last_name` for regular users; `/employees/e/f_name.last_name` for self or superusers).

_[1] The new template only displays the `end_date` attribute (employment end date) if one exists._

### DIFF
--- a/tock/employees/tests/test_models.py
+++ b/tock/employees/tests/test_models.py
@@ -20,6 +20,12 @@ class UserDataTests(TestCase):
         userdata.save()
         self.token = Token.objects.create(user=self.regular_user)
 
+    def test_string_method(self):
+        """Check that string method override works correctly."""
+        userdata = UserData.objects.get(user=self.regular_user)
+        expected_string = str(userdata.user.username)
+        self.assertEqual(expected_string, str(userdata))
+
     def test_user_data_is_stored(self):
         """ Check that user data was stored correctly """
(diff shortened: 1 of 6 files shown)


batemapf:
ok @annalee this one is ready for 👀 . functionality may be seen on staging, e.g.:

not self or superuser view: https://tock-staging.fr.cloud.gov/employees/3eae06.p4ytug/ 
self or superuser form view: https://tock-staging.fr.cloud.gov/employees/e/3eae06.p4ytug/

and updated user_list.html is: https://tock-staging.fr.cloud.gov/employees/. since you are a superuser in the stagingdb, all of your URLs will be of the `employees/e/...` variety.

annalee:
:shipit:

PR #495, opened 2016-09-28
[WIP] Add utilization calculations

batemapf:
Beginning of what could become robust, Tock-based billable utilization reporting for the org.


batemapf:
An early view...

screen shot 2016-09-28 at 8 04 38 am

batemapf:
I believe the fail on this is due lack of data in the testing DB. Runs and tests fine locally, but I did notice that the app failed when starting up using an empty DB...

tock/employees/models.py
```diff
@@ -1,3 +1,8 @@
+import datetime
+
+from django.db import models
+from django.db.models import Sum
+
 from django.db import models, IntegrityError
```
ccostino: It looks like the first `django.db import models` line here is a duplicate - I'd remove both of these and just add `Sum` to the import statement on line 9 along with `Q` and `Max`.

tock/employees/tests/test_views.py
```diff
@@ -146,7 +146,7 @@ def test_UserFormViewPostForAdmin(self):
         self.assertEqual(response.status_code, 200)
         # Check if data was changed
         # Check that data was changed
-        user_data = UserData.objects.first()
+        user_data = UserData.objects.get(user__first_name='Regular')
```
ccostino: This might cause an issue if more than 1 object exists with the same first name.  I know this is the test database, but we should still be careful with these assumptions and make sure we explicitly return or find a single record.

tock/employees/utils.py
```diff
@@ -0,0 +1,68 @@
...
+
+def get_timecardobjects(user):
+    queryset = TimecardObject.objects.filter(
+        user=user,
+        timecard__submitted=True
+    )
+    queryset = queryset.exclude(
+        project__name='Out of Office'
+    )
+    return queryset
```
ccostino: This method might be better served as a `TimecardObject` manager method instead; see how to [add new manager methods and bind custom managers here](https://docs.djangoproject.com/en/1.10/topics/db/managers/#adding-extra-manager-methods) in the Django docs.

tock/employees/utils.py
```diff
@@ -0,0 +1,68 @@
...
+
+def utilization_by_rp(queryset, date):
+    """Limits TimecardObject queryset to objects that fall within a single
+    ReportingPeriod."""
+    queryset = queryset.filter(
+        timecard__reporting_period__start_date__lte=date,
+        timecard__reporting_period__end_date__gte=date
+    )
+    return queryset
```
ccostino: Ditto for this method about the custom manager methods.

tock/employees/utils.py
```diff
@@ -0,0 +1,68 @@
...
+
+def utilization_by_range(queryset, start_date, end_date):
+    """Limits TimecardObject queryset to objectst that fall within a range of
+    ReportingPeriod objects."""
+    queryset = queryset.filter(
+        timecard__reporting_period__start_date__gte=start_date,
+        timecard__reporting_period__end_date__lte=end_date
+    )
+    return queryset
```
ccostino: Ditto for this method about the custom manager methods.

tock/employees/views.py
```diff
@@ -1,17 +1,22 @@
...
 from django.shortcuts import get_object_or_404
 
 
 from tock.utils import PermissionMixin, IsSuperUserOrSelf
+from .utils import get_last_n_rp, get_rps_in_fy, calculate_utilization, utilization_by_range, get_fy_first_day
 
 from .forms import UserForm
 from .models import UserData
 
+from hours.models import ReportingPeriod, TimecardObject
```
annalee: We're not actually using ReportingPeriod in this file, so we should be able to nix the import.

PR #504, opened 2016-10-04
[WIP] Float API integration

batemapf:
TBD...


batemapf:
Note to self: need to make sure blue/green binds the CUPS holding the Float API key to the app prior to staging.

batemapf:
...opening fresh PR.

PR #512, opened 2016-10-05
add billable param

batemapf:
all for limiting timecard reports to just those projects that are billable.


batemapf:
cc @MKathrynC

annalee:
:shipit:

Issue #508, opened 2016-10-05
Fix unique_with issue on EmployeeGrade


batemapf:
the EmployeeGrade model restricts to date+grade combos that are unique. this is incorrect. the limitation should be that date+grade+employee combos are unique.

annalee:
And fixed: https://github.com/18F/tock/pull/514

PR #514, opened 2016-10-06
address #508

batemapf:
@annalee this one is ready for review!

This patch:

1. Reworks the model validation via the `unique_for_date` field in EmployeeGrade, which now compares the employee field and the date field. Note, this only raise a validation error when working through the Django interface, not at the database-level. To address this...

2. Adds a custom save method that checks at the data for existing employee+g_start_date combinations and raises an IntegrityError if found.


batemapf:
Ah, should have mentioned, addresses #508!

annalee:
Nice work on the test! Also this isn't from this commit but using Start Date by itself instead of trying to track their start and end dates for each grade is quite elegant. :shipit:

batemapf:
😊😊😊 thank you!

PR #530, opened 2016-10-17
Utilization Reports

annalee:
New features to begin tracking employee utilization rates within Tock.

@batemapf started this work in https://github.com/18F/tock/pull/495. We're using the templates he created, and borrowing a lot of his logic. Big thanks to @ccostino for pairing with me on some of the under the hood code to chase down what was causing some serious problems with earlier versions.

TODO:
- [x]  Tests
- [x]  Add link to group utilization from the index page.


tock/tock/templates/utilization/group_utilization.html
```diff
@@ -0,0 +1,72 @@
...
+</p>
+<p>
+The contents of this page may only be viewed by Tock users who are marked as
+"staff" users. This attribute may be updated via the Tock
+<a href="/admin"> admin </a> interface, as well, via the <b>Users</b> page.
+</p>
+<p>
+Utilization is calculated by dividing the total number of hours submitted on
+projects that are marked "billable" in Tock, divided by the total number of
+hours submitted on all projects for the given period.
```
ccostino: This paragraph feels a bit clunky in its wording, would this work instead?

> Utilization is calculated by taking the total number of hours submitted on projects that are marked "billable" in Tock and dividing it by the total number of hours submitted on all projects for the given reporting period.

tock/utilization/tests/test_views.py
```diff
@@ -0,0 +1,127 @@
...
+
+class TestUtils(TestCase):
+    fixtures = FIXTURES
+
+    def test_fy_first_date(self):
+        date = datetime.date(2014, 7, 4)
+        self.assertEqual(get_fy_first_day(date), datetime.date(2013, 10, 1))
+
+    def test_get_dates(self):
+        periods = len(ReportingPeriod.objects.all())-1
```
ccostino: I caught this in a few spots in the code; if you find yourself looking to get the number of items in a QuerySet, use the QuerySet `count()` method instead, e.g. in this case:

`ReportingPeriod.objects.count() - 1`

When you use `len(ReportingPeriod.objects.all())`, you'll evaluate the whole QuerySet and since this is returning all models, that could be quite an expensive query and large result set!  Let the database do the heavy lifting here. :-)

For more info:  https://docs.djangoproject.com/en/1.10/ref/models/querysets/#count

tock/utilization/views.py
```diff
@@ -0,0 +1,160 @@
...
+
+    def get_queryset(self):
+        """Gets submitted timecards limited to the reporting periods in
+        question."""
+
+        """Although recent_rps is set to last four reporting periods, could
+        accept a form response that allows the user or app to dynamically
+        customize number of periods to include in the queryset."""
+
+        available_periods = len(ReportingPeriod.objects.all())
```
ccostino: Same note above:  `ReportingPeriod.objects.count()`

tock/utilization/views.py
```diff
@@ -0,0 +1,160 @@
...
+            related user and accounting code for later use."""
+            max_tos = TimecardObject.objects.filter(
+                timecard__user=staffer,
+                submitted=True,
+                timecard__reporting_period__start_date__gte=recent_rps[3]
+            ).prefetch_related(
+                'timecard__user'
+            ).prefetch_related(
+                'project__accounting_code'
+            )
```
ccostino: These two `prefetch_related` called could be combined into one:

`prefetch_related('timecard__user', 'project__accounting_code')`

The order does matter, but I think that's more in cases where you have one field referencing another.  More information can be found here:  https://docs.djangoproject.com/en/1.10/ref/models/querysets/#prefetch-related

tock/utilization/views.py
```diff
@@ -0,0 +1,160 @@
...
+                last_billable_hours['hours_spent__sum'],
+                last_all_hours['hours_spent__sum']
+            )
+
+        return billable_staff
+
+    def get_context_data(self, **kwargs):
+        context = super(GroupUtilizationView, self).get_context_data(**kwargs)
+
+        available_periods = len(ReportingPeriod.objects.all())
```
ccostino: Same as above:  `ReportingPeriod.objects.count()`

tock/tock/templates/utilization/group_utilization.html
ccostino: Disregard, this is likely going to get a writing lab treatment!

ccostino:
Final changes look good to me.  Thanks again, @annalee and @batemapf!

PR #546, opened 2016-10-31
Adds automatic saves to local storage

### LINKED ISSUE #545: Automatically save hours as they are entered
As a user, I would like my hours to be saved automatically as I enter them so that I don't accidentally lose them by forgetting to hit the enter button.

[PR incoming]

mgwalker:
## Description

Save hours to the browser's local storage, if available, as they are entered. When the page loads, check the local storage and populate the hours input boxes accordingly.  Clear the local storage when the form is saved to the server.  Resolves #545


maya:
Nice idea, @mgwalker! This would mean we could remove the "Save" button. 

Note: we'll need some comms to let people know about this feature.

mgwalker:
For reference, it turns out that `localStorage` is very well supported!  😃 

http://caniuse.com/#search=localstorage

mgwalker:
FWIW, I'd prefer keeping the server-side save.  It's nice to have the option of moving between computers.  Longer-term, it might be a good idea to change this from local storage to actually sending the hours to the server quietly behind the scenes.

maya:
Just a heads up that this will be on hold until the team complete's our research for the next steps for Tock.

tock/tock/templates/hours/timecard_form.html
```diff
@@ -131,6 +131,18 @@
   }
 
   function populateHourTotals(){
+    // Save hours to localStorage, if available
+    i(window.localStorage) {
```
ccostino: Looks like we're missing an "f" here, it should be `if`, right?

ccostino:
And if you're not interested, then we'll just close out this PR.  Just let me know, thanks! :-)

mgwalker:
@ccostino I fixed the typo, but I didn't see a problem with how it was saving data to local storage.  However, there is an issue where if you add a tock line and then refresh the page without clicking the save button, that tock line doesn't show up and the script doesn't know what to do with those hours from local storage.  So if I get some time before the end of the week, I'm going to try to get that sorted out.

ccostino:
@mgwalker that sounds good, thank you!

As for the issue you described, yeah, I can see how that'd be problematic.  The elements are added solely on the client side and aren't tracked until you actually click save or submit ([that code is here](https://github.com/18F/tock/blob/master/tock/tock/templates/hours/timecard_form.html#L212)).

Perhaps it would be a matter of saving not just the hours themselves but also the full line item, and restoring that upon a refresh/page load?  Or, keeping it simpler, trying to restore the hours and if the line item isn't found, ignoring it and clearing it from the local storage?  Whichever makes more sense to you, or if there's a different way entirely! :-)

mgwalker:
@ccostino Okay, now if you add a new line item and refresh the page, the script will add the line item if necessary.  😄

ccostino:
:clap: that's awesome!  I'll take a look again tomorrow, thank you @mgwalker! :-)

ccostino:
@mgwalker I'm sorry this has fallen off the radar a bit given end of year reviews, IRL, and a bunch of other things!

Seeing as how it should be good to go, do you mind taking another quick look at this PR to update it with the recent changes (which should be very small) and making sure it's ready to go?  I'll make sure it gets in. :-)  Thanks!

mgwalker:
@ccostino Looks like it still works to me.  👍

tock/tock/templates/hours/timecard_form.html
```diff
@@ -202,6 +253,12 @@ <h3 class="usa-alert-heading">{{ timecard_note.title }}</h3>
     };
 
     $("#save-timecard").on("click", function() {
+      // Clear anything saved locally.  The server is the
+      // source of truth.
+      if(window.localStorage) {
+        window.localStorage.removeItem('tock-entered-hours');
+      }
+
```
ccostino: One last piece change needed!  We probably want to clear the `localStorage` if a user clicks submit too, since that is also a save action. :-)

PR #549, opened 2016-11-04
right aligns date fields in employee table

OpenGlobe:
## Description

Right aligns date fields in employee table.

## Additional information

None.


maya:
Hi @OpenGlobe, what need is it solving to right align these dates?

OpenGlobe:
Hi @maya - I think it should reduce the strain on the reader - and be a bit more consistent with representing dates in tables.

It's a minor adjustment, so ignore if you like!

maya:
Hi @OpenGlobe, it seems like it would be more readable if all the text was left aligned and consistent, so we will ignore this. Thanks for the idea though!

PR #560, opened 2016-11-15
notes validation + meaningful error message

batemapf:
## Description

User could not save a timecard that contained a notes-required entry. This changeset addresses this by:

1. Only applying notes-required validation to cases where the user is attempting to submit the form, as opposed to all postings of the form;

2. Checking to see if there are any TimecardObject errors upon submission (non-save) and, if so, notifying the user that there is an error that needs to be corrected.

## Additional information

Successful save with no notes:

screen shot 2016-11-15 at 3 00 02 pm

Error message when a TimecardObject submission has an issue:

screen shot 2016-11-15 at 3 01 09 pm

### DIFF
--- a/tock/hours/forms.py
+++ b/tock/hours/forms.py
@@ -184,21 +184,25 @@ class TimecardObjectForm(forms.ModelForm):
 
     def clean(self):
         super(TimecardObjectForm, self).clean()
-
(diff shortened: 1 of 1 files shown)


batemapf:
@annalee this one is ready for your review!

annalee:
Nice work! Elegant solution.

Issue #552, opened 2016-11-07
Hide Project names with 0 hours for submitted timesheets


rdpiazza:
I review the timesheets each week for the members of my team using the report at https://tock.18f.gov/reports/[date]/[name]/.  Many people leave Tock lines on their timesheets that have 0 hours against them for the week, which makes them hard to review.  It would be very helpful if the reports at that location hid Project lines that had 0 hours charged against them.

batemapf:
closed via #564

PR #564, opened 2016-11-18
on submit, removes any TimecardObjects that have zero hours

batemapf:
## Description

Addresses #552 by marking TimecardObject's to be DELETE'ed if they have a zero value for hours_spent. Upon merge and deployment, this will eliminate the existence of all future timecards having entries with 0 hours spent.  cc @HuixianXu and @MKathrynC who will also probably be happy to eliminate this issue.

## Additional information

Here is a screenshot of the problem:

screen shot 2016-11-18 at 11 19 29 am

After merge and deploy of this change, I also recommend we clear out all existing TimecardObjects with zero hours spent from the production database.

### DIFF
--- a/tock/hours/forms.py
+++ b/tock/hours/forms.py
@@ -256,6 +256,9 @@ class TimecardInlineFormSet(BaseInlineFormSet):
                     )
                 total_hrs += form.cleaned_data.get('hours_spent')
         if not self.save_only:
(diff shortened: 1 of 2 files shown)


batemapf:
another one for your 👀 @annalee!

annalee:
Nice work! Clean, elegant test.

PR #568, opened 2016-11-21
Patch #564 to handle errors correctly

batemapf:
## Description

This PR patches #564 to handle situations where a submitted form within the TimecardInlineFormset is completely empty (i.e. no project, no hours_spent) and a KeyError is thrown.


batemapf:
@annalee please bump this to the top of your review list. Need to merge in this patch before any further deployments!

Issue #565, opened 2016-11-18
Strategists Missing from CSV


abrouilette:
Jeremy Zilar, Nikki Lee, Nikki Zeichner, and Cyd Harrell are sorta missing from the downloaded csv for the week of November 6, 2016 to November 12, 2016, per @acolter. (Well, their emails are.  See below.) @sfvictor reviewed their profiles and can find no reason they are not showing up, so the question is now posed to you, @annalee & @batemapf.

abrouilette:
Updating: Same people missing from the prior week's report.

abrouilette:
Further, they do not appear in the CSV at all, so this is not an error in Strategy's function.

acolter:
They don't appear in the CSV, but their time entries do. There's just no email address listed in the User column. (Try this with Cyd Harrell's timesheet, for example. If you go to the Reports by weekly reporting period link for November 6, 2016 to November 12, 2016 and navigate to her time card, you'll see she worked 14.0 hours on FFD. 

But if you search for "Harrell" in the CSV file, you don't see her. You *do* see an entry for 14.0 hours for FFD with no email in the User column. 

Let me know if I can answer any other questions.

batemapf:
closed via #571

PR #571, opened 2016-11-21
Fix Reporting Period CSV Report

batemapf:
## Description

Addresses issues raised in Slack on 11/18 re: missing names in the reports/[YYYY-MM-DD].csv file. The CSV writer relied on email address data for the "User" column of the CSV. If the user did not have an email address in the app, a blank was written in its place. This created the appearance of missing records in the output CSV.

This changes switches out email address for username, which is more reliable because all users have a username.

cc @GailSwanson @acolter


batemapf:
addresses https://github.com/18F/tock/issues/565

Issue #471, opened 2016-09-20
General Snippets Report


joshuabailes:
I'd like to create a report to show the 
- employees who tocked time to general 
- their hours for that week to general
- and their snippets for that activity

joshuabailes:
@batemapf

joshuabailes:
just checking in on this. timeline? @abrouilette

batemapf:
@abrouilette where does this fall in the @joshuabailes priorities?

abrouilette:
First, as of 11/22.

PR #575, opened 2016-11-23
Snippets report

batemapf:
## Description

Addresses #471, which was listed as @joshuabailes top priority. This change creates a new CSV report that is limited to all General entries that have snippets (notes is the Tock terminology). The report is accessible in the Reports section under "Regular Reports".

## Additional information

Example of CSV output:

screen shot 2016-11-23 at 11 17 17 am


annalee:
👏  .

Can we add tests here before merge?

batemapf:
tests added!

PR #577, opened 2016-11-24
Util summary row

batemapf:
## Description

Adds a summary row to the utilization report page for each unit.

## Additional information

Screen shot of changes:

---
screen shot 2016-11-24 at 9 32 01 am


maya:
A (late) LGTM for design. Totals at the bottom is what I've most other tables do.

maya:
@batemapf one quick thing I just noticed is the totals at the bottom don't have a space on each side of the `/` between the total amounts and they have a spaces on the individual amounts above. It would be good to keep it consistent.

Issue #551, opened 2016-11-07
Add sum of billable/nonbillable hours for submitted timesheets


rdpiazza:
As a Chapter director, I go in and review timesheets each week using the reports at https://tock.18f.gov/reports/[date]/[name]/.  One thing I look for is the sum of billable vs. nonbillable hours for that week for the employee.  It would be extremely helpful if the submitted timesheets included a summary of the number of billable vs. non-billable hours for that employee.

PR #578, opened 2016-11-25
User summary row

batemapf:
## Description

Adds a new table to the user timecard detail page that summarizes billable hours, all hours, and billable percentage for a user in a specific reporting period. Feature requested in #551.

This change also addresses an existing n+1 query issue by prefetching project and accounting code information along with timecard info for eventual use in the template.

## Additional information

Screenshot of changes:

----

screen shot 2016-11-25 at 11 29 18 am


maya:
@batemapf This looks good, one thing I noticed was since there's some new microcopy with the `Reporting Period Summary`, I wonder if we wanted any content 👀 to check if it should it be more specific referring to its billable utilization and should that match the utilization report page title or not?

PR #583, opened 2016-11-29
[WIP] 511 addhours endpoint

elg0nz:
## Description
This PR implements #511: an AddHours endpoint that allows a user to add hours for a specific line item to the current time card so that it's easier to contemporaneously track time in Tock.

PR #584, opened 2016-11-29
Add a default timezone to .travis.yml

elg0nz:
## Description

This PR adds a default timezone to .travis.yml
This prevents datetime drift related errors when running CI.
See https://github.com/18F/tock/pull/583#issuecomment-263477596
for an example of tests that can fail due to that drift.


batemapf:
👍 thanks @elg0nz!

PR #585, opened 2016-11-29
Content Tweaks

batemapf:
## Description

Minor changes for usability.


batemapf:
@maya re: your comments on #577 and #578, how does  this look?

maya:
@batemapf #577 looks good and #578 is more generally about our microcopy - like headings, etc... I've opened an issue here to address it: https://github.com/18F/tock/issues/586

PR #593, opened 2016-12-01
Fix FY2017 report

batemapf:
## Description

Aligns the "FY2017" link on the reports list page to actually align with FY2017.

cc @MKathrynC


batemapf:
merging after getting the 👍 from @annalee!

Issue #511, opened 2016-10-05
Feature: Timecard "intents"


adelevie:
**As a user, I want a clickable link which will add hours for a specific line item to my current time card so that it's easier to contemporaneously track my time in Tock.**

This project is a [sealed bid auction](https://micropurchase.18f.gov/auctions/rules/sealed-bid).

---

### Acceptance criteria 

**Scenario: Clicking a link with valid structure**
- Given I am an authenticated Tock user
- And I have entered a Tock URL* into the browser address bar
- And it contains an [add hours command]
- And it defines a valid ‘project number’ value
- And it defines a numeric ‘hours’ value
- When I hit submit and load the page
- Then the defined time will be added to the user’s unsubmitted timecard during the present reporting period for the defined project
- And my current timecard page will load
- And that page will display a confirmation message stating, “[Time] hours have been added to [Project name].”
- And the confirmation message will have an “Undo” link

*Note: URL structure should be `/addHours?project=231&hours=1` 

**Scenario: Undoing an ‘Add Hours’ addition**
- Given I am an authenticated Tock user
- And I am on the timecard page
- And I just added time to a project via a Tock URL with valid ‘Add Hours’ command and values
- When I click on the ‘Undo’ link in the confirmation message
- Then the hours just added will be removed from the defined project
- And I will see a confirmation message stating, “[Hours value] has been removed from [project name].”

**Scenario: Loading a Tock ‘Add Hours’ URL with invalid structure**
- Given I am an authenticated Tock user
- And I have entered a Tock URL into the browser address bar
- And it contains the [add hours command]
- And it does NOT include valid project and hour values
- When I hit submit and load the page
- Then my current timecard page will load
- And it will display an error stating, “Oops. That command was not correct and no time was added to your timecard. Try again by entering a URL with this format: `tock.gov/addHours?project=231&hours=1`

### General acceptance criteria 

- B or better Code Climate grade and 90% or higher test coverage for all modifications.
- Tests adequately cover any new functionality.
- Tests run on Travis-CI and do not break the build.
- A payment URL is [specified in your micro-purchase profile](https://micropurchase.18f.gov/docs/getting_paid) so we can pay the award

edmullen:
Commenting to be notified.

edmullen:
Congratulations @mikejihbe on winning this micro-purchase auction!

[...]

PR #594, opened 2016-12-01
511 Add hours endpoint [updated]

elg0nz:
## Description
This PR implements #511: an AddHours endpoint that allows a user to add hours for a specific line item to the current time card so that it's easier to contemporaneously track time in Tock.

### DIFF
--- a/tock/hours/tests/test_views.py
+++ b/tock/hours/tests/test_views.py
@@ -665,6 +665,16 @@ class ReportTests(WebTest):
         new_tco.refresh_from_db()
         self.assertEqual(new_tco.hours_spent, Decimal('12.12'))
 
+        undo_url = "%s?project=%d&hours=-2" % (reverse('AddHours'), new_project.id)
+
+        response = self.app.get(
+            undo_url, None,
+            headers={'X_AUTH_USER': self.user.email}
+        )
+        self.assertEqual(response.status_code, 302)
+        new_tco.refresh_from_db()
+        self.assertEqual(new_tco.hours_spent, Decimal('10.12'))
+
     def test_do_not_prefill_timecard(self):
         """
         Test that a timecard doesn't get prefilled
--- a/tock/hours/tests/test_views.py
+++ b/tock/hours/tests/test_views.py
@@ -626,10 +626,15 @@ class ReportTests(WebTest):
         )
         self.assertEqual(prefilled_hours, {None})
 
-    def test_add_hours_with_existing_tco(self):
+    def test_add_hours(self):
         """
         Test that addHours increases the current timecard by the
-        specified amount
+        specified amount.
+
+        This test exists both as a sanity check, and for checking HoursAdder is being used
+        internally. Normally we would assert HoursAdded is called with the Mock library,
+        but since that dependency isn't in that project yet, we'll have to settle
+        and test side effects instead.
         """
         self.user = self.regular_user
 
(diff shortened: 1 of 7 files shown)


tock/hours/views.py
```diff
@@ -313,6 +315,64 @@ def form_valid(self, form):
...
+        error_msg
+    )
+
+
+def add_hours_view(request):
+    project_id = request.GET['project']
+    hours = Decimal(request.GET['hours'])
+    r = ReportingPeriod.objects.latest()
+
+    try:
```
batemapf: @elg0nz Please add test coverage for this error handler.

tock/hours/views.py
```diff
@@ -313,6 +315,64 @@ def form_valid(self, form):
...
+    updated_hours = tco.hours_spent + hours
+    updated_hours = max(0, updated_hours)
+
+    tco.hours_spent = updated_hours
+    tco.save()
+
+    if hours > 0:
+        plural_hours = pluralize(hours, 'hour,hours')
+        has = pluralize(hours, 'has,have')
+        undo_query = "?hours=-%s&project=%s" % (hours, proj.id)
```
batemapf: @elg0nz Please add test coverage for the undo functionality.

tock/hours/views.py
```diff
@@ -313,6 +315,64 @@ def form_valid(self, form):
...
+        user_id=request.user.id)
+
+    if created:
+        tc.save()
+
+    tco, created = TimecardObject.objects.get_or_create(
+        timecard_id=tc.id,
+        project_id=proj.id)
+
+    if tco.hours_spent is None:
```
batemapf: @elg0nz Please add test coverage for this error handler.

tock/hours/views.py
batemapf: @elg0nz Also please include a test case where the user submits an invalid value for the `hours=` param.

tock/hours/views.py
```diff
@@ -313,6 +315,64 @@ def form_valid(self, form):
...
+        tco.hours_spent = 0
+        display_add_hours_error_message(request)
```
batemapf: @elg0nz When new TimecardObject's are initialized, they have a hours_spent value of None. Redirecting a user with an error message if a TimecardObject's hours_spent attribute is None means that the success of a URL-based addHours request is dependent on whether the user has: (1) an existing timecard for the period; and (2) has a TimecardObject with a project attribute that matches the project param of the request. Otherwise, the code will (correctly) initialize a new TimecardObject with hours_spent as None and promptly redirect with an error message.

tock/hours/tests/test_views.py
```diff
@@ -367,6 +368,69 @@ def test_prefilled_timecard(self):
...
+
+        url = "%s?project=%d&hours=2" % (reverse('AddHours'), new_project.id)
+
+        old_count = new_timecard.timecardobjects.count()
+        response = self.app.get(
+            url, None,
+            headers={'X_AUTH_USER': self.user.email}
+        )
+        new_count = new_timecard.timecardobjects.count()
+        self.assertEqual(old_count + 1, new_count)
```
batemapf: In addition, testing the redirect response info and to see if the correct message is included.

tock/hours/views.py
elg0nz: Yes, that was a typo on our hand. We should display that error when the 'hours' parameter is missing and not when the tco doesn't have any hours spent. Fixing right now.

batemapf:
@elg0nz we are almost there. The last sticking point is this requirement:

> Then the defined time will be added to the user’s unsubmitted timecard during the present reporting period for the defined project

It doesn't look like this change set controls for `unsubmitted` status (looked [here](https://github.com/18F/tock/pull/594/files#diff-2b1a8c7bcdad4553bc847dd86df6aa43R77)) or `present` (looked [here](https://github.com/18F/tock/pull/594/files#diff-bbf773df48e042c9e0b2f421ddc3ee38R319)). These both need to be addressed, with test coverage, in order for us to accept.


cc @mtorres253

elg0nz:
Hi @batemapf could you provide a couple clarifications?

1. For the timecards: if an unsubmitted timecard is found, 
what should happen? 

   a. Create a new unsubmitted Timecard. 
**or**  
   b. Fail and display an error?

2. For the "Present Reporting Period": 
The ReportingPeriod model doesn't have a present attribute, so we're thinking about just adding a unit test to the add_hours views to check that we're fetching the present (latest) reporting period. 

Is this the right approach? Are there any other conditions we should assert on our unit tests?

Thanks!

batemapf:
> For the timecards: if an unsubmitted timecard is found,
what should happen?

If an unsubmitted timecard is found, nothing should change. If a submitted timecard is found, an error message should be shown.

> For the "Present Reporting Period":
The ReportingPeriod model doesn't have a present attribute, so we're thinking about just adding a unit test to the add_hours views to check that we're fetching the present (latest) reporting period.

This is a vocabulary issue. `Present` was intend to mean having a `start_date` that precedes the current date and an `end_date` that occurs after the current date. If no reporting exists, an error message should be shown.

elg0nz:
@batemapf updated.

mtorres253:
Thanks @elg0nz!  @batemapf let us know if everything is good.

PR #588, opened 2016-11-29
Add appropriate elements to table markup

maya:
## Description

- Adds `<thead>`, `<tfoot>`, and a `<tbody>` elements to properly mark up the table. 
- Fixes markup indentation.

TODO:

- [x] Add `scope="col"` or `scope="row"` to `th`s. See: https://standards.usa.gov/tables/
- [x] Add `<caption>` to table for title
- [x] Remove unused table classes

## This is what it looks like:
screen shot 2016-11-29 at 2 59 52 pm

PR #597, opened 2016-12-02
Fix bug while comparing dates on TimeCardTests

elg0nz:
This PR Fixes the time drift bug reported on 
https://github.com/18F/tock/pull/588#issuecomment-264018029

While #584 did fix the issue where Travis didn't had a timezone setup, the root cause of the bug was that the datetime.datetime.today method ignores the current timezone, so we need to use a different one. See the commit below for more details.


elg0nz:
@batemapf here's another fix for the Time drift issue. Hopefully, this one will fix it for good 😎

batemapf:
Closing and replacing with #597, which extends to all test cases. Great contribution @elg0nz!

PR #605, opened 2016-12-06
Patch Ops Report

batemapf:
## Description

Rolls back the ability to "zoom in" on specific units in the Operations Dashboard, which will require a better understanding of how @joshuabailes and the business calculates such things. Also updates report so that only 18F "units" are included (excludes PIF and AcqStack) in the org-level report.

## Additional information

Screenshot:
screen shot 2016-12-06 at 2 21 52 pm


batemapf:
@annalee this is what i mentioned during our stand up today. for your 👀!

Issue #542, opened 2016-10-25
Utilization Report: Please show non-billable also


hollyallen:
It would be helpful to me to see the Engineering Chapter "exempt" staff in the utilization report. Some of us do bill hours occasionally, and this is the only view in Tock where I can see folks filtered by chapter. Thanks!

batemapf:
@hollyallen this is an easy tweak. however, before implementing, would you mind checking with your supervisor colleagues to confirm that they will not be bothered by all of the additional names of "exempt" employees? immediate impact will be more stuff to scroll through, and possibly a tiny (probably not noticable) decrease in performance.

rdpiazza:
I don't have an issue with this since Product's utilization report already shows a few people with 0 hours, like Claire (an ELP) and Clara (on detail from USDS).  The point raised about performance touches on a related feature I was thinking of that would probably be helpful for this:  showing chapter performance totals for both staff who are primarily billable and the team at large.  Would be helpful to see that primarily billable people average, say, 81% billable even if the team at large (with exempt staff and details) average, say, 70%.  Would also give a more accurate number of total hours for the team per week if some of the exempt staff are able to bill.

batemapf:
@rdpiazza could definitely do total rows for team at large plus components of the team. any idea where i get the staff "types", or is it just "exempt" and "billable"?

rdpiazza:
Hmmm...I'm asked to fill in spreadsheets frequently to flag people as Primarily Billable, Partially Billable, or Exempt but I'm not sure how it's stored in Tock or other systems you could access via an API.  Maybe Bailes or Talent would know?

PR #606, opened 2016-12-06
Billable vs. All Toggle for Util Report

batemapf:
## Description

Addresses need in #542 by allowing the user to toggle the utilization report between "billable" and "all" employees, with default to "billable" only. Accomplishes this functionality through including the parameter "all" in the request to view all employees.

Also makes slight tweak to display of units for a more consolidated design.

## Additional information

Screen shot:

screen shot 2016-12-06 at 6 04 13 pm

### DIFF
--- a/tock/tock/templates/utilization/group_utilization.html
+++ b/tock/tock/templates/utilization/group_utilization.html
@@ -21,12 +21,21 @@
   projects that are marked "billable" in Tock, divided by the total number of
   hours submitted on all projects for the given period.
 </p>
+<h3>Select employee type: </h4>
+  {% if request.GET.all == '' %}
+  All | <a href="{% url 'utilization:GroupUtilizationView' %}">Billable Only</a>
+  {% else %}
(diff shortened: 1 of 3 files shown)


batemapf:
@maya and @annalee ready for your 👀

maya:
@batemapf what do you think of designing the "Jump to" filter like this:

ut-filter

batemapf:
Um that is awesome!

maya:
In the new filter here there's All and Billable. What kind of billable is that: primary or partially? 

A complete filter would include parameters for: all, primary billable, partially billable, and exempt.

batemapf:
right now we only have `is_billable` as a T/F boolean field. would be happy to update it a non-binary field (maybe a choice field?), but still not really clear on where this data would come from / who owns...

maya:
Who'd be able to tell us what the current `is_billable` field means (primary or primary + partially, etc...)? Then we can find out what the use case of it is. Since formerly exempt staff are now billing, it would make sense to include them in the report to get the most accurate data.

PR #608, opened 2016-12-08
Add acquisitions project filter

batemapf:
## Description

Filters out projects that have Acquisition Services profit / loss information from the operations report, per @joshuabailes.


annalee:
We shouldn't make a habit of hard-coding this kind of thing in. If filtering is a regular issue, we'll want to create a 'include in reports' field on the pla model and filter based on that.

PR #611, opened 2016-12-15
Turn Back the Tock

batemapf:
## Description

This changeset rolls Tock back to an earlier time in order to refocus on the its core responsibility: tracking time.

Here is my thinking, all premised on the idea that Tock is going away in the next 90 - 120 days.

(1) Regardless of what tool we end up with, I hypothesize that some amount of bespoke reporting / data munging will have to be done. This means we should get used to a more modular world where reporting (micro)app(s) consume data from _n_ +1 sources, combined, and produce reports (whether for business analysis by management or to support operational execution, like billing). At most, we should look at our timekeeping solution as simply one of many sources of data, instead of containing multiple datasets (project, personnel, accounting info). The sooner we can yank this stuff out of Tock (w/o effecting the things that are presently relied upon) and start thinking about more appropriate architecture, the better.

(2) Since we will keep moving forward with the Tock ATO in the meantime, I suspect reducing the dataset, and access to that dataset, that are within the app’s scope will be helpful in smoothing that path. Ideally, the only data that Tock would hold would be Timecard and TimecardObjects with the most basic person/period/hour/project info. In essence, returning Tock to its roots 😄 .

(3) Given the diversity of persons responsible for Tock’s current datasets (e.g. TTS Talent for personnel; TTS Ops for finance and agreements; BUs for project info), I suspect a single app solution for all of these things is likely not possible. I believe a more workable solution is to develop lightweight processes (technical and/or non-technical) that allows these various points of responsibility to easily expose their data to those who need it. For you long-term 18F’ers out there, yes, this sounds a lot like the Team API 🎉 .


annalee:
Code looks good, but we're on hold pending confirmation we're not going to mess anyone up by rolling back these features.

abrouilette:
What do we need from who to confirm that?

ccostino:
Pouring one out for @batemapf's effort here - but the code has changed enough that this would require a rewrite of sorts.

We may revisit in the future, thoug!

PR #622, opened 2017-02-03
Update slim report

batemapf:
## Description

Adds the `project_id` attribute from the `Project` class to the `SlimBulkTimecardSerializer` per @HuixianXu's request. Tests updated to reflect addition.

While making change, updated two failing tests to rely on UTC time, instead of local.


batemapf:
@annalee 🏀 is in your ⚖️ !

PR #624, opened 2017-03-14
Improve grade saves

batemapf:
## Description

Builds on #461 by extending the application of `EmployeeGrade` objects to `TimecardObjects` that existed previously.

## Additional information
The second commit on this PR address an issue raised by a Tock admin earlier today. If an admin needs to add a new `EmployeeGrade` object that represents a change in the subject's grade prior to the current `ReportingPeriod`,  the only way to associate this new grade with old timecard info is to access the production app via SSH and call the `save()` function on each `TimecardObject` in questions. _Very scary!_ This set of commits remedies this situation by calling those `save()` functions automatically when a new `EmployeeGrade` is created.

The first commit addresses the unintended `CASCADE` setting for the foreign key relationship from `TimecardObject` >>> `EmployeeGrade`. Under the current setup, an admin user could unintentionally delete all of a user's `TimecardObject`s associated with a grade that is purposefully being deleted. _Not fun!_


tbaxter-18f:
This is really old...  is it still a desired improvement? What's the hold-up.

@ccostino ?
@rogeruiz ?

ccostino:
I think we can close this work, right @abrouilette?  We do not need this?

tbaxter-18f:
Given it's 18 months old, I'm going to assume it's either unimportant or was handled elsewhere. At the very least it's outdated made more interesting by changes in Django since then around the use of CASCADE. So, I'm gonna make the call to close it and will reopen if needed.

ccostino:
That sounds totally reasonable to me!

PR #630, opened 2017-06-27
Show Float hours in timecard form

batemapf:
## Description

Connects with the [Float](https://float.com) API to display a user's scheduled hours for the week in which they are reporting.

Sort of like:

screen shot 2017-06-27 at 5 53 50 pm

This is a refactor of #504.


tock/employees/models.py
```diff
@@ -96,34 +96,39 @@ class UserData(models.Model):
     is_18f_employee = models.BooleanField(default=True, verbose_name='Is 18F Employee')
     is_billable = models.BooleanField(default=True, verbose_name="Is 18F Billable Employee")
     unit = models.IntegerField(null=True, choices=UNIT_CHOICES, verbose_name="Select 18F unit", blank=True)
+    float_people_id = models.CharField(max_length=50, null=True, blank=True, verbose_name='Float "people_id" attribute')
```
batemapf: I _believe_ the decision was based on what was coming across from Float as the `people_id` via the [/people](https://github.com/floatschedule/api/blob/master/Sections/people.md) end point. Maybe I was future proofing in case there was a change in the content of the people_id to include non-integer characters? As you noted below, I tried to control for a blank people_id (that is, Tock pinged Float to find info a user but returned `''` when it couldn't find a match) in the `save()` function.

Of course happy to flip it to an `IntegerField` if you think it is better practice!

tock/employees/models.py
```diff
@@ -96,34 +96,39 @@ class UserData(models.Model):
...
         verbose_name='Profit/loss Accounting String'
     )
+
     class Meta:
         verbose_name='Employee'
         verbose_name_plural='Employees'
 
     def __str__(self):
-        return '%s' % (self.user)
+        return '{0}'.format(self.user)
```
batemapf: Ah yes!  Something like `return f'{self.user}'`?

tock/employees/models.py
```diff
@@ -96,34 +96,39 @@ class UserData(models.Model):
...
             try:
                 token = Token.objects.get(user=self.user)
                 token.delete()
             except Token.DoesNotExist:
                 pass
-        user.save()
+            user.save()
+
+        if self.float_people_id is '':
+            self.float_people_id = None
```
batemapf: See above note.

tock/tock/settings/test.py
```diff
@@ -17,3 +17,54 @@
...
+class TestMockServer(object):
+    def run_server(port):
+        # Configure mock server.
+        mock_server_port = port
+        mock_server = HTTPServer(('localhost', mock_server_port), MockServerRequestHandler)
+        # Start running mock server in a separate thread.
+        # Daemon threads automatically shut down when the main process exits.
+        mock_server_thread = Thread(target=mock_server.serve_forever)
+        mock_server_thread.setDaemon(True)
+        mock_server_thread.start()
```
batemapf: Ooooo. That is much cleaner and definitely agree on the increasing complexity / length required for adding additional external endpoints. Will give it a shot!

tock/hours/views.py
```diff
@@ -613,6 +616,71 @@ def get_object(self, queryset=None):
...
+            endpoint='tasks',
+            params={
+                'weeks': base.FLOAT_API_TASK_WEEKS,
+                'start_day': datetime.datetime.strptime(
+                    self.kwargs['reporting_period'], "%Y-%m-%d"
+                ).date()
+            }
+        )
+        if r.status_code != 200:
+            return None
```
batemapf: Yeah, I hear ya. I had a whole elaborate (read: overbuilt) error reporting system in an [earlier commit](https://github.com/18F/tock/blob/1715b52925227b863d60637d28cf1059639ec62e/tock/hours/views.py#L345). A little logging might be good though, will add in a `print()` statement if response is not 200.

toolness: Cool--the "django-y" way to do it is via [the Django/Python logging system](https://docs.djangoproject.com/en/1.11/topics/logging/) but honestly I find that to be one of the most confusing things I've ever encountered in programming, so it's up to you.  `print` is fine with me as long as it's okay with whoever is doing devops for Tock!

tock/hours/views.py
```diff
@@ -613,6 +616,71 @@ def get_object(self, queryset=None):
...
+        else:
+            task_data = r.json()
+        #From response, derive meta data and tasks for the period.
+        f_start_date = datetime.date(
+            int(task_data['start_yr']), 1, 1) + datetime.timedelta(
+            days=int(task_data['start_doy']
+            )-1)
+        f_end_date = f_start_date + datetime.timedelta(
+            days=(base.FLOAT_API_TASK_WEEKS * 6))
```
batemapf: I've [refactored](https://github.com/18F/tock/pull/635/files#diff-bbf773df48e042c9e0b2f421ddc3ee38R636) this one a bit in #635. Originally, I was getting the start and end date of the period in question from the Float response (hence the `f_...`). Float, of course, was getting it from the Tock reporting period start and end dates supplied in the request. I believe my thinking was to ensure what was ultimately being passed to the template in terms of start and end dates came directly from Float (to avoid any confusion over what period the Float data was regarding). However, I later decided (in #635) that this was redundant and nixed it.

The `6` represents weekdays, as in:
```python
>>> import datetime as dt
>>> a = dt.date(2017, 7, 9) # Sunday, or the standard start day of a reporting period
>>> a + dt.timedelta(days=6)
datetime.date(2017, 7, 15) # Saturday, or standard the last day of a reporting period.
```

The `FLOAT_API_TASK_WEEKS` is a overly verbose variable representing the number of weeks of Float data that should be requested (from 1 - 52). I don't really know why we would want to get more than one week of data (since Tock reporting periods are one week long), but I think this may have been some future proofing in case the length of reporting periods changed? Of course, multiplying 6 by 2 (weeks) doesn't get you to a Saturday if the start date is Sunday. Will give this one a rethink...

tock/hours/views.py
```diff
@@ -613,6 +616,71 @@ def get_object(self, queryset=None):
...
+        clean_data = {
+            'metadata': {
+                'float_period_start': f_start_date,
+                'float_period_end': f_end_date
+            },
+            'tasks': [ item for sublist in \
+                [ ii for ii in [i['tasks'] for i in task_data['people'] \
+                if i['people_id'] == float_people_id] ] \
```
batemapf: Yup, can definitely add in a helper :-)

tock/employees/models.py
toolness: Ah interesting--I agree that it makes the most sense to use whatever data type the float API is using, so `CharField` sounds good to me.  I think what the Django best practices docs are getting at, though, is that we should try normalizing to only one type of "there is no value for this field" if possible. Do you think it's OK to only use `''` as that blank value, i.e. make `None` an illegal value rather than allowing both but manually coercing all `''`'s to `None` in `save()`?  (One potential danger of relying on the latter is that in certain edge cases `save()` doesn't actually get called, which is super annoying.)

Anyways, either way this probably isn't that big a deal.  In the worst case we can do a migration at a later date to make the schema more consistent.

tock/hours/views.py
```diff
@@ -613,6 +616,71 @@ def get_object(self, queryset=None):
...
+            for item in sublist ]
+        }
+        # Calculate estimated hours per week. Need to take into account
+        # holidays in future build.
```
batemapf: Would definitely work for holidays (assuming there isn't any Federal employee weirdness around administrative closings that might not be captured by `businesstime`, but also need to figure out individual time off (see the [/timeoffs](https://github.com/floatschedule/api/blob/master/Sections/timeoff.md) endpoint). Dove into this over on #635.

tock/hours/views.py
```diff
@@ -613,6 +616,71 @@ def get_object(self, queryset=None):
...
+
+    def get_float_data_for_context(self):
+        userdata = UserData.objects.get(user=self.request.user)
+        float_people_id = userdata.float_people_id
+        if float_people_id is None:
+            # Attempt to get the user's Float people_id.
+            r = get_float_data(
+                endpoint='people')
+            if r.status_code != 200:
+                return None
```
batemapf: Yup, agreed.

tock/tock/templates/hours/timecard_form.html
```diff
@@ -10,6 +10,26 @@
...
 </div>
+{% if float_data.tasks %}
+  <div class="usa-alert usa-alert-error">
+    <div class="usa-alert-body">
+      <p class="usa-alert-text">
+        <b>
+          For {{ float_data.metadata.float_period_start }} to
+          {{ float_data.metadata.float_period_end }} you were scheduled in Float
+          for:
+        </b>
```
batemapf: Got it, nothing wrong with doing things the right way!

tock/tock/templates/hours/timecard_form.html
```diff
@@ -10,6 +10,26 @@
...
+      </p>
+    <ul>
+      {% for i in float_data.tasks %}
+       <li>-  {{ i.hours_wk }} hours on {{ i.project_name }}
```
batemapf: Will have a look. I agree on the `<li>`, but maybe there was some USWDS weirdness?

tock/tock/templates/hours/timecard_form.html
```diff
@@ -56,14 +76,20 @@
...
+      <div class="usa-alert-body">
+        <p class="usa-alert-heading"><strong>{{ formset.non_form_errors }}</strong></p>
+      </div>
+    </div>
   {% endif %}
   {% if messages %}
-  <ul class="messages">
     {% for message in messages %}
-      <b><li{% if message.tags %} class="alert-{{ message.tags }}"{% endif %}>{{ message }}</li></b>
+    <div class="usa-alert usa-alert-info">
```
batemapf: Yeah, this was a shortcut b/c I know the tags aren't being use. Should definitely get this updated later on.

tock/hours/tests/test_views.py
```diff
@@ -1012,3 +1028,11 @@ def test_ReportingPeriodDetailView_current_employee_toggle(self):
...
         )
         self.former_employee
+
+class TestFloatViewIntegration(TestCase):
+    fixtures = FIXTURES
+    def setUp(self):
+        self.userdata = UserData.objects.get(pk=1)
+
+    def test_basic(self):
+        pass
```
batemapf: Hehe! Looks like that whole class may have been a vestige of something earlier. Will get it cleaned up.

tock/tock/settings/base.py
```diff
@@ -10,15 +10,30 @@
...
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
 import os
+import json
 
 from django.utils.crypto import get_random_string
 
+def get_float_key(keyname):
+    if os.environ.get('VCAP_SERVICES'):
+        key = str(json.loads(os.environ.get(
+            'VCAP_SERVICES'))['user-provided'][0]['credentials'][keyname])
```
batemapf: Yup, good point. Will get that in there.

tock/tock/settings/base.py
```diff
@@ -10,15 +10,30 @@
...
 DATABASES = {}
 ROOT_URLCONF = 'tock.urls'
 WSGI_APPLICATION = 'tock.wsgi.application'
 SECRET_KEY = os.environ.get('DJANGO_SECRET_KEY', get_random_string(50))
 
+FLOAT_API_KEY = get_float_key('float-key')
+FLOAT_API_URL_BASE = 'https://api.float.com/api/v1'
+FLOAT_API_HEADER = {'Authorization': 'Bearer ' + FLOAT_API_KEY}
+FLOAT_API_TASK_WEEKS = 1
+FLOAT_API_WEEKDAYS = 5
```
batemapf: Nah, that is all Tock specific. Will get it documented and/or rethink a bit (see above discussion).

tock/tock/templates/hours/timecard_form.html
batemapf: Had to override the styling w/ `<ul style="list-style-type:disc">` after trying the USWDS code line for line based on the docs. Not sure what's going on, but patched for now...

toolness: Hmm, I'll have to take a look at this separately--since I'm working on USWDS now, I gotta make sure it's free of confusing paper cuts like these.

That said, I just noticed that Tock is using USWDS 0.12, so it might have been fixed in a recent revision.  Hmmmmmm!

batemapf: Ah, well then :-). Will try bumping and see if that takes care of it.

tock/employees/models.py
batemapf: Went ahead and just flipped this to an `IntegerField`. Simpler and better approach.

tock/tock/templates/hours/timecard_form.html
```diff
@@ -14,19 +14,22 @@
...
-       <li>-  {{ i.hours_wk }} hours on {{ i.project_name }}
-          {% if i.task_name %}
-          assigned to {{ i.task_name }}
+      <ul style="list-style-type:disc">
+        {% for i in float_data.tasks %}
+         <li>
+           {{ i.hours_wk }} hours on {{ i.project_name }}
+           {% if i.task_name %}
+            assigned to {{ i.task_name }}
           {% endif %}
```
toolness: It seems like this `{% endif %}` is not indented at the same level as the `{% if %}` above it?  or is github's diff view misaligning things?  not really a big deal either way...

tock/tock/utils.py
```diff
@@ -40,21 +41,56 @@ def has_permission(self, request, view):
...
-    testing = 'test' in sys.argv
-    shell = 'shell' in sys.argv
-    if testing or shell:
-        url = '{}/{}'.format(test.FLOAT_API_URL_BASE, endpoint)
+      variables used for testing / shell work with the mock Float API."""
+
+    url = '{}/{}'.format(base.FLOAT_API_URL_BASE, endpoint)
+
+    # If testing, get mock response.
+    if 'test' in sys.argv or 'shell' in sys.argv:
```
toolness: Would it be useful to control this via an environment variable?  That way devs could also use it during manual testing, if they wanted... Anyways, just an idea.

I am a *little* bit wary of doing a test like `if 'somestring' in sys.argv` just because of the possibility that the string might appear somewhere in some other use case and trigger behavior that no one expects--I'd feel a bit more confident in this if it e.g. checked if `manage.py` was in it?  I've typically done something like [this](https://github.com/toolness/hive-django/blob/master/hive/settings_utils.py#L45-L47):

```py
def is_running_test_suite():
    return (os.path.basename(sys.argv[0]) == 'manage.py' and 
            len(sys.argv) > 1 and sys.argv[1] == 'test')
```

tock/tock/templates/hours/timecard_form.html
batemapf: Nope, just some sloppy indenting on my part. Fixed!

tock/tock/utils.py
batemapf: Good call. Will add in `manage.py` to the mix.

tock/hours/views.py
```diff
@@ -617,43 +616,40 @@ def get_object(self, queryset=None):
...
-            'tasks': [ item for sublist in \
-                [ ii for ii in [i['tasks'] for i in task_data['people'] \
-                if i['people_id'] == float_people_id] ] \
-            for item in sublist ]
+            # Get all of the tasks (ii) associated with a Float user if the
+            # user's Float people_id matches their Tock float_people_id.
+            # See https://github.com/floatschedule/api/blob/805663be98c9f48a275c9a13e824b0d6701df398/Sections/tasks.md.
+            'tasks': flatten(
+                [ ii for ii in [i['tasks'] for i in task_data \
+                    if int(i['people_id']) == float_people_id] ]
```
toolness: Er wait... didn't you say that sometimes the Tock API has an empty string for `people_id` or something?  Will that cause the `int()` call to throw a `ValueError`?

toolness:
Cool, this is much better, thanks!  My only major concern is regarding the `int()` call above, other than that I think it's 👍 .

tock/hours/views.py
batemapf: Ah, so a Float user will always have a `people_id` (at least as far as I know...). The biggest hangup here is matching a Tock user's `User.username` with a Float `im` value (it was the only field that can be edited in Float via the GUI; Float admins manually enter a user's `User.username` into this field). _If_, however, Tock cannot match a `User.username` with a Float `im` value, an empty list is returned [here](https://github.com/18F/tock/pull/630/files/a1f789f9703054a0266e577a40ea5ae819437c0c..8843b820bf4ab1fa5fa98c32efa80950c1471d7f#diff-bbf773df48e042c9e0b2f421ddc3ee38R675), with nothing save to the user's UserData.float_people_id (thus avoiding the potential for a `ValueError: invalid literal for int()...`).

I know, a bit confusing.

toolness: Ahh got it! Sorry, my bad, thanks for the clarification.

toolness:
Cool, this looks fine to me!  Feel free to merge at will.

PR #631, opened 2017-06-29
WIP alternate work schedule code. DO NOT MERGE.

annalee:
## Description

Allows admins to set working hours on a user-by-user basis, so that staff on alternate work schedules don't get prompted to log 40 hours each week.

TODO:
[ ] Finish refactoring out previous, non-user-specific working hours code
  [ ] clean up old working hours on admin
  [ ] clean up old working hours tests
[ ] Add new working hours tests
[ ] Improve hours validation?

## Additional information

Include any of the following (as necessary):

* Relevant research and support documents
* Screenshot images
* Notes

PR #634, opened 2017-07-07
Skip max/min hour validation for alt. work schedule eligible users

batemapf:
## Description

Adds a new attribute on the `UserData` model to indicate eligibility for an alternative work schedule (AWS). Defaults to False. If changes to True via admin panel, users will not receive an validation error if trying to submit a timecard that does not meet the max/min criteria for a given reporting period.

This is a more basic implementation of #631.


annalee:
:shipit:

PR #635, opened 2017-07-07
Use holiday and time off data from Float in Tock

batemapf:
## Description

To merge with `float-api-integration` work. This addresses an outstanding need in the underlying Float integration work (noted [here](https://github.com/18F/tock/compare/float-api-integration...add-holiday-consideration?expand=1#diff-bbf773df48e042c9e0b2f421ddc3ee38L652)). These changes ping the Float API for information about scheduled holidays and scheduled time off (Float does not automatically deduct these occurrences from the "task" data). If a holiday and/or timeoff falls during the reporting period, the Float assignment reported to the user in Tock will be adjusted accordingly.

For example, if a user were assigned to a project in Float for 6.5 hours per day for a full week (32.5 hours total) and that week contained a holiday and a 5.0 hours of time off for the user, the Float assignment would be calculated by:

(hours per day assigned * (5 working days - number of holidays)) - (hours of time off / number of assignments)

In the above case, the calculation would be:

(6.5 * 4) - (5.0/1) = 21.0

**Note**: It may make sense to merge the underlying branch with `master`, then tackle this one.

## Additional Information

This change also simplifies the Float fixtures for better clarity.


tock/hours/views.py
```diff
@@ -615,48 +615,107 @@ def get_object(self, queryset=None):
...
+        r = get_float_data(
+            endpoint='holidays',
+            params={'start_day': (self.end_date - dt.timedelta(
+                weeks=2)).isoformat()
+            }
+        )
+        if not r:
+            return None
+        holidays = [ i['date'] for i in r.json()['holidays'] ]
+        holidays_in_period = [ i for i in holidays \
```
brendansudol: L667-668 could be combined (to avoid an additional loop through the days)

tock/hours/views.py
```diff
@@ -615,48 +615,107 @@ def get_object(self, queryset=None):
...
+            endpoint='timeoffs',
+            # Get last 26 weeks of Float timeoff data to guard against
+            # an extreme edge case where someone has a timeoff
+            # starting <= 26 weeks ago.
+            params={
+                'weeks': 26,
+                'start_day': self.end_date - dt.timedelta(weeks=26)})
+        if not r:
+            return None
+        def overlapping_weekdays(sd_one, ed_one, sd_two, ed_two):
```
brendansudol: a simpler, more efficient function:

```
def overlapping_weekdays(sd_one, ed_one, sd_two, ed_two):
    last_start = max(sd_one, sd_two)
    first_end = min(ed_one, ed_two)
    diff = (first_end - last_start).days

    if diff > 0:
        days = [last_start + dt.timedelta(days=i) for i in range(diff + 1)]
        return [d for d in days if d.weekday() < 5]
```

brendansudol:
in general, i think a lot of this float stuff can / should be refactored to live outside of this view which would making testing easier (because you wouldn't have to instantiate the view class and check for stuff in the response) and could instead more directly isolate and unit test the individual functions (i.e., `assertEqual(calculate_workdays(foo, bar), 'baz')`), but i'm happy to merge this as is, and can leave that improvement as an exercise for a future PR

tock/hours/views.py
batemapf: Ugh, that is beautiful. Incorporating.

batemapf:
@brendansudol took your advice and moved all of the Float stuff outside of the view and into its own module. Did a bit of refactoring too, then added in a bunch of unit tests. Mind taking another gander?

tock/hours/float.py
```diff
@@ -0,0 +1,147 @@
...
+
+from employees.models import UserData
+from tock.utils import get_float_data, flatten
+
+# Variables.
+BASE_HOURS_PER_WEEK = 32.5
+BASE_HOURS_PER_DAY = 6.5
+BASE_DAYS_PER_WEEK = 5
+
+def float_tasks_for_view(request, kwargs):
```
brendansudol: instead of passing in the `request` and `kwargs`, since you're only using one thing from each, just pass those in instead (`float_tasks_for_view(user_id, reporting_period)`) -- this way, it's more clear and easier to test (since you wouldn't have to mock a request object)

tock/hours/float.py
```diff
@@ -0,0 +1,147 @@
...
+            work_days=work_days
+        )
+        return get_tasks(
+            start_date=rp.start_date,
+            end_date=rp.end_date,
+            tasks=float_tasks,
+            work_days=work_days,
+            work_hours=work_hours
+        )
+    return None
```
brendansudol: stylistic pref (though i also think it's easier to parse): move the `return None` failsafe above the code that does the meaty stuff:

```
if float_people_id:
    return None

float_tasks = ...
return something_awesome
```

tock/hours/float.py
```diff
@@ -0,0 +1,147 @@
...
+
+def get_tasks(start_date, end_date, tasks, work_days, work_hours):
+    for i in tasks:
```
brendansudol: no need to change, but i tend to reserve `i` to refer to the index of an element in an array, vs the actual element itself (i would use `t` or `task` here)

tock/hours/float.py
```diff
@@ -0,0 +1,147 @@
...
+    return get_float_data(
+        endpoint='holidays',
+        params={'start_day': (start_date - dt.timedelta(
+            weeks=weeks))
+        }
+    ).json()['holidays']
+
+def get_float_people_id(userdata):
+    if userdata.float_people_id:
+        print('foo')
```
brendansudol: ✂️

tock/hours/float.py
```diff
@@ -0,0 +1,147 @@
...
+
+def get_float_holidays(start_date, weeks=2):
+    return get_float_data(
+        endpoint='holidays',
+        params={'start_day': (start_date - dt.timedelta(
+            weeks=weeks))
+        }
+    ).json()['holidays']
+
+def get_float_people_id(userdata):
```
brendansudol: what do you have against blank lines? :)

not changing any of the code, i would, in general, pepper in a few more to make things a bit more legible (in my HUMBLE OPINION, take or leave):

```
def get_float_people_id(userdata):
    if userdata.float_people_id:
        return userdata.float_people_id

    float_id = [
        p['people_id'] for p in get_float_data(endpoint='people').json()['people']
        if p['im'] == userdata.user.username
    ]

    if not float_id:
        return None

    # Use most recent Float people_id in the rare case a
    # person has multiple people_ids.
    userdata.float_people_id = int(float_id[-1])
    return userdata.float_people_id
```

tock/hours/float.py
batemapf: ✔️

tock/hours/float.py
batemapf: But supposing I was iterating through a list representing igloos?

brendansudol:
looks great -- one meaningful comment and 3 fluffy ones (i'll let u decide which)

tock/hours/float.py
batemapf: ✔️

tock/hours/float.py
brendansudol: you salty dog!

tock/hours/float.py
batemapf: ✔️

tock/hours/float.py
batemapf: ✔️

brendansudol:
💥

PR #636, opened 2017-07-10
bumping python version, fixing req's issue, typo

batemapf:
## Description

Fixing some issues I ran into when trying to deploy to staging.


annalee:
:shipit:

Issue #628, opened 2017-04-11
Automatically open this week's timecard


mgwalker:
As a person who needs to track their time, I want the current reporting period to be open on Monday morning so that I can track my time accurately in Tock from the beginning of the week instead of having to either use a separate tool or guess at it later.

batemapf:
@lsgitter Are you cool if we auto-create reporting periods? I'm thinking whoever the first user is to open Tock after the close of the last reporting period (say, @mgwalker first thing on Monday morning 😄) will trigger creation of a standard reporting period. An admin user would still have power to create / amend / delete reporting periods, as well as attach a holiday as needed. We could also add an exception in to _not_ autocreate during the last and first weeks of the fiscal year (since those get a bit wonky).

Thoughts?

cc @joshuabailes @abrouilette

lsgitter:
@joshuabailes is the one who uses them so if he is fine with it, I am.

batemapf:
Closed via #637.

PR #637, opened 2017-07-11
Trigger creation of reporting period if last reporting has ended (usually)

batemapf:
## Description

Addresses #628 by automatically creating a new `ReportingPeriod` if the `end_date` attribute of the latest `ReportingPeriod` has passed. The new `ReportingPeriod` will have a `start_date` that is one day after the `end_date` of the previous reporting period (with it's own `end_date` six days after that).

To avoid situations where a new `ReportingPeriod` is created that includes work week days in more than one fiscal year, auto-creation is disabled for any potential start dates seven days before and after October 1st of any year, as long as October first is _not_ on a weekend. For instance, auto-creation would be disabled for the last week of September 2015 and first week of October 2015. 

Auto-creation will remain active in the last week of September 2017 and first week of October 2017 because October 1st is, in fact, on a weekend.

**NOTE**: If a week includes a holiday, a Tock administrative user will still need to "attach" a `HolidayPrefill` to the auto-created `ReportingPeriod`.

### DIFF
--- a/tock/hours/tests/test_views.py
+++ b/tock/hours/tests/test_views.py
@@ -628,6 +628,83 @@ class ReportTests(WebTest):
         new_rps = hours.models.ReportingPeriod.objects.all()
         self.assertEqual(len(existing_rps), len(new_rps))
 
+    def test_disallowed_dates_safe_dates(self):
+        """ Check that dates are correctly disallowed when auto-creating a new
+        reporting period."""
+        safe_end_dates = [(2011, 9, 28), (2011, 10, 2), (2016, 9, 29),
+            (2016, 10, 3)]
+        ct_existing_rps = len(hours.models.ReportingPeriod.objects.all())
+        for i in safe_end_dates:
+            self.reporting_period.end_date = datetime.date(
+                year=i[0], month=i[1], day=i[2])
+            self.reporting_period.save()
+            response = self.app.get(
+                reverse('ListReportingPeriods'),
+                headers={'X_AUTH_USER': self.regular_user.email},
+            )
+            ct_new_rps = len(hours.models.ReportingPeriod.objects.all())
+            self.assertNotEqual(ct_existing_rps, ct_new_rps)
+
+    def test_disallowed_dates_not_safe_dates(self):
+        """ Check that dates are correctly allowed when auto-creating a new
+        reporting period."""
+        not_safe_end_dates = [(2015, 10, 2), (2015, 9, 30), (2014, 10, 2),
+            (2014, 9, 30)]
+        ct_existing_rps = len(hours.models.ReportingPeriod.objects.all())
+        for i in not_safe_end_dates:
+            self.reporting_period.end_date = datetime.date(
+                year=i[0], month=i[1], day=i[2])
+            self.reporting_period.save()
+            response = self.app.get(
+                reverse('ListReportingPeriods'),
+                headers={'X_AUTH_USER': self.regular_user.email},
+            )
+            ct_new_rps = len(hours.models.ReportingPeriod.objects.all())
+            self.assertEqual(ct_existing_rps, ct_new_rps)
+
+    def test_disallowed_dates_irrelevant_dates(self):
+        """ Check that dates are correctly disallowed / allowed when
+        auto-creating a new reporting period. """
+        irrelevant_end_dates = [(2015, 2, 2), (2015, 11, 30)]
+        ct_existing_rps = len(hours.models.ReportingPeriod.objects.all())
+        for i in irrelevant_end_dates:
+            self.reporting_period.end_date = datetime.date(
+                year=i[0], month=i[1], day=i[2])
+            self.reporting_period.save()
+            response = self.app.get(
+                reverse('ListReportingPeriods'),
+                headers={'X_AUTH_USER': self.regular_user.email},
+            )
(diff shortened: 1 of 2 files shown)


tock/hours/views.py
```diff
@@ -512,17 +512,52 @@ class ReportingPeriodListView(PermissionMixin, ListView):
...
+                end_date=start_date + datetime.timedelta(days=7),
+                max_working_hours=40,
+                min_working_hours=40,
+                exact_working_hours=40
+            )
+
     def get_context_data(self, **kwargs):
         context = super(
             ReportingPeriodListView, self).get_context_data(**kwargs)
+        self.auto_create_reporting_period()
```
brendansudol: i would perhaps rename this function for clarity since most of the time it is not creating a new reporting period; maybe adding a 'maybe' would work :)

btw, if you're curious, here's how i would write this (note that it's nearly identical to yours and definitely doesn't need to be amended)

```
def maybe_create_new_reporting_period(self):
    latest_rp = self.queryset.first()

    if not latest_rp:
        return None

    latest_end = latest_rp.end_date
    new_start = latest_end + datetime.timedelta(days=1)
    today = datetime.datetime.utcnow().date()

    if latest_end < today and new_start not in self.disallowed_dates(latest_end):
        ReportingPeriod.objects.create(
            start_date=new_start,
            end_date=new_start + datetime.timedelta(days=7),
            max_working_hours=40,
            min_working_hours=40,
            exact_working_hours=40
        )
```

tock/hours/views.py
```diff
@@ -512,17 +512,52 @@ class ReportingPeriodListView(PermissionMixin, ListView):
...
+        else:
+            return []
+
+    def auto_create_reporting_period(self):
+        # Automatically creates a new reporting period if the latest reporting
+        # period has concluded.
+        latest_rp = self.queryset.first()
+        if not latest_rp:
+            return None # In case there are no reporting periods created yet.
+        if latest_rp.end_date >= datetime.datetime.utcnow().date():
```
brendansudol: i would pull out and only import the datetime functions you need in this file, i.e., `from datetime import datetime, timedelta, ...` but then you can avoid having to type datetime all over the place

batemapf: Yeah, I hear ya. It is done differently all over the Tock codebase, but the most common is just plain ol' `import datetime`, so I usually leave it alone. But, yeah, it gets pretty tiresome reading...

tock/hours/views.py
batemapf: Good call on the naming and combing the logic.

tock/hours/tests/test_views.py
```diff
@@ -599,6 +599,35 @@ def test_report_list_authenticated(self):
         )
         self.assertEqual(response.status_code, 200)
 
+    def test_auto_create_reporting_period(self):
```
brendansudol: i would also add a test or two that explicitly test your disallowed_date function to make sure the ranges return are what's expected when supplied a few diff dates

tock/hours/views.py
```diff
@@ -512,17 +512,52 @@ class ReportingPeriodListView(PermissionMixin, ListView):
...
 
+    def disallowed_dates(self, date):
+        # If the end of the fiscal year is not on a weekend, then create
+        # a list of dates around (the buffer) the end of the fiscal year
+        # where a reporting period will not be automatically created. This
+        # prevents the automatic creation of a reporting period that has
+        # working week days that span two fiscal years.
+        fy_start_date = datetime.date(year=date.year, month=10, day=1)
+        if fy_start_date.weekday() < 5 and date.month < 10:
+            buffer_days = 7 # A week before and after.
```
brendansudol: this can be refined a bit -- it looks like you start with a number, then create some dates, then use those dates to convert back to a number in the list comprehension. perhaps a little cleaner: `return [fy_start_date + timedelta(days=i) for i in range(-7, 7)]` (negative days work!)

tock/hours/views.py
```diff
@@ -512,17 +512,54 @@ class ReportingPeriodListView(PermissionMixin, ListView):
...
     permission_classes = (IsAuthenticated, )
 
+    def disallowed_dates(self, date):
+        # If the end of the fiscal year is not on a weekend, then create
+        # a list of dates around (the buffer) the end of the fiscal year
+        # where a reporting period will not be automatically created. This
+        # prevents the automatic creation of a reporting period that has
+        # working week days that span two fiscal years.
+        fy_start_date = datetime.date(year=date.year, month=10, day=1)
+        if fy_start_date.weekday() < 5 and date.month < 10:
```
brendansudol: is `date.month < 10` (the second condition) accurate? isn't only September relevant?

batemapf: If the end date of the prior period is November or December, then the fiscal year switch for the calendar year already occurred and we don't need to worry about generating disallowable dates.

batemapf: Ah, should be `<=` 10, needs to include October. I think. Testing now.

brendansudol:
looks great; left small comments above; nice work thinking through the pesky date edges cases

brendansudol:
nice! 🚢

batemapf:
@ccostino I feel like this one should have include an admin toggle for turning auto creation on/off. Well, maybe next time!

PR #640, opened 2017-07-18
Require authentication for all views.

### LINKED ISSUE #614: Former TTS GSA employees can access parts of tock
We've discovered that employees who are past their end date and not listed as "current" can still access parts of tock _if they still have an active GSA email account_. They'll get a 403 if they go to tock.18f.gov, but tock.18f.gov/reports will load.

This only affects users who have been marked as inactive but who have not left GSA. They need to be able to log into gsa's gmail in order to do this.

batemapf:
## Description

Resolves #614 by requiring requests to all views to be made by an authenticated user. For class-based views, relies on the existing `PermissionMixin` and `IsAuthenticated` class from the Django REST Framework. For function-based views, relies on the `user_passes_test` decorator and the `is_authenticated` attribute of an instantiated `User` class.

Requests made by an unauthenticated user will receive a `403` response (either directly if a class-based view, or via to`/` if a function-based view).

**Note**: *All* first-time users with a gsa.gov email address may access Tock. There is no "white list" of 18F or TTS users that the application compares against. This change set only applies to users who are prevented from authentication because their `is_active` attribute on their respective `User` instance is marked `False`. In other words, in order to restrict a gsa.gov email address holder from Tock, they have to first become a user.

Updated all tests to include authentication information in requests.

## Additional information

To get authentication working with certain tests, re-classed them as `WebTests` (from the `django_webtest`) instead of basic Django `TestCase`'s. This change necessitated updating the text/csv content-type decoder relied upon by the tests in question.


ccostino:
@annalee:  I think we might need this for the ATO because this is an access issue.  I'll review this and update it as necessary and let you know if I have any questions!

ccostino:
I'm very curious if this would fix issue #666, or at least put us on the path toward it.

rogeruiz:
Closed by #720

PR #641, opened 2017-07-24
new end date should be six days from start date, not seven

batemapf:
## Description

I noticed that when this week's reporting period auto-created it had a Sunday end date (7/30) instead of a Saturday end date (7/29). Fix was a one liner, switching a `timedelta()` to `6` instead of `7` days, plus a test to verify success.


brendansudol:
looks lovely

PR #644, opened 2017-08-02
Fix delete checkbox label

### LINKED ISSUE #643: [Bug] Checkboxes for deleting items are not working
The labels for the "delete this item" checkbox are missing a `for` attribute and as a result you can't select the checkbox.

Here is what the code looks like currently:
```
<div class="entry-delete usa-width-two-thirds">
  {{ project_entry.DELETE }}
  <label>Delete this item</label>
</div>
```

The code should look something like this:
```
<div class="entry-delete usa-width-two-thirds">
  {{ project_entry.DELETE }}
  <label for="VARIABLE-ID-HERE">Delete this item</label>
</div>
```

batemapf:
## Description

Resolves #643 by building an id from an existing `forloop` variable and some static text, then set it as the `for` value for the label.

### DIFF
--- a/tock/tock/templates/hours/timecard_form.html
+++ b/tock/tock/templates/hours/timecard_form.html
@@ -71,7 +71,7 @@
       {% if unsubmitted %}
       <div class="entry-delete usa-width-two-thirds">
         {{ project_entry.DELETE }}
-        <label>Delete this item</label>
+        <label for="id_timecardobjects-{{ forloop.counter0 }}-DELETE">Delete this item</label>
       </div>
       {% endif %}
     </div>


abrouilette:
@annalee, is this the fix for the Delete-checkbox problem @kimberbat encountered?

waldoj:
FWIW, I am also having this problem, on Chrome.

nicoleslaw:
Plz merge this soon 🙏

abrouilette:
@ccostino is putting on his cape even now.

ccostino:
Will get this in ASAP!

PR #647, opened 2017-08-04
Add browser notifications for 3pm local time on Friday

mgwalker:
## Description

Pop-up a browser notification in browsers that support it, at 3pm local time on Fridays.  Useful for people who leave Tock open all the time.

## Additional information

screen shot 2017-08-04 at 3 00 07 pm

* The notification is currently flagged with `requireInteraction`, which means it requires the user to click to dismiss it.  I enabled that because it's my personal preference, but I recognize it might be more universally preferred for it to just go away on its own after a few seconds (whatever the browser does by default).

* Doesn't take into account folks on alternate work schedules

### DIFF
--- a/tock/tock/static/js/components/notifications.js
+++ b/tock/tock/static/js/components/notifications.js
@@ -1,4 +1,4 @@
-(function setup_notifications() {
+(function setupNotifications() {
   if("Notification" in window) {
     Notification.requestPermission().then(function(result) {
       // If the user grants permission, or has granted permission in the past,
@@ -45,6 +45,8 @@
       };
 
       setTimer();
+    }).catch(function(error) {
+      console.error('error:', error);
     });
   }
 })();
--- a/tock/tock/static/js/components/notifications.js
+++ b/tock/tock/static/js/components/notifications.js
@@ -1,0 +1,50 @@
+(function setup_notifications() {
+  if("Notification" in window) {
+    Notification.requestPermission().then(function(result) {
+      // If the user grants permission, or has granted permission in the past,
+      // setup the timers.  Otherwise bail out.
+      if (result !== 'granted') {
+        return;
+      }
+
+      var setTimer = function() {
+        var now = new Date();
+        var target;
+
+        // If it's Friday and before 3pm... (local times)
+        if(now.getDay() == 5 && now.getHours() < 15) {
+          // ...target is later today
+          var hoursForward = 15 - now.getHours();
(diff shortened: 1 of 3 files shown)


mgwalker:
Going to try to add some tests here, but I don't want to add a bunch of new dependencies so we'll see how well that works out.  :)

abrouilette:
Thank you, @mgwalker.  Love this.  @ccostino, at your discretion.

ccostino:
Thank you, @mgwalker!  I hope to have this in soon, I just want to have one other person more familiar with browser notifications to take a quick look before I merge.

tock/tock/static/js/components/notifications.js
```diff
@@ -0,0 +1,50 @@
+(function setup_notifications() {
```
msecret: JS standards at 18F usually say to camelCase code rather then `_` case. But if there are other standards on this project that's fine too.

tock/tock/static/js/components/notifications.js
```diff
@@ -0,0 +1,50 @@
...
+  if("Notification" in window) {
+    Notification.requestPermission().then(function(result) {
```
msecret: Does this `then` block need a `catch` statement? Or I guess it will go to catch if the user doesn't grant permission, in which case, nothing has to be done?

tock/tock/static/js/components/notifications.js
ccostino: I don't think there are any set standards in this project (at least not yet), but duly noted, thanks @msecret!

tock/tock/static/js/components/notifications.js
ccostino: Looks like it's captured just inside the beginning with this:

```javascript
// If the user grants permission, or has granted permission in the past,
// setup the timers.  Otherwise bail out.
if (result !== 'granted') {
    return;
}
```

mgwalker: It ignores the promise rejection case.  My logic was that there's nothing to do if this fails for some reason.  I guess as a matter of good practice, though, there's no harm in catching the rejection anyway and logging to the console (or whatever the project standard is), since a rejection is exceptional in this case (the promise should always resolve under normal circumstances).

tock/tock/static/js/components/notifications.js
mgwalker: I'll make the change!

ccostino:
Thanks for taking the time to review, @msecret, and thanks for making the minor adjustments, @mgwalker, I definitely appreciate it!

PR #649, opened 2017-08-11
Add CircleCI support.

toolness:
This adds support for running continuous integration on Tock via CircleCI (instead of Travis, which 18F will be dropping support for in coming months).

I've forked this repo to my personal account and you can see it running here: https://circleci.com/gh/toolness/tock/2

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ jobs:
       # documented at https://circleci.com/docs/2.0/circleci-images/
       - image: circleci/postgres:9.5
         environment:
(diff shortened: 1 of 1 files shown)


toolness:
Note that I don't seem to have access to actually add CircleCI support to this repository:

> 

So someone who has access will need to do that (or alternatively, someone could give me admin rights on this repo and I can set it up).

Setup is pretty straightforward, just make sure we're configured to use CircleCI 2.0, rather than the legacy 1.0.

toolness:
Awesome, let me know if you have questions about anything!

Issue #660, opened 2017-08-31
Project alerts not displaying


ccostino:
It seems that with some of the recent UI improvements we may have lost the project alerts from appearing, e.g,. `Out of Office - Annual Leave` should show a message about making sure your time is marked in ALOHA.

ccostino:
This will be fixed with PR #662!

Issue #659, opened 2017-08-27
Unexpected UI behavior: Mouse scrolling changes hours


pburkholder:
![out](https://user-images.githubusercontent.com/1211146/29752586-cffcb67a-8b15-11e7-8581-8bc2c6c9b222.gif)

If I touch my mouse scroll at all while entering hours it will increase/decrease
the hours.

This UI feature continually confounds me: I enter my hours, then I intend to scroll to the next
form entry, and my number of hours for that field increases by some
arbitrary amount.

Can you please remove this feature?

ccostino:
I did a bit of research and this appears to be due to the browser's default behavior/handling of the `number` input form type.  The most common solution I found was to just disable the event on that input element type, so when there's a moment we'll take a look to see if that'll help alleviate it.

Just to confirm, @pburkholder, this is with a mouse with a scroll wheel that you're seeing this behavior with?  Or is it a magic mouse with the swipe action?  I ask because I can kind of simulate the behavior with the touch pad on the laptop, but if I just move the cursor out of the field (even if it still has focus), it goes back to scrolling the page itself.

pburkholder:
@ccostino Thanks for the response. I've had this happen with a magic mouse with swipe, so what you describe with touchpad is what I observe with the magic mouse: "if I just move the cursor out of the field (even if it still has focus), it goes back to scrolling the page itself."

ccostino:
Okay, this bit me today and I almost missed it!  Not sure how as I thought I moved out of focus of the form element, but I noticed a weird thing too where the element comes back into focus after an action.  I definitely share your pain about this.

ccostino:
This will be fixed with PR #662!

Issue #651, opened 2017-08-18
Can't easily search for "general" and/or scroll in project drop-down menu


ccostino:
It would appear the bug that prevents you from searching for things like `General` has resurfaced again, and once you do (and don't see the result), you can't scroll easily in the project drop-down menu.  I'm guessing this is related to #367, too, but am not sure yet.

Note though that if you search for the exact project ID and name, e.g., `2 - General`, it will come up.

hollyallen:
This hit me as well. It was pretty frustrating.

ccostino:
From @leahbannon:

>If I start scrolling through the dropdown list in projects, it speedily jumps to the end. I heard this is happening to someone else, too. This prevents me from using tock if I don't know the exact number and project name to type.
>I'm in Chrome Version 60.0.3112.90 (Official Build) (64-bit)

line47:
Maybe updating the version of [Chosen](https://harvesthq.github.io/chosen/) to the latest would help with this? Alternatively there is another plugin I've used for long dropdown lists called [Select2](https://select2.github.io/examples.html)

ccostino:
Either of those are good place to start - I was definitely going to take a look at upgrading to see, but I'll do a quick comparison to see if Select2 is lighter-weight, more stable, etc.  Thanks, @line47!

ccostino:
In looking into this more today, this speaks to the need of getting better static asset management in place for Tock...

ccostino:
Looks like it was an outdated version of jQuery impacting this.  Fixing coming soon!

ccostino:
This will be fixed with PR #662!

ccostino:
The scrolling issue is fixed, but searching for `General` is not - I uncovered what the cause is though and am discussing it with the Tock dev team.

ccostino:
Fixed via #664!

PR #662, opened 2017-09-08
Fix UI and admin annoyances; start improving static asset management

ccostino:
This changeset fixes several UI annoyances within Tock currently:

* Being unable to search/scroll in the project list correctly
* Missing project alerts from recent UI improvements
* Scrolling numbers in the numeric input fields via the mouse wheel/touch pad
* Creating/editing Profit/Loss Accounting models (the date validation was incorrect)

This includes upgrades to `chosen`, `jQuery`, `node-sass`, and `underscore.js` to their latest respective versions.  It also includes the beginning steps of better static asset management.  Ideally these libraries/frameworks would be managed by `npm`/`yarn` in `package.json` and `package-lock.json`, but that requires adding in static asset management with something like Webpack, which will take more time.  Hopefully this is something we can lean on the frontend guild for when time allows.

This also includes updates to run with Node.js 8.x, which will become the next LTS release, and updated Django to account for the recent security release in the 1.10 series (along with several lower level Python dependencies).


ccostino:
This addresses the following issues:

* #651
* #659 
* #660

ccostino:
This PR includes some minor but extremely useful enhancements to the Admin screen:

* Employees and Projects now show a bunch more info in their listing pages, providing useful info at a quick glance.
* Employees and Projects also now have some filtering capabilities against boolean (True/False, on/off, enabled/disabled, etc.) attributes.
* You can now search for Employees based on their username and `float_people_id`.
* You can now search for Projects based on their `name`, accounting code `code`, and `mbnumber`.

pburkholder:
🎈 🎉

tock/tock/templates/hours/timecard_form.html
```diff
@@ -161,7 +159,7 @@
   }
 
   function displayAlerts(selectBox) {
-    var $fieldset = $(selectBox).parent().parent(),
+    var $fieldset = $(selectBox).parent().parent().parent(),
```
msecret: I think you can pass `parent()` a selector to find a specific parent. For instance:

`$(selectBox).parent('.someClass');`

msecret: Chaining three parent's like this is a little dangerous because now this code relies on the structure of the HTML to function. So if a developer comes along and changes the hierarchy of the HTML here, they would have to reason to know that this piece of code relies on that certain hierarchy.

tock/tock/static/js/chosen/bower.json
```diff
@@ -1,20 +1,6 @@
 {
```
msecret: It'd be ideal if these libs came from npm, but for now, it might help a bit if they were in a `vendor` directory, so at least all contributors know they are vendor libs and not src code.

tock/tock/templates/base.html
```diff
@@ -12,7 +12,7 @@
     <link rel="stylesheet" href="{% static 'css/vendor/jquery-ui.min.css' %}">
     <link rel="stylesheet" href="{% static 'js/chosen/chosen.min.css' %}">
     <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.3.0/css/font-awesome.min.css">
-    <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/2.2.0/jquery.min.js" ></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.2.1/jquery.min.js" ></script>
```
msecret: Another future improvement would be to serve all these together as one file. Also, we generally try and stay away from cdns at 18f, mainly due to security. Is it possible to move the cdn files to the local repo?

ccostino:
Thank you, @annalee and @msecret!

@msecret, I'll try and get this feedback incorporated ASAP as I think these are great improvements that can we can still make now before releasing this. :-)

tock/tock/templates/hours/timecard_form.html
ccostino: Totally agreed and the fact that this did break because of the chaining and change in HTML structure with other recent UI updates speaks to this.  I'll look into specifying a selector to get the right parent without relying on the structure entirely, thanks!

tock/tock/static/js/chosen/bower.json
ccostino: Totally agree, but that's going to be a longer-term solution and ties into the other change suggested as well (combining them all into one file, probably with something like Webpack and just making it a part of the build/deployment process).

tock/tock/templates/base.html
ccostino: Ah good call - I'll pull them in directly!

tock/tock/static/js/chosen/bower.json
msecret: Yea that sounds fine

ccostino:
@annalee and @msecret, if you both wouldn't mind giving this one more pass just to verify things look okay, I'd appreciate it!  I've taken this PR as far as it can go without introducing too many changes at once. :-)  Thanks!

Issue #669, opened 2017-09-28
Users experiencing 500 errors when there's a Float connectivity problem


ccostino:
We are experiencing numerous 500 level server errors at the the moment and it all seems to be due to the fact that there's a problem with the underlying API calls to Float directly.  The strange thing about this is that services themselves all appear to be fine according to Float's status page at the time of this writing.

The code managing the Float data processing does not have good error handling or logging in it currently, so we have no clear visibility into what is actually wrong, nor do we have graceful handling of such errors so that users can still see and enter their timecards.

This needs to be fixed immediately to address the error handling issue and provide visibility into the underlying issue with the Float API calls.

ccostino:
The fix has been deployed and time sheets should be accessible again.  It looks like the Float API key is now invalid for some reason based on the errors I'm seeing in the logs.

ccostino:
And after checking this morning the Float integration is working fine again.  I'm not sure what the issue actually was as the status page indicated everything was okay.

@abrouilette since we now have better error handling at least and don't have to worry about the Float integration making life difficult for folks, do you mind if we close this issue and deal with future problems separate?

abrouilette:
Not at all, sir!

PR #670, opened 2017-09-28
Handling Float API errors gracefully

ccostino:
Fixes issue #669 

This changeset provides much more robust error handling for the Float API calls and more graceful handling of them so that users can still see and enter their timecards.  If an error occurs while accessing Float data, the user is presented with a message saying so and the error is logged behind the scenes so we can see what went wrong and why.  We currently only see a stack trace saying that we cannot call a method on a NoneType object.

There's more to be done here for the future, but at this point we just need to make sure the timecards always work regardless of whether or not retrieving the Float data does and, in the case of such errors, where the breakdown actually happened and why.

### DIFF
--- a/tock/hours/float.py
+++ b/tock/hours/float.py
@@ -1,9 +1,13 @@
 import datetime as dt
+import logging
+import sys
 
 from .models import ReportingPeriod
 
 from employees.models import UserData
 from tock.utils import get_float_data, flatten
+
+logger = logging.getLogger(__name__)
 
 # Variables.
 BASE_HOURS_PER_WEEK = 32.5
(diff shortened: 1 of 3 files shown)


tock/hours/float.py
```diff
@@ -13,43 +17,76 @@
...
-        timeoffs=float_timeoffs,
-        work_days=work_days
-    )
+    try:
+        float_tasks = get_float_tasks(
+            start_date=rp.start_date,
+            float_people_id=float_people_id,
+            weeks=1
+        )
+    except:
```
tbaxter-18f: Did you think about handling the exception a bit more and passing it through? something along the lines of 

```
except Exception as e:
    logger.error(e)
```

I think that's right. If not, it's close.

tock/hours/float.py
```diff
@@ -13,43 +17,76 @@
...
+                start_date=rp.start_date,
+                end_date=rp.end_date,
+                tasks=float_tasks,
+                work_days=work_days,
+                work_hours=work_hours
+            )
+        except:
+            logger.error(sys.exc_info()[0])
+
+    return tasks
```
tbaxter-18f: Would you want to return some feedback that there were errors?

tock/tock/templates/hours/timecard_form.html
```diff
@@ -11,7 +11,7 @@
     </p>
   </div>
 </div>
-{% if float_data %}
+{% if float_data is not None %}
```
tbaxter-18f: Why was this change needed?

tock/tock/templates/hours/timecard_form.html
```diff
@@ -28,7 +28,13 @@
            {% if i.task_name %}
             assigned to {{ i.task_name }}
            {% endif %}
-        </li>
+         </li>
+        {% empty %}
+         <li>We're sorry, there was an error retrieving your data from Float.
```
tbaxter-18f: We could, if we wanted, surface the error here. I can see where that might not be desirable, though. Just a thought.

tock/hours/float.py
tadhg-ohiggins: I agree, we should be catching specific exception types only.

ccostino: You're right in that this isn't ideal, but at the moment we don't necessarily know what the error will be, or at least I'm not 100% confident we will (by and large I'm seeing `attempting to call a method on NoneType objects`, but again, no guarantee).  As far as the handling itself goes, that may be better, let me give it a shot!

tock/hours/float.py
ccostino: See answer to the template change. :-)

tock/tock/templates/hours/timecard_form.html
ccostino: This was needed because there is a state where `float_data` is `None` and that's valid - in that case we don't want to display the Float information box at all.  What this change does is allows me to send an empty list instead and check for that - if there's no data, I can assume there was an error retrieving the information and provide a message.

tock/tock/templates/hours/timecard_form.html
ccostino: Displaying the error here would be useless as this is user facing; folks aren't going to know what it means.  It's on us as the devs/admins to look in the logs and debug from there. :-)

ccostino: Correction:  displaying the raw error here.

tock/tock/templates/hours/timecard_form.html
tbaxter-18f: I think -- but I could be wrong -- that since Django templates are more truthy, an empty list still evaluates to none. So the prior way should still work. But it's probably not worth changing it back, either.

ccostino: Ah I'll have to check!  That said, this is also more explicit. :-)

tbaxter-18f:
Looks like it meets the immediate need to me.

tock/hours/float.py
ccostino: And to add a bit more color here:  the handling of the Float API calls in general could be improved quite a bit.  Again, for the sake of time and stability, this is just to shore things up for the interim.

ccostino:
Thank you for taking a look so quickly, @tbaxter-18f and @tadhg-ohiggins!

PR #673, opened 2017-10-03
Remove Float integration, fix New Relic, and bolster production

ccostino:
This changeset does four things:

* Removes all integration with Float
* Fixes the New Relic integration (for production; staging will require more work)
* Cleans up and adds a bit of extra stability in the production manifest
* Updates Node.js to the current latest release (8.6.0)

The primary purpose is due to the immediate cut-off from using the Float service.  Rather than leave that code lying around, we have opted to remove it in order to help with the ATO process (we can now prove that Tock does not speak to that service at all with none of the code in place).  Should we ever need to turn it back on, we have this PR as a history of the change.

The New Relic fix will help us going forward if there are ever any errors with Tock.  We should be notified about them immediately now instead of relying solely on people to report them.

Lastly, the updates to the production manifest file will help with application stability, namely due to the amount of instances being run for production.  Should any movement happen with the containers/virtual machines under the hood in cloud.gov, Tock should not be impacted at all.  This may have been contributing to some of the 500 errors folks see from time to time as that maintenance can happen at any moment.  There is more work to be done in this regard too, but this gets us much closer to the goal!


ccostino:
The commit message will be slightly off with this:  Node.js was upgraded too.

Also, I've already updated the user-provided `VCAP_SERVICES` in cloud.gov by removing the Float API key and fixing up the New Relic licensing key.

Issue #672, opened 2017-10-02
Add support for manageable timecard notes


ccostino:
Tock doesn't currently provide a way to manage any general notes in a timecard except for modifying the text in the template directly.  We have a need to make this just a tad more customizable to help our operations team, so we're looking to do the following:

* [x] Make a global timecard notes model
* [x] Allow an ops admin to manage multiple notes that would appear on all timecards
* [x] Display all of the timecard notes present in the system at the top of any timecard
* [x] Move the existing note in the template into the timecard notes

ccostino:
@joshuabailes, what types of fields/options would you need for these notes?  I see three things that would be useful:

* A title
* A body for the note itself
* A toggle for enabling/disabling it

Would this cover it?  Do you feel you need anything else?

joshuabailes:
This would. And the ability to edit it without bothering you.

ccostino:
Yup!  These would be in the admin.  Whatever notes you add, all of the enabled ones would show up at the top of any given timecard. :-)

ccostino:
This is now done with PR #677!

@joshuabailes if something doesn't work right for you, please feel free to re-open this issue.

PR #677, opened 2017-10-04
Remove hardcoded note and fix reporting period-specific messages

ccostino:
Finalizes #672 

This changeset finalizes the new timecard notes feature by removing the hardcoded message at the top of the timecard form and making it an initial model available for immediate use.  It also fixes the display of the reporting-period specific messages and provides a few extra options for them as well:

* An optional title
* A style to choose from (same as timecard notes)
* Whether or not the message is actually enabled for display in someone's timecard

These messages will appear above any general messages as they are reporting period specific if they are enabled.  If they are not, they can still be useful for an admin for internal notes.


requirements-dev.txt
```diff
@@ -3,3 +3,5 @@ django-debug-toolbar==1.8
 django-webtest==1.7.9
 factory-boy==2.8.1
 nplusone==0.8.1
+pdbpp==0.9.1
+ipython==6.2.1
```
annalee: Do we need these in dev requirements?

ccostino: Nope!  I find them incredibly useful for helping with debugging and development, but if you rather I keep it local, happy to take them back out.

annalee: I find them useful too, but probably best to keep them local so as not to annoy folks who have a different system. Plus fewer dependencies = easier ATO.

ccostino: Out they go!

PR #684, opened 2017-11-22
Enable FY18-only reports

ccostino:
This changeset adds links to make FY18 data as a standalone.  Note that a more in-depth fix will need to be made to ensure that things like the FY17 reports contain just FY17 data as right now they will still pull in FY18 data as well.

### DIFF
--- a/tock/tock/templates/hours/reports_list.html
+++ b/tock/tock/templates/hours/reports_list.html
@@ -10,22 +10,27 @@
 		<li>Complete timecard data:
 			<a href="{% url 'reports:BulkTimecardList' %}">All</a>
 			<a href="{% url 'reports:BulkTimecardList' %}?after=2016-10-02">FY2017</a>
+            <a href="{% url 'reports:BulkTimecardList' %}?after=2017-10-01">FY2018</a>
 		</li>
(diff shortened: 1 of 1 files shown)


ccostino:
@joshuabailes and @abrouilette, I just want to confirm the dates are right!  And as noted in the description, there is still an issue with the reports pulling in more data than they should, but that's a larger effort fix (which I believe [to be here](https://github.com/18F/tock/blob/master/tock/api/views.py#L163) - mental note for myself later on!).

ccostino:
@annalee good call!  I'd add that as a part of the deeper cleanup/refactoring that has to happen anyway to ensure that the reports return just the data set for the requested fiscal year, not the requested fiscal year plus anything else after. :-)

ccostino:
Actually one thing I was thinking about to help manage that was creating a very light-weight model that tracks fiscal years (essentially just a label, start date, and end date) that Team Ops could manage, and that in turn could be used to generate these links automatically.

annalee:
@ccostino I like the idea of sticking it in a model. I was thinking of using a function to calculate fiscal years because _in theory_ their start and end dates are consistent, but of course, in programming as in life, nothing about time is ever consistent.

Issue #689, opened 2017-12-18
Add a 'department' filter for employees


annalee:
Admin users should be able to filter employees by their place in TTS, pulling only 18F or only Acq employees as needed.

This department label should appear on downloaded spreadsheets, as the LAST column for a given user.

cc @joshuabailes @abrouilette

annalee:
We could do this either with a table or with an old-fashioned choices field. [Insert joke here about how many reorgs it takes to justify a table for this].

ccostino:
We caught up about this some more, we're leaning toward just going with the model because of these reasons:

* We can leverage it in multiple spots (`Employee`, `Project`, etc.)
* Admins can manage it directly with no code changes

ccostino:
I knew we discussed this before!  See #668 - this work should help fill the need for both issues. :-)

tbaxter-18f:
I don't think this is needed anymore. If there's disagreement, we can happily reopen.

Issue #668, opened 2017-09-27
Add Office Data Element to Tock Lines


joshuabailes:
I am requesting a data element be added to each Tock Line created that specifies the office responsible for that line. Those offices are 18F, Acquisitions, OPP, and PIF. I'd like this information to show up in the reports on the dash, and if not all of them, at least these: 

Aggregate hourly data by project and reporting period: All FY2017

Aggregate hourly data by user, reporting period, and project billable status:

Thanks!

abrouilette:
Right now, we need to throw all the dev-time we have at broken things and the ATO.  We'll come back to this feature when we can.

ccostino:
Based on another discussion around holiday pre-fills and the fact that they can only be associated with a single Tock line (project), this is speaking to a broader addition of organization information.

What it sounds like we need is a way to manage internal organizations in Tock, and then associate an organization with projects and employees (Tock users).  The link to a project will then allow for a holiday pre-fill to work for a specific organization as well.

Additionally, we could theoretically pull projects for a timecard based on a user's organization and only show projects (Tock lines) that are of the same org.  One caveat though is someone may be Tocking against something in an org outside of their own, so perhaps this is not a good direction to take this in.

ccostino:
Oh, and if OPP doesn't use Tock, would they still have to be added as an org?  Though if this is exposed through the admin panel, then no need to worry about it because if they do it could be easily added by an admin.

joshuabailes:
OPP may use Tock at some point. As long as we can add later we are fine.

Joshua Bailes
COO/Director of Operations
Joshua.Bailes@gsa.gov
202-550-6659
18F.gsa.gov <https://18f.gsa.gov/>

ccostino:
Okay up, if implemented the way I'm describing that would definitely be possible!

ccostino:
This should be ready to go with the next round of updates that get released!  We should plan to get together beforehand to do a final review of it.

PR #695, opened 2017-12-22
Add support for organizations

### LINKED ISSUE #678: Fix time/date dependency in test
[One of our tests](https://github.com/18F/tock/blob/master/tock/hours/tests/test_views.py#L694-L709) has a nasty little gotcha if it's run at a certain point in the day.  The fix should be to just use specific dates so the test runs consistently.

ccostino:
Address #689 and #668, and fixes #678 

This changeset adds support for organizational units into Tock.  It introduces an `Organization` model, which is intended to be associated with anything that is managed, owned, controlled, etc., by an organization (e.g., an employee belongs to an organization, a project is managed by an organization, etc.).  At present an organization can be associated with an employee and a project (Tock line) with this change.  The associations and organizations themselves can all be managed in the admin panel, which now includes support for filtering by organization in the case of employees and projects.

Furthermore, this also updates all of the reports to include organization information in their output, usually as the last column(s).  In cases where both employee and project information is contained in the report, the organization information for both is included and explicitly labeled.

Lastly, this change fixes a long-standing issue with one of our tests.


ccostino:
@annalee you're absolutely right about the cleanup changes, they should be a separate PR.  I was a bit pressed for time today in trying to get this done so I rolled it all together but I think it's good to be more vigilant about that, thanks for calling it out. :-)

annalee:
Datetime! 🤦‍♀️

Issue #687, opened 2017-12-18
Pre-populate expected hours on time cards


annalee:
When users load a new timecard, they should see the hours they are scheduled to work pre-populated, and be able to modify hours from there.

Tock admins should be able to modify users' scheduled hours in the admin interface, both in the employee admin panel and the project admin panel.

Lastly, we'll want to add a link to the new [Tock guidance in the handbook](https://handbook.18f.gov/tock/) to the timecards so people are aware of the changes and what they're supporting.

cc @joshuabailes @abrouilette

annalee:
@ccostino and I discussed and we're leaning towards doing a model that tracks a relationship between a person and a project, and the hours assigned:

Fields:
Person - manytomany
project - manytomany
hours - int

Then set the timecards to generate by pulling the given user's projects and hours from that table.

ccostino:
In thinking about this more, each record in this table is a mapping of a single person with a single project.  Therefore, the employee and project foreign key relationships should be defined as one to many, but the table housing all of these records will certainly have multiple mappings between the same person and multiple projects and vice versa.

annalee:
Yup, exactly this--we're basically doing a manual version of Django's existing behavior of making tables to map many-to-manys. Since we also need to know assigned hours, we can't use the existing pass-through, but we can mimic the behavior.

ccostino:
Really need a :high_five: emoji reaction here. :-)

ccostino:
Whoops, forgot to close this out now that we finished! :tada:

PR #696, opened 2017-12-29
Add support for timecard prefill data

ccostino:
Implements #687 

This changeset adds support for managing staff assignments to projects in order to prefill timecards for users.  It adds a new model that tracks the assignment of a user to a project, the number of hours they are assigned, and any additional notes about the particular assignment.  Timecards will take this data and prefill existing entries with the information, adding the Tock lines (projects) if need be but also preserving the existing functionality of loading all Tock lines from the previous week.


ccostino:
@annalee please feel free to update the description and title of this PR as things get done!

tock/hours/models.py
```diff
@@ -1,15 +1,15 @@
...
-from employees.models import EmployeeGrade
-from django.core.validators import MaxValueValidator
 from django.core.exceptions import ValidationError
+from django.core.validators import MaxValueValidator
 from django.db import models
 from django.db.models import Q, Max
 
+from employees.models import EmployeeGrade, UserData
+from projects.models import ProfitLossAccount, Project
+from .utils import ValidateOnSaveMixin
```
tbaxter-18f: +1 on cleaning up imports here

tock/hours/models.py
```diff
@@ -330,3 +333,26 @@ def save(self, *args, **kwargs):
...
+    employee = models.ForeignKey(UserData, related_name='timecard_prefill_data')
+    project = models.ForeignKey(Project, related_name='timecard_prefill_data')
+    hours = models.DecimalField(
+        decimal_places=2,
+        max_digits=5,
+        help_text='The amount of hours this user is assigned to this project.'
+    )
+    notes = models.TextField(
+        blank=True,
+        default='',
```
tbaxter-18f: Super minor, but is minor required here?

tock/hours/tests/test_models.py
```diff
@@ -200,13 +200,21 @@ def setUp(self):
...
+        self.reporting_period2 = ReportingPeriod.objects.create(
             start_date=datetime.date.today() - datetime.timedelta(days=7),
             end_date=datetime.date.today()
         )
         self.timecard = Timecard.objects.create(
             user=self.user[0],
             reporting_period=self.reporting_period
         )
+        self.timecard2 = Timecard.objects.create(
+            user=self.user[0],
```
tbaxter-18f: This may just be me, but I'd modify L195 to always return [0] from get_or_create. I think self.user[0] could read as though there is a user_set, which is just weird. If you need self.user[1], I'd just assign it to another var.

tock/hours/views.py
```diff
@@ -800,26 +802,30 @@ def last_timecard(self):
 
     def prefilled_formset(self):
         timecard = self.last_timecard()
-        project_ids, extra = [], 1
+        prefills = dict(TimecardPrefillData.objects.filter(
```
tbaxter-18f: I honestly don't know, but would using values() rather than values_list save you some further conversion to a dict here?

tock/hours/views.py
```diff
@@ -800,26 +802,30 @@ def last_timecard(self):
...
+            employee=self.request.user.user_data).values_list(
+            'project', 'hours'))
+        project_ids = []
+        extra = 1
         if timecard:
-            project_ids = set(
+            project_ids = [
                 tco.project_id for tco in
                 timecard.timecardobjects.all()
```
tbaxter-18f: if you're just extracting project_id, then this looks like a great use for values_list rather than iterating?

tock/hours/views.py
```diff
@@ -800,26 +802,30 @@ def last_timecard(self):
...
             .get(start_date=self.kwargs['reporting_period'])
 
-        init = []
         if rp.holiday_prefills:
-            init += [
-                {'hours_spent': hp.hours_per_period, 'project': hp.project.id}
-                for hp in rp.holiday_prefills.all()
-            ]
-        init += [{'hours_spent': None, 'project': pid} for pid in project_ids]
+            for hp in rp.holiday_prefills.all():
```
tbaxter-18f: A comment might be good here. It starts to get a bit unclear what's going on

tock/hours/models.py
ccostino: I'm not sure what you're referring to, @tbaxter-18f, do you mind expanding your question a bit more?  Are you asking if this field is required?

tock/hours/tests/test_models.py
ccostino: Ah okay, I'll try and adjust a bit here for clarity!  And for future reference, in Tock parlance the time period is the reporting period, which must be created first before any timecards can be generated.  The timecards in turn are essentially a container for a reporting period tied to a specific user that contains one or more Tock lines (timecard objects), which are the mappings of hours worked to a project for a user.  Does that help clear up some of the confusion at all?

tock/hours/views.py
ccostino: Ha, it's funny you bring this up because @annalee and I went back and forth on this very piece!  I'll take another quick look just to see, but for the moment this at least works. :-)

tock/hours/views.py
ccostino: I'll double check on that!

tock/hours/views.py
ccostino: Duly noted, thanks!

tock/hours/models.py
tbaxter-18f: Nothing that large. I was just saying I don't think you need to have `default=''`. But I don't think it's hurting anything, either.

tock/hours/tests/test_models.py
tbaxter-18f: It does. I'd still like to see a better var name than timecard2, if we can think of one, though

tock/hours/views.py
ccostino: Ah I know why we did this - we needed the project ID itself to serve as the key in this dictionary, not the field name itself that we were pulling.

PR #707, opened 2018-01-22
Revert Python version for CF buildpack

ccostino:
It turns out that the latest Python buildpack for Cloud Foundry is not deployed in cloud.gov just yet, so we need to go back one patch level version for Python.

### DIFF
--- a/runtime.txt
+++ b/runtime.txt
@@ -1,1 +1,1 @@
-python-3.6.4
+python-3.6.3


runtime.txt
```diff
@@ -1 +1 @@
-python-3.6.4
+python-3.6.3
```
jmcarp: @rogeruiz: wdyt about using `python-3.6.x` so that the buildpack picks the latest minor version on stage?

rogeruiz: great idea. I'll update this. thanks!

ccostino: I didn't know you could do this, thanks for the heads up @jmcarp!

PR #709, opened 2018-01-23
[wip] Reduce multiple migrations on deployments

rogeruiz:
Running into this when deploying the latest work from the Jingle Sprint. The migrations are running for each instance type, and we don't want that as migrations should only run once.

 ## Acceptance Criteria

- [ ] Push a multi-instance version of Tock to staging and see the migrations only running once.
- [ ] Migrations only run on the zeroth instance.
- [ ] Migrations that take a long time (longer than three minutes) are run as tasks and this is documented.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,1 @@
-web: cd tock && python manage.py migrate --settings=tock.settings.production --noinput && python manage.py collectstatic --settings=tock.settings.production --noinput && gunicorn -k gevent -w 2 tock.wsgi:application
+web: bin/run.sh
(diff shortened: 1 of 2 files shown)


rogeruiz:
@ccostino and I found this issue when we were deploying a migration that could only run once and took a long time. We we seeing it run across multiple instances of the applications. The [Cloud Foundry docs mention](https://docs.cloudfoundry.org/devguide/services/migrate-db.html#frequent-migration) that migrations should be run only on a single instance.

rogeruiz:
Flagging these as bugs since this actually delayed the production deployment yesterday and could possibly delay the deployment in the future if we run into similar issues. But it's not directly ATO work, so it's a lower priority for me.

annalee:
We are turning this into an issue.

PR #712, opened 2018-01-24
Continuous deployment strategy

rogeruiz:
This PR includes the changes needed for the ATO around continuous deployment.

 ## Acceptance Criteria

- [x] Staging is automatically deployed whenever commits are made to `master` branch
- [x] Production is automatically deployed whenever there is a new `git tag` is made. Latest tag gets deployed.
- [x] ~~Footer includes version that is deployed with link to GitHub release page and notes~~ (making this another feature as a nice-to-have)
- [x] Documentation exists for this deployment process
- [x] Deployment is agnostic to occurring on CircleCI ~~or locally~~ using `cf autopilot`.


manifest-production.yml
```diff
@@ -0,0 +1,17 @@
+---
+inherit: base-manifest.yml
```
jmcarp: Manifest inheritance was deprecated recently: https://docs.cloudfoundry.org/devguide/deploy-apps/manifest.html#deprecated. The new recommendation is to use yaml anchors for DRY.

bin/run.sh
```diff
@@ -0,0 +1,9 @@
+#/bin/sh
+
+set -o errexit
+set -o pipefail
+
+cd tock
+python manage.py migrate --settings=tock.settings.production --noinput
```
jmcarp: We can wrap this line with something like `if [ "${CF_INSTANCE_INDEX"} == "0" ]` to limit migrations to the 0th instance.

rogeruiz:
Alrighty, this is deploying now, but the scripts are a mess. I'm pulling a 18FGSA docker image for this instead of rolling my own deployment script, but that also means I need to modify the config yaml a bit. I'll get to it in the morning. but thankfully this is less code we have to maintain here!

Thanks and props to @adborden :tada

adborden:
Not sure if you want to support deploying from macOS or not. If _not_, the docker image already includes a `cf_deploy.sh` script with autopilot installed and a [CI config](https://github.com/18F/cloud-foundry-cli#circleci) that should work for most cases with a little tweaking.

ccostino:
Interesting, both CircleCI jobs finished successfully but the PR is still waiting for a CircleCI status.  I'm not sure what's hung up.

rogeruiz:
Alrighty, this is ready for review. Thanks folks! 🎉

.circleci/config.yml
```diff
@@ -62,3 +62,77 @@ jobs:
...
+      - run:
+          name: deploy Tock Production to cloud.gov
+          command: |
+            export CF_DEPLOY_USER=$CF_DEPLOYER_USERNAME_PRODUCTION
+            export CF_DEPLOY_PASSWORD=$CF_DEPLOYER_PASSWORD_PRODUCTION
+            cf_deploy.sh tock gsa-18f-tock staging manifest-production.yml
+
+workflows:
+  version: 2
+  build_pull_requests:
```
ccostino: Do we need to specify building pull requests?  Don't these happen automatically?  Or is it better to specifically denote?

.circleci/config.yml
```diff
@@ -62,3 +62,77 @@ jobs:
...
+            branches:
+              ignore: /.*/
+      - deploy_to_production:
+          requires:
+            - build
+          filters:
+            tags:
+              only: /v20[1-9][0-9][0-9]+\.[0-9]+/
+            branches:
+              ignore: /.*/
```
ccostino: Should we be explicitly specifying only the `master` branch here too for production?  While this is not a correct use case, it could be possible for someone to accidentally apply a version tag on another branch, and this would pick up on the tag, wouldn't it?

bin/run.sh
```diff
@@ -0,0 +1,21 @@
...
+else
+  echo "Migrations did not run."
+  if [[ -v CF_INSTANCE_INDEX ]]
+  then
+    echo "CF Instance Index is ${CF_INSTANCE_INDEX}."
+  fi
+fi
+
+python manage.py collectstatic --settings=tock.settings.production --noinput
+gunicorn -t 120 -k gevent -w 2 tock.wsgi:application
```
ccostino: Thank you!  So much easier to read and maintain. :-)

docs/deploy.md
```diff
@@ -0,0 +1,205 @@
...
+plugin for Cloud Foundry, which is used for zero-downtime deploys.
+You can install via
+`cf install-plugin autopilot -f -r CF-Community`.
+
+Tock is deployed to the GovCloud instance of cloud.gov. You will need to login
+to via the GovCloud api of cloud.gov:
+`cf login -a api.fr.cloud.gov --sso`
+
+Then target the org and space you want to work with. For example, if you wanted to work with the dev space:
+`cf target -o gsa-18f-tock -s dev`
```
ccostino: Do we have a `dev` space now, or just `staging` and `prod` still?

docs/deploy.md
```diff
@@ -0,0 +1,205 @@
...
+- Organization: `gsa-18f-tock`
+- Spaces: `staging`, `prod`
+- Apps:
+  - `staging` space:
+    - `tock-staging`
+  - `prod` space:
+    - `tock`
+- Routes:
+  - tock.app.cloud.gov -> `staging` space, `tock-staging` app
+  - calc-staging.app.cloud.gov -> `staging` space, `calc-staging` app
```
ccostino: We don't need `calc` here, right? :-)

docs/deploy.md
```diff
@@ -0,0 +1,205 @@
...
+cf cups tock-credentials -p credentials-<ENVIRONMENT>.json
+cf bind-service <APP_INSTANCE> tock-credentials
+cf restage <APP_INSTANCE>
+```
+
+You can update the user-provided service with the following commands:
+
+```sh
+cf uups tock-credentials -p credentials-staging.json
+cf restage <APP_INSTANCE>
```
ccostino: For zero-down time, we should suggest also redeploying the app, e.g., rebuilding the last production build in CircleCI.

docs/deploy.md
```diff
@@ -0,0 +1,205 @@
...
+
+The staging server updates automatically when changes are merged into the
+`master` branch. Check out the `workflows` sections of
+the [CircleCI config](../.circleci/config.yml) for details and settings.
+
+Should you need to, you can push directly to tock.app.cloud.gov with:
+
+```sh
+cf target -o gsa-18f-tock -s staging
+cf push -f manifest-staging.yml
```
ccostino: `cf zero-downtime-push` instead?

docs/deploy.md
```diff
@@ -0,0 +1,205 @@
...
+```
+-->
+
+### Logs
+
+Logs in cloud.gov-deployed applications are generally viewable by running
+`cf logs <APP_NAME> --recent`
+
+Note that the web application and the `rq` worker application have separate
+logs, so you will need to look at each individually.
```
ccostino: Tock has no `rq` application.  Yet, anyway. :-)

.circleci/config.yml
ccostino: Discussed this and because we're trying to continuously deploy to `staging` from the `master` branch, we need to make sure CircleCI is still building every PR too.  There is still room for improvement, but this puts us on the right path!

.circleci/config.yml
ccostino: We'll have to research this in the future for a couple of reasons:

* What if someone is being intentional about applying a tag elsewhere?
* Circle's docs are a bit ambiguous about only building when a tag is applied and pushed - we're trying to enforce a tag on `master` only here, but it seems like the filters do not interact in the way that we would expect.

As is, this works with the expected behavior, so we'll leave for now!

PR #713, opened 2018-01-24
Update timeout for gunicorn

rogeruiz:
This timeout is arbitrarily set at two minutes to allow for long API requests for timecard data that may take longer than 30 seconds. 30 seconds was the default coming from gunicorn after #708

## Acceptance criteria.

- [x] Allow for pulling timecard data that takes longer than 30s to process. https://gsa-tts.slack.com/archives/C1JFYCX3P/p1516811227001124?thread_ts=1516808162.000966&cid=C1JFYCX3P

### DIFF
--- a/manifest.yml
+++ b/manifest.yml
@@ -2,6 +2,9 @@
 applications:
 - name: tock
   instances: 4
+  routes:
+  - route: tock.18f.gov
+  - route: tock.18f.gov/api
 buildpack: python_buildpack
 env:
   NEW_RELIC_APP_NAME: Tock (Production)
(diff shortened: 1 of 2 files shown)


manifest.yml
```diff
@@ -2,6 +2,9 @@
 applications:
 - name: tock
   instances: 4
+  routes:
+  - route: tock.18f.gov
+  - route: tock.18f.gov/api
```
ccostino: Should we have these defined in here, or should they be managed elsewhere?  I ask because I've not seen the routes defined for FEC in production, I believe we use the CDN service.

rogeruiz: These are defined here because of how the org / space was setup in the first place for cloud.gov. The `gsa-18f-tock` org is the owner of the `18f.gov` domain and needs to have these explicitly defined in the manifest. Otherwise, CF on push attempts to assign it the default `app.cloud.gov` domain which we don't want as we use this for the staging URL right now.

Tock doesn't use the CDN service because the domain it uses is owned by the org it's being deployed in.

Additionally, the reason we're setting the `routes:` list here is because all other ways to define `host:`, `path:`, & `domain:` [have been deprecated](https://docs.cloudfoundry.org/devguide/deploy-apps/manifest.html#-attribute-routes-replaces-domain,-domains,-host,-hosts,-and-no-hostname)

ccostino: Gotcha, thank you for explaining!

Issue #652, opened 2017-08-19
Compliance and ATO Prep - Bring Tock up to Code


ccostino:
Tock is in need of some help in order to be brought up to code, so to speak.  Several compliance-related items must be taken care of to prep for an ATO (and some ought to be done regardless).  These are listed below (and will likely be updated a fair amount), though some may be broken out into their own issues if warranted.

* [ ] Switch to Circle CI from Travis CI (PR #649 is a start, courtesy of @toolness!)
* [ ] Automate deployment with [`autopilot` CloudFoundry plugin](https://github.com/contraband/autopilot) (issue #548 addresses this separately)
* [ ] Make code reviews mandatory for PRs against `master`
* [ ] Update New Relic agent and ensure proper notifications are sent out (create a `tock-dev` Google Group to receive notifications?)
* [ ] Shore up Python dependency installation and management with [`pip-tools`](https://github.com/jazzband/pip-tools/) (including updating all dependencies to latest versions)
* [ ] Introduce and document a structured release process to be tied in with automated deployments (e.g., with [`git flow`](https://github.com/nvie/gitflow) and possibly a tool like [`invoke`](http://www.pyinvoke.org/))
* [ ] Enable automatic app refreshes (not sure if Circle CI supports this as of yet; Travis CI does with `cron` jobs you can schedule)
* [ ] Introduce a tool like [`bandit`](https://wiki.openstack.org/wiki/Security/Projects/Bandit) for static code analysis to help manage security concerns
* [ ] Update Node packages and ensure they run with the latest version of Node and NPM (or [`yarn`](https://yarnpkg.com/en/))
* [ ] Ensure cloud.gov services bound to Tock are set and configured properly
* [ ] Verify logging setup and make adjustments as necessary, especially for any user actions

As mentioned, there might be other things in the future that I could be missing at the moment.  We'll update this as we learn more!

abrouilette:
@hollyallen, FYI.

ccostino:
Switching from Circle CI to Travis CI is underway - I merged @toolness' PR and have another one in the wings to shut off Travis: #654

ccostino:
Just made code reviews mandatory for any merges made to the `master` branch, mirroring FEC's setup (and with an added bonus of ensuring that code is up to date and status checks pass).

hollyallen:
Carlo, please tell me you mean switching to Circle CI. :)

ccostino:
[...]

PR #719, opened 2018-02-06
Add link to latest release in footer

rogeruiz:
## Description

Helpful after #652 and #712 to show what version is currently deployed to staging and production. 

## Acceptance criteria 

* Check the footer for a link to the mainline branch if on staging 
* Check the footer for a link to a GitHub release if on production

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -113,7 +113,10 @@ workflows:
               only: /.*/
   build_and_deploy_staging:
     jobs:
-      - build
+      - build:
+          filters:
+            branches:
+              only: master
       - deploy_to_staging:
           requires:
             - build
--- a/tock/tock/context_processors.py
+++ b/tock/tock/context_processors.py
@@ -6,8 +6,8 @@ def version_url(request):
     base_url = 'https://github.com/18F/tock/'
     name = settings.VERSION
     response = {
-        'release_url': '%stree/%s' % (base_url, name),
-        'release_name': settings.VERSION,
+        'x_tock_release_url': '%stree/%s' % (base_url, name),
+        'x_tock_release_name': name,
     }
     if re.match('/v20[1-9][0-9][0-9]+\.[0-9]+/', name):
         response['release_url'] = '%sreleases/tag/%s' % (base_url, name)
--- a/tock/tock/context_processors.py
+++ b/tock/tock/context_processors.py
@@ -1,0 +1,13 @@
+def version_url(request):
+    import re
+    from django.conf import settings
+    base_url = 'https://github.com/18F/tock/'
+    name = settings.VERSION
+    response = {
+        'release_url': '%stree/%s' % (base_url, name),
+        'release_name': settings.VERSION,
+    }
+    if re.match('/v20[1-9][0-9][0-9]+\.[0-9]+/', name):
+        response['release_url'] = '%sreleases/tag/%s' % (base_url, name)
+
+    return response
--- a/tock/tock/context_processors.py
+++ b/tock/tock/context_processors.py
@@ -1,6 +1,8 @@
+import re
+from django.conf import settings
+
+
 def version_url(request):
-    import re
-    from django.conf import settings
     base_url = 'https://github.com/18F/tock/'
     name = settings.VERSION
     response = {
--- a/tock/tock/context_processors.py
+++ b/tock/tock/context_processors.py
@@ -9,6 +9,9 @@ def version_url(request):
         'x_tock_release_url': '%stree/%s' % (base_url, name),
         'x_tock_release_name': name,
     }
+
+    # Regular expression here matches the regular expression found in
+    # .circleci/config.yml for filtering on tags for Production deployments
     if re.match('/v20[1-9][0-9][0-9]+\.[0-9]+/', name):
         response['release_url'] = '%sreleases/tag/%s' % (base_url, name)
 
--- a/tock/tock/settings/base.py
+++ b/tock/tock/settings/base.py
@@ -55,6 +55,7 @@ TEMPLATES =  [
                 'django.template.context_processors.tz',
                 'django.contrib.messages.context_processors.messages',
                 'django.template.context_processors.request',
+                'tock.context_processors.version_url',
             ],
         },
     },
@@ -104,3 +105,5 @@ REST_FRAMEWORK = {
         'rest_framework.authentication.TokenAuthentication',
     ),
 }
+
+VERSION = env.get_credential('CIRCLE_TAG', 'master')
--- a/tock/tock/templates/base.html
+++ b/tock/tock/templates/base.html
@@ -121,5 +121,9 @@ m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
       {% block js %}
       {% endblock js %}
 <div>
+  <footer>
+    <p><i>The current release is <code>{{ release_name }}</code>, <a href="{{ release_url }}" title="{{ release_url }}">view this release on GitHub</a></i>.</p>
+  </footer>
+
   </body>
 </html>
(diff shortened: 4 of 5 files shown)


rogeruiz:
Oh. Looks like I will need to update all the settings.py files so tests pass.

.circleci/config.yml
```diff
@@ -113,7 +113,10 @@ workflows:
               only: /.*/
   build_and_deploy_staging:
     jobs:
-      - build
+      - build:
+          filters:
+            branches:
+              only: master
```
rogeruiz: Leaving this in this PR as I don't feel like it merits it's own PR.

tock/tock/context_processors.py
```diff
@@ -0,0 +1,13 @@
...
+    base_url = 'https://github.com/18F/tock/'
+    name = settings.VERSION
+    response = {
+        'release_url': '%stree/%s' % (base_url, name),
+        'release_name': settings.VERSION,
+    }
+    if re.match('/v20[1-9][0-9][0-9]+\.[0-9]+/', name):
+        response['release_url'] = '%sreleases/tag/%s' % (base_url, name)
+
+    return response
```
rogeruiz: This will return a `response` consisting of the following:

```py
# if $CIRCLE_TAG doesn't exist or is equal to anything other than a release tag.
response = {
    'release_url': 'https://github.com/18F/tock/tree/master',
    'release_name': 'master',
}

# if $CIRCLE_TAG meets our release tag convention from the CI/CD PR.
response = {
    'release_url': 'https://github.com/18F/tock/releases/tag/v20180207.1',
    'release_name': 'v20180207.1',
}
```

tock/tock/settings/base.py
```diff
@@ -104,3 +105,5 @@
         'rest_framework.authentication.TokenAuthentication',
     ),
 }
+
+VERSION = env.get_credential('CIRCLE_TAG', 'master')
```
rogeruiz: Defaults to `master` since staging auto-deploys anything on that branch.

tock/tock/templates/base.html
```diff
@@ -121,5 +121,9 @@
       {% block js %}
       {% endblock js %}
 <div>
+  <footer>
+    <p><i>The current release is <code>{{ release_name }}</code>, <a href="{{ release_url }}" title="{{ release_url }}">view this release on GitHub</a></i>.</p>
```
rogeruiz: Reads like,

_The current release is `v20180202.1`, [view this release on GitHub](https://github.com/18F/tock/releases/tag/v20180202.1)_.

tock/tock/context_processors.py
```diff
@@ -0,0 +1,13 @@
+def version_url(request):
+    import re
+    from django.conf import settings
```
tbaxter-18f: It's not a big deal, but is there a reason the imports are here rather than up top as PEP-8 would usually have them?

tock/tock/context_processors.py
```diff
@@ -0,0 +1,13 @@
...
+    base_url = 'https://github.com/18F/tock/'
+    name = settings.VERSION
+    response = {
+        'release_url': '%stree/%s' % (base_url, name),
+        'release_name': settings.VERSION,
```
tbaxter-18f: Why not use `name` here, too, since you bothered to define it?

tock/tock/context_processors.py
```diff
@@ -0,0 +1,13 @@
...
+    }
+    if re.match('/v20[1-9][0-9][0-9]+\.[0-9]+/', name):
```
tbaxter-18f: Rather than regexing the tag pattern, would it be adequate (and clearer) to just say `if name.startswith('v20'):`

tock/tock/settings/base.py
```diff
@@ -55,6 +55,7 @@
                 'django.template.context_processors.tz',
                 'django.contrib.messages.context_processors.messages',
                 'django.template.context_processors.request',
+                'tock.context_processors.version_url',
```
tbaxter-18f: Double-check me on this, but I'm not sure this is the optimal order for context processors. If I recall correctly the order of them does matter, and I think you want your somewhere higher. But I may be mis-remembering.

tock/tock/settings/base.py
tbaxter-18f: Why not have align on either `CIRCLE_TAG` or `VERSION`? Why the difference?

tock/tock/templates/base.html
tbaxter-18f: That `div` wrapping footer is probably redundant, since `footer` can fulfill the same role in a more semantic way. If you do need it for some reason, I don't think it's currently closed.

Also, I'd double-check on this, but I think that `title` attribute may be hurting your accessibility.

tock/tock/settings/base.py
ccostino: Good catch @tbaxter-18f, the order definitely matters, but I can't remember off the top of my head either.  [The documentation on them](https://docs.djangoproject.com/en/1.10/topics/templates/#context-processors) (accounting for the version of Django currently running for Tock) is a little light, but I might be looking in the wrong place, too.

Asking in our #django channel in Slack is a good place to follow up on this to see if someone else remembers or at least knows where to look. :-)

ccostino:
Thanks for taking a run through this on our behalf, @tbaxter-18f!  It is greatly appreciated. :-)

tock/tock/context_processors.py
rogeruiz: I'm using the same regular expression in the `.circleci/config.yml` to match versions for tag filters. I wanted to maintain consistency on how the release tags are matched. wdyt? There are surely better regular expressions here, but maybe a comment about how it matches in that other file is appropriate there?

tbaxter-18f: I think that's a perfectly good reason for doing it this way. The startwith() might be a bit simpler and avoid my regexitis, but this will be more consistent as well as more reliable.

tock/tock/settings/base.py
rogeruiz: From what I gathered from the docs, it seems like context processor's order matters in the sense that it's possible for them to overwrite variables found in other context processors. I'm going to keep the order here, but I'm going to export the dictionary with a `x_tock_` prefix to avoid any possible name collisions with internal Django stuff. Thanks!

tock/tock/settings/base.py
rogeruiz: The difference comes from the fact that `$CIRCLE_TAG` comes from the CI. I was thinking of solving this issue in many ways (e.g. using Bash), but I found that it made more sense to just rely on the variables given to me in the environment for the CI we've all standardized on. If the variable doesn't exist, then it just defaults to the `master` branch which is fine if a Git tag isn't being deployed.

In essence, I wanted to avoid a line like this somewhere in the CI configuration. 🤷‍♂️ 

```yaml
    command: |
        export VERSION=${CIRCLE_TAG:master}
```

tock/tock/templates/base.html
rogeruiz: The `title=""` attribute seems to neither be hurting nor helping my accessibility here. Though my age may be showing here, I just like having the hover tooltip when mouseover the link 😄.

Good catch on the `<div>`. I think it's supposed to be a closing tag. Looks like it's been that way for a while 😆.

tbaxter-18f: Here's some info on title attributes and a11y: https://developer.paciellogroup.com/blog/2012/01/html5-accessibility-chops-title-attribute-use-and-abuse/
https://silktide.com/i-thought-title-text-improved-accessibility-i-was-wrong/

I'm not sure if yours is a problem or not. It's probably not a big deal, but I'd suggest a look on those links, and maybe double check in #g-accessibility

tock/tock/settings/base.py
tbaxter-18f: I was thinking more about having CIRCLE_TAG and VERSION being the same thing... two names for the same thing, but I see why it is this way now.

Issue #699, opened 2018-01-22
Update Django to the latest version


annalee:
Per Django's update docs, we'll need to update to 1.11 before updating to 2.0.x.

ccostino:
There is at least one known issue with this update that has prevented it from happening sooner:  The timecard form breaks with the move to 1.11 due to a change in how the initial form data is serialized and sent to the page.  In 1.10, the JSON comes through as expected and we're able to initialize the form properly, but in 1.11, we see the raw JSON being printed on the page and I haven't been able to figure out where the breakdown is in terms of marking something safe or escaping something.

There might be an interaction going on with the [Chosen JS library](https://harvesthq.github.io/chosen/) that also needs to be accounted for, so that could be another piece to check.

toolness:
Ok, I've filed #740, which gets us to 1.11, which is actually a [LTS release that will be supported until at least April 2020](https://www.djangoproject.com/weblog/2015/jun/25/roadmap/).  Do we think that 1.11 is enough for ATO-hood, or is 2.0 required?

ccostino:
Ah that is a good question, @toolness.  @rogeruiz what do you think, should we stick with just the LTS releases for the purpose of ATO activities, or do you think it's okay to go with the latest stable?

rogeruiz:
I think this is good to close for now. We're good for now.

PR #720, opened 2018-02-07
Add in cg-django-uaa support

rogeruiz:
This PR is a work in progress which will allow for Tock to move away from the UAA Guard Service that has been deprecated by :cloud:.gov. This is a new body of work not derived from #640. Although, when authentication is working with UAA via JWT tokens, some of that work will be ported over into this PR.


rogeruiz:
@toolness hey, we're going to be going over this PR and some of the code found in these files tomorrow. Thanks again for your help with this! 👍 💯 🎈 

- https://github.com/18F/tock/blob/auth-you-do-to-me/tock/tock/utils.py#L16-L34
- https://github.com/18F/tock/blob/auth-you-do-to-me/tock/tock/tests/test_remote_user_auth.py#L31-L35
- https://github.com/18F/tock/search?l=Python&q=PermissionMixin&type=&utf8=✓

jmcarp:
Looking forward to this landing! We should be able to tear down the guard service once tock is done with it, since there aren't any other users.

docker-compose.yml
```diff
@@ -12,28 +12,16 @@ app:
...
-    - ./provision/dev/nginx_site_docker.template:/etc/nginx/conf.d/nginx_site_docker.template
-  command: /bin/bash -c "envsubst '$$EXTERNAL_PORT $$DJANGO_SERVER $$USER_EMAIL' < /etc/nginx/conf.d/nginx_site_docker.template > /etc/nginx/conf.d/nginx_site_docker.conf && exec nginx -g 'daemon off;'"
-  environment:
-    - EXTERNAL_PORT=${EXTERNAL_PORT}
-    - DJANGO_SERVER=http://app:1234
-    - USER_EMAIL=aaron.snow@gsa.gov
-  links:
-    - app
-  ports:
-    - "${EXTERNAL_PORT}:${EXTERNAL_PORT}"
```
ccostino: Nice that we don't need these extra pieces anymore!

provision/dev/nginx_site_docker.template
```diff
@@ -1,11 +0,0 @@
...
-  listen ${EXTERNAL_PORT} default_server;
-  server_name _;
-
-  location / {
-    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
-    proxy_set_header X-Forwarded-Host $host:$server_port;
-    proxy_set_header X-Auth-User "${USER_EMAIL}";
-    proxy_pass ${DJANGO_SERVER};
-  }
-}
```
ccostino: Thanks for cleaning this up too!

requirements.txt
```diff
@@ -13,3 +13,4 @@ httmock==1.2.6
 # Cloud.gov specific
 newrelic==2.98.0.81
 cfenv==0.5.3
+cg-django-uaa==1.2.0
```
ccostino: 🎉

tock/employees/tests/test_views.py
```diff
@@ -93,17 +94,19 @@ def test_parse_date(self):
...
-            response.html.find('input', {'id': 'id_end_date'})['value'],
-            '2016-01-01')
+        self.assertContains(
+            response,
+            '<input class="datepicker" id="id_start_date" name="start_date" type="text" value="2014-01-01" />'
+        )
+        self.assertContains(
+            response,
+            '<input class="datepicker" id="id_end_date" name="end_date" type="text" value="2016-01-01" />'
+        )
```
ccostino: In looking at these I think this is okay because as you said, this is going off of the template response now instead of the HTTP response directly.

tock/hours/views.py
```diff
@@ -468,6 +470,7 @@ class AdminBulkTimecardSerializer(serializers.Serializer):
     )
 
 
+@user_passes_test(lambda u: u.is_authenticated, login_url='/')
```
ccostino: Nice!

tock/test_common.py
```diff
@@ -0,0 +1,78 @@
...
+import unittest
+
+from django.test import TestCase, override_settings
+from django.contrib.auth.models import User, Group
+
+from employees.models import UserData
+
+
+@override_settings(
+    PASSWORD_HASHERS=['django.contrib.auth.hashers.MD5PasswordHasher'],
```
ccostino: Out of curiosity, why MD5 and not SHA256 ([`django.contrib.auth.hashers.PBKDF2PasswordHasher`](https://docs.djangoproject.com/en/1.10/topics/auth/passwords/#included-hashers))?

tock/test_common.py
```diff
@@ -0,0 +1,78 @@
...
+        self.assertEqual(response.status_code, 302)
+        self.assertEqual(
+            response['Location'],
+            '/auth/login/?next=%s' % url
+        )
+
+    def test_login_is_required(self):
+        if not self.url:
+            raise unittest.SkipTest()
+        self.assertRedirectsToLogin(self.url)
```
ccostino: 👏 nicely done!  I like how you've abstracted this so that it can be reused throughout the tests.

tock/tock/remote_user_auth.py
```diff
@@ -25,38 +24,32 @@ def email_to_username(email):
...
+    def create_user_with_email(email):
+        username = email_to_username(email)
 
-class EmailHeaderMiddleware(RemoteUserMiddleware):
-    header = 'HTTP_X_AUTH_USER'
+        try:
+            print("Try getting a user that exists already.")
+            user = User.objects.get(username=username)
+        except User.DoesNotExist:
+            print("Create a new user if they don't exist.")
```
ccostino: We'll want to clean up these `print` statements, or turn them into `logging` statements if they would still be useful (and, come to think of it, required for ATO purposes).

tock/tock/templates/uaa_client/login_error.html
```diff
@@ -0,0 +1,14 @@
...
+{% block content %}
+
+<h2>Whoops, login error!</h2>
+
+{% if error_code == "missing_state" or error_code == "missing_session_state" or error_code == "invalid_state" %}
+  <p>Bad session state</p>
+{% elif error_code == "missing_code" %}
+  <p>Missing session code</p>
+{% elif error_code == "authenticate_failed" %}
+  <p>Authetication failed</p>
```
ccostino: `Authentication` - missing an `n`. :-)

tock/tock/utils.py
```diff
@@ -39,9 +47,13 @@ def has_permission(self, request, view):
...
 
+
 def is_running_test_suite():
-    return (os.path.basename(sys.argv[0]) == 'manage.py' and
-            len(sys.argv) > 1 and sys.argv[1] == 'test') or base.FLOAT_API_KEY == ''
+    return (
+        (os.path.basename(sys.argv[0]) == 'manage.py' and
+            len(sys.argv) > 1 and sys.argv[1] == 'test') or
+        base.FLOAT_API_KEY == ''
+    )
```
ccostino: D'oh!  Remnant Float hook; this can be refactored to remove the Float piece(s).

tock/test_common.py
toolness: Oh, this is just for test cases, where we want the tests to run as quickly as possible (rather than be as secure as possible).  I believe PBKDF2 in particular is designed to be resistant to brute force attacks via computational cost, so it's great for production but actually makes tests run _really_ slowly.

I think that really new versions of Django might have introduced a better way to work around the slowness of PBKDF2 during testing, but if they did, I forget what it is...

docker-compose.yml
toolness: Keep in mind that if we're getting rid of `EXTERNAL_PORT`, we should probably get rid of it in [`.env.sample`](https://github.com/18F/tock/blob/master/tock/.env.sample) too, lest it confuse folks!  (Or we could use `EXTERNAL_PORT` in this new configuration, I'm happy to help make that happen.)

tock/employees/tests/test_views.py
toolness: By the way, it might be helpful to pass `html=True` to [`assertContains()`](https://docs.djangoproject.com/en/2.0/topics/testing/tools/#django.test.SimpleTestCase.assertContains) to make the tests less fragile:

> Set `html` to `True` to handle text as HTML. The comparison with the response content will be based on HTML semantics instead of character-by-character equality. Whitespace is ignored in most cases, attribute ordering is not significant.

tock/hours/views.py
toolness: Is there any way for `login_url` to be specified in the Django settings, e.g. via [`settings.LOGIN_URL`](https://docs.djangoproject.com/en/2.0/ref/settings/#login-url)?  It'd help make things a bit more DRY.

Also, is there a reason the [`login_required`](https://docs.djangoproject.com/en/2.0/topics/auth/default/#the-login-required-decorator) decorator can't be used here?  It seems like it'd make the code a bit more readable but I dunno.

tock/tock/utils.py
```diff
@@ -3,14 +3,18 @@
 import os
 import sys
 
-from django.core.exceptions import PermissionDenied
+from django.contrib.auth.mixins import LoginRequiredMixin
 from rest_framework.permissions import BasePermission
+from django.shortcuts import redirect
+
+# from uaa_client.authentication import UaaBackend
```
toolness: Should this commented-out line of code be removed?

tock/tock/settings/test.py
```diff
@@ -6,6 +6,9 @@
 
 SECRET_KEY = get_random_string(50)
 
+# Needed for CG Django UAA to leverage internal cloud.fake service
+DEBUG = True
```
toolness: Hmm, so this is a pretty big deviation from the [standard Django defaults](https://docs.djangoproject.com/en/2.0/topics/testing/overview/#other-test-conditions):

> Regardless of the value of the `DEBUG` setting in your configuration file, all Django tests run with `DEBUG=False`. This is to ensure that the observed output of your code matches what will be seen in a production setting.

Are there particular tests that will fail if `DEBUG` is `False`?  If there's only a handful, it might be useful to use `override_settings` for just the ones that need them... but if it'd make a lot of tests fail, it might be a sign that those tests aren't isolating the functionality to be tested as well as they could, and I could help with that.

tock/tock/utils.py
ccostino: Good catch, most likely unless it'll be needed again. :-)

tock/tock/settings/test.py
ccostino: That's a great question and one I'm not sure either.  Thanks so much for pitching in, @toolness, I'm sure @rogeruiz would appreciate the help!

toolness:
It just occurred to me that we don't have a logout link.  We might want to add one, so that e.g. someone who wants to be super secure can do so, and so that e.g. if two people happen to be at the same computer and need to use tock, they can switch accounts without having to reset their browser cache.

Should this be done in this PR or should I file it as a separate issue?

ccostino:
That's a good idea, @toolness, but I'd file a separate issue for it and work on it either in tandem or after this is complete.

toolness:
Any idea what else needs to be done here for this PR to be merged?  I'd like to work on other tickets for the ATO like #699, but I'm not sure if there's more I can do here first...

rogeruiz:
Thanks @toolness I think we're good for this. I'll review and merge #729 #728 #732 (once it's done) and we should be good to merge this in and continue working on other tickets. If you get started before I work on any of this, please be ready to rebase your work or work from the tip of whatever branch has the latest work that's gonna land on `master`.

toolness:
Just to keep track, there are some issues that should be filed once this lands:

* Do something about [`EXTERNAL_PORT`](https://github.com/18F/tock/pull/720#discussion_r171552848).

* Try to get rid of [`DEBUG = True`](https://github.com/18F/tock/pull/720#discussion_r171556911) in the test settings.

* Fix the `DoesNotExist: ReportingPeriod matching query does not exist` error raised by `test_reports:ReportingPeriodDetailView` (discovered in #729).

* Fix the `IndexError: list index out of range` at `/utilization/` (discovered in #729).

PR #723, opened 2018-02-20
Update Node dependencies

rogeruiz:
## Proposed Acceptance Criteria

- [x] Update node-sass to latest release & all patches.
- [x] Update Node dependencies to be `devDependencies` as these never run in ~~the server~~ :cloud:.gov servers.
- [x] Any documentation relating to the inability to update `hoek` beyond version `2.16.3` until `node-sass` gets released to 5.x.
    - https://github.com/sass/node-sass/pull/2170#issuecomment-347845532

Addresses https://gemnasium.com/github.com/18F/tock/alerts

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+package-lock.json binary
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   "dependencies": {
     "chosen-js": "1.8.2",
     "jquery": "3.2.1",
-    "node-sass": "4.5.3",
+    "node-sass": "4.7.x",
     "underscore": "1.8.3",
     "uswds": "^1.2.0"
   }
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   "engines": {
     "node": "8.6.0"
   },
-  "dependencies": {
+  "devDependencies": {
     "chosen-js": "1.8.2",
     "jquery": "3.2.1",
     "node-sass": "4.7.x",
--- a/package.json
+++ b/package.json
@@ -8,9 +8,9 @@
     "predeploy": "npm install --silent && npm run build-css"
   },
   "engines": {
-    "node": "8.6.0"
+    "node": ">=8.6.0"
   },
-  "devDependencies": {
+  "dependencies": {
     "chosen-js": "1.8.2",
     "jquery": "3.2.1",
     "node-sass": "4.7.x",
--- a/.cfignore
+++ b/.cfignore
@@ -68,3 +68,8 @@ local_settings.py
 .env
 docker-compose.yml
 Dockerfile
+
+# Node related dependencies
+node_modules/
+package.json
+package-lock.json
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,14 +5,12 @@
     "@types/node": {
       "version": "8.0.24",
       "resolved": "https://registry.npmjs.org/@types/node/-/node-8.0.24.tgz",
-      "integrity": "sha512-c3Npme+2JGqxW8+B+aXdN5SPIlCf1C8WxQC6Ea39rO/ASPosnMkWVR16mDJtRE+2dr2xwOQ7DiLxb+wO/TWuPg==",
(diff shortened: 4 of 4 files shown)


rogeruiz:
@afeld, could you take a look at the last bullet here? Do we need to update anything in our documentation that states why we're leaving this vulnerability alone for now? Perhaps something along the lines that we need to wait for the package to get updated and that we're also only using this as a Dev Dependency in local development and CI asset building?

.gitattributes
```diff
@@ -0,0 +1 @@
+package-lock.json binary
```
ccostino: What is this line for?  I've not come across this before, curious to learn!

package.json
```diff
@@ -10,10 +10,10 @@
   "engines": {
     "node": "8.6.0"
   },
-  "dependencies": {
+  "devDependencies": {
```
ccostino: `chosen`, `jquery`, and presumably `underscore` and `uswds` aren't just `devDependencies` - should we move just `node-sass` to `devDependencies` instead?

package.json
```diff
@@ -10,10 +10,10 @@
   "engines": {
     "node": "8.6.0"
```
ccostino: While we're in here should we take this whole piece out, too?  I think we had talked about it before and if anything we should make sure we're running/using the latest Node.js LTS version (8.9.4).

package.json
rogeruiz: After reading the documentation on `npm` [here](https://docs.npmjs.com/files/package.json#devdependencies), I think I'll make these `dependencies` again. But follow up on using `.cfignore` to not upload these files to the server. :cloud:.gov should not be leveraging any `npm`-related code to run Tock. All the assets should be copied over to `static` at integration-time.

.gitattributes
rogeruiz: > `package-lock.json` is **automatically generated** for any operations where npm modifies either the `node_modules` tree, or `package.json.` It describes the exact tree that was generated, such that subsequent installs are able to generate identical trees, regardless of intermediate dependency updates.

The __emphasis is mine__ above. Because it's automatically generated, it shouldn't show up in `git add --patch` or `git diff` commands since it's essentially not to be edited by humans. [This `.gitattribute` file adds a flag of `binary` to any pathname matching the first part](https://www.git-scm.com/docs/gitattributes). It's useful for other things like saying that a `.doc[x]` file is a text file or a particular path contains different line endings. _ 🙄 at you AppleScript_

package.json
rogeruiz: You're right. Thanks!

rogeruiz:
This PR will ultimately be the documentation for why we can't update `hoek`. I'll add something to the project's documentation about how this was determined as part of the ATO process.

rogeruiz:
For posterity, here's the new dependency list after this PR lands.

```bash
$ npm ls hoek && npm ls tough-cookie
~/Developer/eighteenf/tock
└─┬ node-sass@4.7.2
  └─┬ request@2.79.0
    └─┬ hawk@3.1.3
      ├─┬ boom@2.10.1
      │ └── hoek@2.16.3  deduped
      ├── hoek@2.16.3
      └─┬ sntp@1.0.9
        └── hoek@2.16.3  deduped

~/Developer/eighteenf/tock
└─┬ node-sass@4.7.2
  └─┬ request@2.79.0
    └── tough-cookie@2.3.3
```

afeld:
If hoek is only used for SASS compilation, very likely that it's not a (runtime) vulnerability, and therefore is nothing to worry about. As long as you document that fact _somewhere_ (I'd argue doing so in this pull request or in Gemnasium is fine), you should be good. Thanks for the quick turnaround!

ccostino:
Thanks, @afeld!  Going to take care of Gemnasium given this. :-)

ccostino:
Ah just to clarify - Acknowledging the alert in Gemnasium just stops notifications, but closing it is for when you've fixed it or you're not impacted by it.  Closing both of these alerts for Tock given this update.

PR #724, opened 2018-02-21
Add static analysis tools

cmc333333:
Add bandit, flake8 to CI and resolve the "F" flake8 errors (which are most likely to cause bugs). This should catch more errors before they get to production. In an ideal world, we wouldn't be excluding most of these flake8 rules, but it'll take a good chunk of time to update the codebase to match.

Easiest to review changeset by changeset.


rogeruiz:
@cmc333333 what's the best way to fix the issues in the `manage.py` that you recommend in 8daaf6cd6252c6ee77f7a392dab3f4be314df9c6? It looks to me that it's complaining about `shell=True` the most, but the string interpolation is coming from python's `os` library right? so is the severity as high as the error message states?

<details>
<summary>Excerpt from <code>manage.py</code> regarding <a href="https://docs.openstack.org/bandit/1.4.0/plugins/subprocess_popen_with_shell_equals_true.html">[B602:subprocess_popen_with_shell_equals_true]</a></summary>

```py
    def get_children():
        output = subprocess.check_output(
            "ps --ppid=%d -o pid | awk 'NR>1' | xargs echo" % os.getpid(),
            shell=True
        )
        return map(int, output.split())
```

</details>

I'd like to not have to ignore a file that's running in ☁️.gov, but I don't see away around with without refactoring how it works or ignoring spawning subshells.

ccostino:
@rogeruiz Oh!  We can use the [`shlex`](https://docs.python.org/3/library/shlex.html) module in Python to account for this and quote the Shell commands:  `shlex.quote()`

Wrapping them with this should address the issue if I'm understanding this correctly, we had to do the same in FEC.

cmc333333:
Thanks for pointing out `shlex`, @ccostino! I think that could resolve a good chunk of these warnings. However, @rogeruiz is correct that the `shell=True` is the bigger concern due to its use of the caller's env variables (and I don't think `shlex` would resolve that). We could definitely put in some work to remove `shell=True` by ensuring the necessary env vars are explicitly defined, but I'd also consider removing that logic (if it's not in use) altogether. Allowing an auto-use of Docker is neat, but it's in the prod code, which is inherently risky. If no one is using it, there's no harm in removing.

Should that decision and how to resolve the warnings (and removing exceptions from flake8) be addressed after this PR? I think this (partial) linting is an improvement regardless of tact we take there.

cmc333333:
@rogeruiz exactly, that awareness has a cost. I don't have a good sense of how often it's used (i.e. it's benefit), so I can't give a good analysis. It's plausible to me that it's not particularly helpful now, in which case, we should just remove it rather than working around it.

rogeruiz:
I believe it's only value is this. if that's the case, then I'd like to remove it and also remove the docs from the readme regarding it's use. @toolness can you weigh in here? A `git blame` reveals you added this functionality in 761ca752

<details>
<summary>From <code>README.md</code></summary>
<blockquote>Note that if you don't have Django installed on your host system, you can just run python manage.py directly from outside the container--the manage.py script has been modified to run itself in a Docker container if it detects that Django isn't installed.</blockquote>
</details>

ccostino:
@rogeruiz that is correct (@toolness is out today so I'm answering on his behalf); this was added as a convenience and not having to worry if you were in the Docker container or not.  It seems like this is more trouble than it's worth and since the project is setup to be used and run with Docker, well, just run it in Docker. :-)

ccostino:
Thanks again, @cmc333333!

PR #725, opened 2018-03-01
Fix remote_user_auth.py and its tests.

toolness:
**Note that this is a PR against #720, not the master branch.**

This fixes `remote_user_auth.py` and its tests.

### DIFF
--- a/tock/tock/remote_user_auth.py
+++ b/tock/tock/remote_user_auth.py
@@ -39,17 +39,23 @@ def verify_userdata(user):
 
 
 class TockUserBackend(UaaBackend):
+    @classmethod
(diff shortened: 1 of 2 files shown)


tock/tock/remote_user_auth.py
```diff
@@ -39,17 +39,23 @@ def verify_userdata(user):
 
 
 class TockUserBackend(UaaBackend):
+    @classmethod
+    def get_user_by_email(cls, email):
+        user = super().get_user_by_email(email)
+        if user is not None:
+            verify_userdata(user)
+        return user
```
rogeruiz: 👏

PR #726, opened 2018-03-02
Fix non-API hours and utilization view tests.

toolness:
**Note: This is a PR against #720, not `master`.**

This fixes the hours and utilization view tests that aren't API-related.

I wasn't sure what to do about the views that seem to behave like API views: in the legacy version of the app, these views seem to have been protected by token-based auth, but now they have `@login_required` decorators around them, which suggests that they have cookie-based auth, and I'm not sure if that is actually what we want.  If it _is_ what we want, then I can fix these tests to work properly, but if it's _not_ what we want, I guess we should remove the `@login_required` decorators from the relevant views and have their auth managed by Django Rest Framework.  Um, I think.

### DIFF
--- a/tock/hours/tests/test_views.py
+++ b/tock/hours/tests/test_views.py
@@ -48,14 +48,14 @@ class DashboardReportsListTests(WebTest):
     def test_response_ok(self):
         response = self.app.get(
             reverse('reports:DashboardReportsList'),
-            headers={'X_AUTH_USER': self.user.email},
+            user=self.user,
         )
         self.assertEqual(response.status_code, 200)
 
     def test_template_render(self):
         response = self.app.get(
             reverse('reports:DashboardReportsList'),
-            headers={'X_AUTH_USER': self.user.email},
+            user=self.user,
         )
         self.assertEqual(response.status_code, 200)
         self.assertEqual(
@@ -151,7 +151,7 @@ class DashboardViewTests(WebTest):
                 'reports:DashboardView',
                 kwargs={'reporting_period':'1999-12-31'}
             ),
-            headers={'X_AUTH_USER': self.user.email},
+            user=self.user,
         )
         self.assertEqual(response.status_code, 200)
 
@@ -162,7 +162,7 @@ class DashboardViewTests(WebTest):
                 kwargs={'reporting_period':'1999-12-31'}
             ),
             {'unit':'1'},
-            headers={'X_AUTH_USER': self.user.email},
+            user=self.user,
         )
         self.assertEqual(response.status_code, 200)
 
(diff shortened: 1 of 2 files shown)


rogeruiz:
Hmm. Good catch @toolness, so the reason they have decorators is so that they have authentication. I guess, should we have something like the PermissinsMixin or functional views?

Previously I believe these views has no authentication in them. See #640.

toolness:
Hmm, I guess one question I have is, where are these API-like requests being accessed from?  Are they intended for use by API clients that developers write, or are they intended more as endpoints that are accessed via standard Django views (e.g. "click here to download this info as CSV")?  If it's the former then I am guessing we want them to go through DRF auth, whereas if it's the latter, it's probably OK to just keep the `@login_required` decorator and fix the tests.

rogeruiz:
@toolness, I believe it’s a mixture of both. Cc: @ccostino @annalee

toolness:
Oh ok.  If it's used for both, I recommend routing these API views through DRF, and having DRF use multiple auth backends--one that's cookie-based and another that's token-based, so that both the website and API clients can easily use them.

If that's too much work for #720 though, we can always just make things work with `@login_required` for now and leave the DRF-ification for later.

ccostino:
Yeah it is both, exactly as @rogeruiz described, and we do have folks accessing the API directly.  If we can get away with just doing one for now without disrupting things and then moving on to the other piece later, then let's do that to help move things forward that are in flight now. :-)

ccostino:
D'oh sorry for confusion, @rogeruiz is in a better spot to know what makes sense to do where!  I was just trying to keep things on the simpler side to help move them forward. :-)

toolness:
Ok cool--so I just committed 4a527637d4377dca8083f9abb029d13aa7d2510b, which makes the tests pass, and should also allow e.g. use cases like "download this as CSV" work,  but depending on how the code of folks who are accessing the API directly works, it may break that.  I guess if the latter isn't a huge deal in the short-term, I recommend just merging this but filing a separate issue to deal with things in a more proper way later.

toolness:
Additional note: I think auth for folks using the API has *always* been really challenging, because the previous auth proxy solution didn't discriminate (afaik) between incoming browser requests vs. incoming API requests. This meant that even API requests had to be authenticated by that funky `X_AUTH_USER` header. So even though the API might have originally been _designed_ to be token-based, it was complicated by the fact that it was behind that proxy, which is probably what led to the super weird situation we're in now.

toolness:
Oh funky.  Fixing this might be as simple as decorating those API-like views with [DRF's `api_view` decorator](http://www.django-rest-framework.org/api-guide/views/#api_view) instead of `login_required`.  This is what our buddy over at `api.views.hours_by_quarter` does and it seems to work fine.

PR #727, opened 2018-03-02
Convert hours API endpoints to be DRF API views.

toolness:
**Note: This is a PR against #720, not `master`.**

This follows up on #726 to make the API views of `hours` (which deliver CSV content rather than HTML) use [DRF's `api_view` decorator](http://www.django-rest-framework.org/api-guide/views/#api_view) instead of `login_required`.

Crucially, they also modify the app's DRF settings to use [`SessionAuthentication`](http://www.django-rest-framework.org/api-guide/authentication/#sessionauthentication), in addition to the existing [`TokenAuthentication`](http://www.django-rest-framework.org/api-guide/authentication/#tokenauthentication), which makes it possible for the endpoints to be accessed by both programmatic API clients (via `TokenAuthentication`) _and_ website visitors (via `SessionAuthentication`).

### DIFF
--- a/tock/hours/tests/test_views.py
+++ b/tock/hours/tests/test_views.py
@@ -9,6 +9,7 @@ from django.contrib.auth import get_user_model
 from django.contrib.auth.models import User
 from django_webtest import WebTest
 
+from api.tests import client
 from api.views import UserDataSerializer, ProjectSerializer
 from employees.models import UserData
 from hours.utils import number_of_hours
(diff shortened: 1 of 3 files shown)


tock/hours/tests/test_views.py
```diff
@@ -539,7 +528,8 @@ def setUp(self):
         )
 
     def test_user_reporting_period_report(self):
-        response = client(self).get(reverse(
```
toolness: I'm not sure why this test was ever using `client`, as it's not testing an API view.  Honestly, I really think `client` should be imported as the name `api_client`--otherwise its use is quite ambiguous, as the containing file tests both standard website views _and_ API views.

rogeruiz: Thanks @toolness. Do you have any opinions on how the app would be rearchitected? Considering we've got you until the 19th on Tock, if you could jump in and make this easier to maintain that would be great. From an ATO perspective, I think it's fine to leave this all as-is, as it's not terrible to maintain right now. But I think I could use some help in the documentation for onboarding new devs. Specifically, explaining the current architecture and how it's a mix of different Django practices.

toolness: Oh yeah sure, I'm happy to help with that!

Issue #728, opened 2018-03-02
Add a test suite that ensures all views are protected by authentication


toolness:
Since #720 is taking Tock from being an app that didn't have to care at all about authentication (since it was behind a proxy that dealt with it) to one that has to care about it all the time, I think it might be useful to have a test suite that goes through the app's URL config and pokes at every URL (except perhaps a very small whitelist of public pages) and ensures that they return either a 403 or redirect to login.

It might be redundant but I think it could also offer some peace of mind.

@rogeruiz what do you think of this?  If it sounds helpful I'm happy to submit a PR.

rogeruiz:
Perfect! I love this idea. and something that would be good to have with the ATO, considering that we now have an authorization-layer we manage ourselves.

The only public page I can think of is the `uaa_client/login_error.html` everything else should be a redirect to UAA or a 403 to that page. Tock doesn't currently have any public-facing pages in accordance to the SSP.

toolness:
Oh, this was fixed by #729!

PR #729, opened 2018-03-03
Add test suite to ensure most URLs are protected by auth.

toolness:
**Note: This PR is against #720, not `master`.**

This adds a test suite that ensures that most URLs are protected by auth, fixing #728.

To run it, use:

```
python manage.py test tock.tests.test_url_auth
```

## Notes

* This test suite finds URLs by digging into Django internals a bit, specifically [`django.core.urlresolvers`](https://docs.djangoproject.com/en/1.9/_modules/django/core/urlresolvers/).  It's likely that this will break when upgrading to new versions of Django, which is probably OK; this test  suite doesn't need to last forever, but it's particularly important for #720.

## To do

- [x] ~~Ensure the test suite is actually run by CI.  It's located in a `tests` directory in the _project_ directory, not an app directory, which makes me worry that Django's `manage.py test` command might not automatically find it.~~ Verified this new test suite is run during CI by inspecting the CircleCI log for this PR.

- [x] Fix auth holes (i.e., failing tests) uncovered by this new suite:

  - [x] `test_reports:ReportingPeriodDetailView`

    The view at `/reports/2018-01-01/` doesn't have a `PermissionMixin`, nor does it have the `permission_classes` property like a lot of other `ListView` subclasses do.

    Aside from that, this view is raising a `DoesNotExist: ReportingPeriod matching query does not exist.`, which suggests that, even if the current user _is_ authenticated, we should be catching this and returning a 404, instead of propagating the exception. (File a separate bug for this.)

  - [x] `test_reports:ReportingPeriodCSVView`

    Looks like `/reports/2018-01-01.csv/` is returning 200.

  - [x] `test_ProjectListView`

    Looks like `/projects/` is returning 200.

  - [x] `test_ProjectView`

    Looks like `/projects/1/` is returning 404, but if we want it protected by auth, we should probably be redirecting to login instead, even on views that return 404.  (Otherwise, I guess we should actually make a project with id 1 before running this suite.)

  - [x] `test_utilization:GroupUtilizationView`

    Looks like `/utilization/` is not only exposed to unauthenticated users, but it's also raising an `IndexError: list index out of range` exception.  We should fix the former and file an issue to fix the latter later.

  - [X] `test_admin:login`

    So `/admin/login` seems to be exposed, which is not good because it has a password field in it--yet we're not using password-based auth!  We can fix by ensuring this view always redirects to UAA Auth--see [how we fix it in CALC](https://github.com/18F/calc/blob/develop/hourglass/urls.py#L12-L16) for details.

### DIFF
--- a/tock/hours/views.py
+++ b/tock/hours/views.py
@@ -925,7 +925,7 @@ class ReportsList(PermissionMixin, ListView):
         return sorted_fiscal_years
 
 
-class ReportingPeriodDetailView(ListView):
+class ReportingPeriodDetailView(PermissionMixin, ListView):
     template_name = 'hours/reporting_period_detail.html'
     context_object_name = 'timecard_list'
 
@@ -961,6 +961,7 @@ class ReportingPeriodDetailView(ListView):
         return context
 
 
+@api_view(['GET'])
 def ReportingPeriodCSVView(request, reporting_period):
     """Export a CSV of a specific reporting period"""
     response = HttpResponse(content_type='text/csv')
--- a/tock/tock/urls.py
+++ b/tock/tock/urls.py
@@ -1,9 +1,21 @@
 from django.conf import settings
 from django.conf.urls import include, url
+from django.contrib.auth.decorators import user_passes_test
+from django.core.exceptions import PermissionDenied
 
 # Enable the Django admin.
 from django.contrib import admin
 admin.autodiscover()
+
+def check_if_staff(user):
+    if not user.is_authenticated():
+        return False
+    if user.is_staff:
+        return True
+    raise PermissionDenied
+
+staff_login_required = user_passes_test(check_if_staff)
+admin.site.login = staff_login_required(admin.site.login)
 
 import hours.views
 import api.urls
--- a/tock/utilization/views.py
+++ b/tock/utilization/views.py
@@ -6,9 +6,10 @@ from django.core.urlresolvers import reverse
 from hours.models import TimecardObject, ReportingPeriod
 from employees.models import UserData
 
+from tock.utils import PermissionMixin
 from .utils import get_dates, calculate_utilization
 
-class GroupUtilizationView(ListView):
+class GroupUtilizationView(PermissionMixin, ListView):
     template_name = 'utilization/group_utilization.html'
 
     def get_queryset(self):
--- a/tock/hours/tests/test_views.py
+++ b/tock/hours/tests/test_views.py
@@ -1004,7 +1004,8 @@ class ReportTests(WebTest):
             reverse(
                 'reports:ReportingPeriodCSVView',
                 kwargs={'reporting_period': '2015-01-01'},
-            )
+            ),
+            user=self.regular_user
         )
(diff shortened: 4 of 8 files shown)


ccostino:
@toolness, really great work with this and catching these!  Based on what I'm seeing, I'd err on the side that these should all be protected (and your notes about the redirects instead of error propagation are spot on, too).

Great catch with the admin login page, too!

toolness:
Hooray!  @rogeruiz I think this is good to merge, all the auth holes have been plugged and the tests pass.

tock/hours/views.py
```diff
@@ -925,7 +925,7 @@ def get_queryset(self, queryset=None):
         return sorted_fiscal_years
 
 
-class ReportingPeriodDetailView(ListView):
+class ReportingPeriodDetailView(PermissionMixin, ListView):
```
tbaxter-18f: Atul, this may be a terrible idea, but I wonder if it makes more sense to create an AuthListView that subclasses ListView (with the PermissionMixin), and we use that everywhere. Might be no better -- either way you'd have to remember to do _something_. But it's a thought.

tock/tock/urls.py
```diff
@@ -1,10 +1,22 @@
...
+from django.core.exceptions import PermissionDenied
 
 # Enable the Django admin.
 from django.contrib import admin
 admin.autodiscover()
 
+def check_if_staff(user):
+    if not user.is_authenticated():
+        return False
+    if user.is_staff:
```
tbaxter-18f: If you wanted, you could just say is user.is_authenticated() and user.is_staff. But I'm not sure you need user.is_authenticated() at all, since I don't think user.is_staff can be true unless they're authenticated and it can check.

tock/tock/urls.py
```diff
@@ -1,10 +1,22 @@
...
+        return True
+    raise PermissionDenied
+
+staff_login_required = user_passes_test(check_if_staff)
+admin.site.login = staff_login_required(admin.site.login)
+
 import hours.views
 import api.urls
 import projects.urls
```
tbaxter-18f: Minor, but the imports should be above the new stuff you added.

tock/utilization/views.py
```diff
@@ -6,9 +6,10 @@
 from hours.models import TimecardObject, ReportingPeriod
 from employees.models import UserData
 
+from tock.utils import PermissionMixin
```
tbaxter-18f: Is tock.utils not .utils?

tock/hours/views.py
rogeruiz: Seems like more indirection to me. But I do also thing that Tock is in a place to have an architectural discussion for both its current architecture and future iterations. 

@tbaxter-18f do you have hours to contribute to it? I could certainly use the help.

tbaxter-18f: I can find some time. I'm definitely not sure globally subclassing ListView would be any real benefit. Just something I was thinking about.

rogeruiz:
:shipit:

PR #730, opened 2018-03-05
Redirect /admin/login to UAA, return 403 if not staff.

toolness:
**Note: This is a PR against #729, not `master`.**

This fixes login in a way similar to what we did for CALC.  It contains a simplified version of the [`staff_login_required`](https://github.com/18F/calc/blob/develop/hourglass/decorators.py) decorator that we used in CALC.

## To do

- [ ] Instead of merging this PR, we might want to consider actually adding CALC's `staff_login_required` decorator to [cg-django-uaa](https://github.com/18F/cg-django-uaa) and using it from there.  It's likely that any Django project that uses both cg-django-uaa and the admin UI will want to wrap the admin login view with a decorator like this. **Update:** Filed https://github.com/18F/cg-django-uaa/issues/43 to address this.
- [x] If we decide to go with this solution, add tests.

Thoughts @rogeruiz ?

### DIFF
--- a/tock/tock/tests/test_admin.py
+++ b/tock/tock/tests/test_admin.py
@@ -1,0 +1,26 @@
+from django.test import TestCase
+from django.contrib.auth.models import User
+
+
+class AdminLoginTests(TestCase):
+    url = '/admin/login/'
+
+    def test_anonymous_users_are_redirected_to_uaa_login(self):
+        response = self.client.get(self.url)
+        self.assertRedirects(
+            response, '/auth/login?next=/admin/login/',
+            fetch_redirect_response=False)
+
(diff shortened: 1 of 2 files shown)


toolness:
Nope, it doesn't--it just shows Django's default, rather bland 403 page:

bland-403

The correct way to show something more user-friendly is to configure Django's 403 page to show a different template, which I believe is what CALC does.  However, for the short-term, especially since I assume not many non-admins will be given admin-only links, I assumed it was OK to just give them this bland page.

rogeruiz:
Hmm, this makes me want to do this right then. I've sprinkled a few renders to that template rather than overriding the base 403 page. PR incoming. Thanks @toolness

PR #732, opened 2018-03-05
Add error views instead of reusing login_error.html

rogeruiz:
**This branch is to be merged into #720 and not `master`.**

[This work refactors the error states](https://github.com/18F/tock/pull/730#issuecomment-370475323) and leaves the any files touched linted into a better state then how they were found.

### Acceptance Criteria

- [x] Add `404`, `403`, `500` error pages
- [x] Add CSRF redirect error page
- [x] Refactor renders to `login_error.html` from various places by raising errors instead

### DIFF
--- a/tock/hours/views.py
+++ b/tock/hours/views.py
@@ -11,10 +11,10 @@ from django.contrib.auth import get_user_model
 from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.core.urlresolvers import reverse
 from django.http import Http404, HttpResponse
-from django.shortcuts import render
 from django.views.generic import ListView, DetailView, TemplateView
 from django.views.generic.edit import CreateView, UpdateView, FormView
 from django.db.models import Sum
+from django.shortcuts import render_to_response
 
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.decorators import api_view
@@ -586,17 +586,11 @@ def user_timeline_view(request):
         timecard__user__user_data__organization__name='organization'
     )
 
+
 @api_view(['GET'])
 def admin_bulk_timecard_list(request):
     if not request.user.is_superuser:
-        return render(
-            request,
-            'uaa_client/login_error.html',
-            context={
-                'superuser': request.user.is_superuser,
-            },
-            status=403
-        )
+        return render_to_response('403.html')
 
     queryset = get_timecards(TimecardList.queryset, request.GET)
     serializer = AdminBulkTimecardSerializer()
--- a/tock/hours/views.py
+++ b/tock/hours/views.py
@@ -189,9 +189,9 @@ class DashboardView(PermissionMixin, TemplateView):
         rev_fytd = hours_billed_fytd * target.labor_rate
 
         # Get data for reporting period and clean result.
+        rp_select_time = rp_selected.start_date.strftime('%Y-%m-%d')
         hours_billed_weekly = clean_result(TimecardObject.objects.filter(
-            timecard__reporting_period__start_date=\
-            rp_selected.start_date.strftime('%Y-%m-%d'),
+            timecard__reporting_period__start_date=rp_select_time,
             project__accounting_code__billable=True,
             timecard__user__user_data__in=employees
         ).exclude(
@@ -481,6 +481,7 @@ def user_data_csv(request):
     serializer = UserDataSerializer()
     return stream_csv(queryset, serializer)
 
+
 @api_view(['GET'])
 def projects_csv(request):
     """
@@ -489,6 +490,7 @@ def projects_csv(request):
     queryset = Project.objects.all()
     serializer = ProjectSerializer()
     return stream_csv(queryset, serializer)
+
 
 @api_view(['GET'])
 def bulk_timecard_list(request):
@@ -499,6 +501,7 @@ def bulk_timecard_list(request):
     serializer = BulkTimecardSerializer()
     return stream_csv(queryset, serializer)
 
+
 @api_view(['GET'])
 def slim_bulk_timecard_list(request):
     """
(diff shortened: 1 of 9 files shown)


tock/hours/views.py
```diff
@@ -586,17 +593,11 @@ def user_timeline_view(request):
...
     if not request.user.is_superuser:
-        return render(
-            request,
-            'uaa_client/login_error.html',
-            context={
-                'superuser': request.user.is_superuser,
-            },
-            status=403
-        )
+        return render_to_response('403.html')
```
toolness: Is there a reason we're not just raising `PermissionDenied` here, and allowing Django (or DRF, I am not sure which) to take care of the actual rendering?

An interesting aspect of these API endpoints in particular: raising `PermissionDenied` might actually be especially effective, because my hunch is that DRF will render a different response based on whether a programmatic client or a web browser is on the other end, because it has a whole funky [content negotiation](http://www.django-rest-framework.org/api-guide/content-negotiation/) strategy.

tock/tock/settings/base.py
```diff
@@ -23,6 +23,8 @@
 LOGIN_URL = '/auth/login'
 LOGIN_REDIRECT_URL = '/'
 
+CSRF_FAILURE_VIEW = 'tock.views.csrf_failure'
```
toolness: Whoa, I had no idea [`CSRF_FAILURE_VIEW`](https://docs.djangoproject.com/en/2.0/ref/settings/#csrf-failure-view) was even a setting! Thanks for teaching me something here, and for making everything more Django Bango

tock/tock/templates/uaa_client/login_error.html
```diff
@@ -1,14 +1,7 @@
...
 
-{% if not superuser %}
-<h2>Whoops, you don't have access</h2>
-
-<p>You don't have access to view this page.<p>
-{% endif %}
-
-{% if error_code %}
-<h2>Whoops, login error!</h2>
+<h2>UAA Login Error</h2>
```
toolness: Do we really want to use the acronym "UAA" here, since this will be shown to users who might not know what it means?  Up to you, really, I just thought I'd point it out.

tock/tock/views.py
```diff
@@ -0,0 +1,5 @@
+from django.shortcuts import render_to_response
+
+
+def csrf_failure(request, reason=""):
+    return render_to_response('403.html')
```
toolness: I wonder if it'd be useful to include an extra bit of info for devs in the final view?  For instance, if I'm working on Tock and I get Django's default CSRF failure view, I'll understand what I need to do to fix things, but if I get a 403 page, I'll start thinking about permissions, which is actually the wrong road to go down when debugging this problem.

What if our error pages had a "technical details" section that could include technical jargon useful for debugging?  (This could be something that only displays when `DEBUG` is active, but it might also be useful to have in production, in case end-users happen to run into it...)

tock/hours/views.py
rogeruiz: Oh good call, thanks!

tock/hours/views.py
```diff
@@ -11,10 +11,10 @@
 from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.core.urlresolvers import reverse
 from django.http import Http404, HttpResponse
-from django.shortcuts import render
 from django.views.generic import ListView, DetailView, TemplateView
 from django.views.generic.edit import CreateView, UpdateView, FormView
 from django.db.models import Sum
+from django.shortcuts import render_to_response
```
tbaxter-18f: Why the reversion to the older render_to_response rather than the newer render() shortcut?

tock/hours/views.py
```diff
@@ -189,9 +189,9 @@ def get_params(key):
         rev_fytd = hours_billed_fytd * target.labor_rate
 
         # Get data for reporting period and clean result.
+        rp_select_time = rp_selected.start_date.strftime('%Y-%m-%d')
         hours_billed_weekly = clean_result(TimecardObject.objects.filter(
```
tbaxter-18f: Not directly related, but I think this query could be a bit cleaner and easier to read. It just feels a bit nested. I'm not sure, but I also wonder if we might want `only()` rather than `exclude()`

Also, what's the variance in rp_selected.start_date and timecard__reporting_period__start_date that necessitates teh strftime above? Aren't they both date objects?

tock/hours/views.py
```diff
@@ -481,6 +481,7 @@ def user_data_csv(request):
     serializer = UserDataSerializer()
     return stream_csv(queryset, serializer)
 
+
```
tbaxter-18f: Minor, but per PEP-8 "Method definitions inside a class are surrounded by a single blank line."

tock/hours/views.py
```diff
@@ -816,10 +820,11 @@ def last_timecard(self):
...
-            TimecardPrefillData.objects.active().filter(
-                employee=self.request.user.user_data).values_list(
-                    'project', 'hours'
-                )
+            TimecardPrefillData
+            .objects
+            .active()
+            .filter(employee=self.request.user.user_data)
+            .values_list('project', 'hours')
         )
```
tbaxter-18f: I think that was way more readable before, especially given the nested queryset.  I'd probably get the nested list first, then pass it in. Also, maybe the `active()` manager could wrap `objects` and tighten things up too. Maybe.

tock/tock/settings/base.py
tbaxter-18f: +1

tock/tock/templates/403.html
```diff
@@ -0,0 +1,8 @@
+{% extends "base.html" %}
+
+{% block content %}
+
+<h2>Forbidden</h2>
```
tbaxter-18f: This may just be me, but un-indented templates drive me nuts :-)

tock/tock/templates/uaa_client/login_error.html
tbaxter-18f: I agree with @toolness. I'm an anti-acronym-ite.

tock/tock/views.py
tbaxter-18f: This may just be my general dislike of render_to_response, but I think we should use render, if only because I think at some point we may get tripped up wondering why expected context ain't in the template.

rogeruiz:
Thanks for the comment @toolness. I'll merge this in once tests pass.

PR #737, opened 2018-03-06
Add logout view and menu item

### LINKED ISSUE #731: Add a logout link
As mentioned in https://github.com/18F/tock/pull/720#issuecomment-370448610:

> It just occurred to me that we don't have a logout link.  We might want to add one, so that e.g. someone who wants to be super secure can do so, and so that e.g. if two people happen to be at the same computer and need to use tock, they can switch accounts without having to reset their browser cache.

toolness:
Fixes #731.

### DIFF
--- a/tock/tock/templates/_navigation.html
+++ b/tock/tock/templates/_navigation.html
@@ -24,6 +24,11 @@
           <span>Users</span>
         </a>
       </li>
+      <li>
+        <a href="{% url 'logout' %}" class="usa-nav-link">
+          <span>Log out</span>
+        </a>
+      </li>
       {% if request.user.is_superuser %}
       <li>
         <button class="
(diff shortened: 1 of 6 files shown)


tock/tock/views.py
```diff
@@ -14,3 +15,8 @@ def csrf_failure(request, reason=""):
         )
     )
     return render(request, '403.html')
+
+
+def logout(request):
```
toolness: I decided to add the view to the project, rather than an individual app, because logging out doesn't seem specific to any app in particular.

tock/tock/templates/_navigation.html
```diff
@@ -24,6 +24,11 @@
           <span>Users</span>
         </a>
       </li>
+      <li>
+        <a href="{% url 'logout' %}" class="usa-nav-link">
+          <span>Log out</span>
+        </a>
```
toolness: Normally we'd guard this with a check to see if the current user is actually logged in, but since that's guaranteed (except for the `logout` view and error views, which this partial isn't included in), I didn't do that.

PR #740, opened 2018-03-06
Upgrade to Django 1 11

toolness:
This moves us closer to #699 by upgrading us to Django 1.11.  From there we will be able to upgrade to yet another version of Django, the infamous 2.0!

## Notes

* In order to properly use the changes in this new PR, Docker users will have to run `docker-compose build`, while non-Docker users will need to re-run `pip install -r requirements.txt`.  **To avoid confusion, we should mention this on Slack as soon as this PR is merged.**

* As @ccostino noted in https://github.com/18F/tock/issues/699#issuecomment-359467001, the select widgets in the timecard form are broken. This is because we have a custom subclass of `forms.widgets.Select` called `SelectWithData` which needed to be upgraded to use [Django's  new form rendering API](https://docs.djangoproject.com/en/2.0/ref/forms/renderers/).  Specifically, I had to override the superclass' [`create_option` method](https://github.com/django/django/blob/master/django/forms/widgets.py#L606-L624) to add the custom data attributes that we expect (among other things).

    The good news is that because of the new form rendering API's separation between widget data and template rendering, the new subclass is more concise than the old one.

## To do

- [x] Make the tests pass again.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,7 +1,7 @@
 bandit==1.4.0
 coverage==4.4.1
 django-debug-toolbar==1.9.1
-django-webtest==1.7.9
+django-webtest==1.9.2
 factory-boy==2.8.1
 flake8==3.5.0
 nplusone==0.8.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-Django==1.10.8
+Django==1.11.11
 dj-database-url==0.4.2
 psycopg2==2.7.3.2
 model-mommy==1.3.0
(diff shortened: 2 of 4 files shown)


ccostino:
NICE, @toolness!  Thank you so much for digging into this and uncovering what the issue was!  I knew there was some breakdown in the process and just hadn't figured out where yet!

toolness:
Sure thing @ccostino!  I was familiar w/ the new widget API from making [django-uswds-forms](http://django-uswds-forms.readthedocs.io/en/latest/) last year, so once I noticed that we used a custom `Widget` subclass, the cause was quick to identify :)

I think this PR is good to go... all the tests pass, and things seem to work okay on my local dev instance, but as with any Django upgrade, I suspect it'll be a good idea to get folks to kick the tires a bit more.  Or we can hold off on manual testing until we've upgraded to Django 2.0, too.

Speaking of which, I guess I'll start working on upgrading us to 2.0...

toolness:
Woot thanks Tim!

PR #741, opened 2018-03-06
Fixed template error in elif

tbaxter-18f:
## Description
I'm not sure if there's an issue for this or if we want to consider it a patch on the UAA work that was merged this morning, but there's a template error in the elif, which needed to resolve in an `else`


tock/tock/templates/uaa_client/login_error.html
```diff
@@ -1,22 +1,22 @@
...
+      <p>Bad session state</p>
 
-<p>Missing session code</p>
+  {% elif error_code == "missing_code" %}
 
-{% elif error_code == "authenticate_failed" %}
+    <p>Missing session code</p>
 
-<p>Authentication failed</p>
+  {% else error_code == "authenticate_failed" %}
```
toolness: Er wait, I'm confused by this... what is this line actually doing?  I understand just an `{% else %}` but I'm unclear on what `{% else error_code == "authenticate_failed" %}` means...

tbaxter-18f: Well, it's replacing L15 from before, which used an elif rather than terminating in the `else` as required. I would assume `authenticate_failed` is a response coming back from the fake cloud.gov? @rogeruiz can you provide some additional illumination here?

rogeruiz: Yes. Those string values are coming from the UAA client but I’m not sure about the `elif` being changed to an `else` here. From the Python docs it looks like it’s `elif`: https://docs.python.org/3/tutorial/controlflow.html

But we could terminate with an empty `else` expression and have a more generic error message.

tbaxter-18f: Take a look at `https://docs.djangoproject.com/en/2.0/ref/templates/builtins/`

As I'm reading it, you have to end with the `else` to catch any remainder.  That would jibe with the `Invalid block tag on line 12: 'elif', expected 'endblock'` error I'm seeing, in that it's got know it's done somehow.

I see what you're saying about changing it here though, and agree it might be better to terminate in a generic error, perhaps something like "There was a problem authenticating you: {{ error_code }}" which would also give some indication we may have another case to catch.

tbaxter-18f:
Looking at the logic more, I made the final "authentication failed" state the catch-all, since at that point, that's really what's going on: we're mopping up any auth failures we didn't catch already.

tock/tock/templates/uaa_client/login_error.html
```diff
@@ -1,22 +1,20 @@
...
-<h2>cloud.gov Login Error</h2>
+{% block content %}
 
-{% if error_code == "missing_state" or
-      error_code == "missing_session_state" or
-      error_code == "invalid_state" %}
+  <h2>cloud.gov Login Error</h2>
 
-<p>Bad session state</p>
+  {% if "_state" in error_code %}
```
toolness: Huh, where does `_state` come from?

tbaxter-18f: We're substring matching. It's in all the of the state error_codes, so it's a fast way to check, without having to explicitly write them all out like we did before.

toolness: Oh LOL, my bad, my brain was in JavaScript-land.

tock/tock/templates/uaa_client/login_error.html
```diff
@@ -1,22 +1,20 @@
...
+      <p>Bad session state</p>
 
-<p>Missing session code</p>
+  {% elif error_code == "missing_code" %}
 
-{% elif error_code == "authenticate_failed" %}
+    <p>Missing session code</p>
 
-<p>Authentication failed</p>
+  {% else %}
```
rogeruiz: Hey @tbaxter-18f I'm not sure why you're removing the explicit check for `error_code == "authenticate_failed"` here? I like the approach of the substring match 👍 and in general having a catchall like an empty `else` is good in the case the underlying `uaa_client` code changes. 

<details>
<summary>
<a href="https://github.com/18F/cg-django-uaa/blob/maser/uaa_client/views.py#L39-L55">Here's</a> where I'm getting these values from.
</summary>

```py
    if state is None:
        return login_error(request, 'missing_state')


    if expected_state is None:
        return login_error(request, 'missing_session_state')


    if state != expected_state:
        return login_error(request, 'invalid_state')


    if code is None:
        return login_error(request, 'missing_code')


    user = django.contrib.auth.authenticate(uaa_oauth2_code=code,
                                            request=request)


    if user is None:
        return login_error(request, 'authenticate_failed')
```
</details>

tbaxter-18f: I started to, and the more I thought about it, the more I though it was unnecessary bloat. My logic was that we've already captured the expected failure states. if we get to here, it's either a straight-up generic authentication failed, or something really weird we didn't foresee. So either way, we can treat is a generic error, provide the same generic message, and output what the code was for further investigation.  I'm not strongly opposed to having a separate authenticate_failed block, but I don't think it's necessary or even useful.

rogeruiz: hmm yeah. I think that's fair! let's stick to that logic then. Thinking about it more, my concern is with exposing the `error_code` variable to the user. I think it'd be better to just log it rather than expose it. That way we don't give folks who encounter it in the wild any insight into what caused their issue, but people with log access can determine it pretty quickly. Maybe we should just update the [`uaa_client` code here](https://github.com/18F/cg-django-uaa/blob/maser/uaa_client/views.py#L15) to output a generic log message. Or in the [new functions I added](https://github.com/18F/tock/blob/master/tock/tock/views.py#L25-L70) to `tock/views.py` maybe but I don't think the `uaa_client` is hitting those functions.

I can approve this based on this explanation, but what do you think of just sending that `error_code` to logs rather than the HTML view the user sees?

PR #745, opened 2018-03-07
adding a footer

wslack:
## Description

the current footer text sits awkwardly on the left margin; this should(?) fix that?

## Additional information

What it looks like now: 

screen shot 2018-03-07 at 5 17 57 pm

### DIFF
--- a/tock/tock/templates/base.html
+++ b/tock/tock/templates/base.html
@@ -62,15 +62,20 @@ m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
       {% block js %}
       {% endblock js %}
 </div>
-    <footer>
-       <div class="usa-footer-secondary_section">
-        <div class="usa-grid">
-          <div class="usa-footer-logo">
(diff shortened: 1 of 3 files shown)


tock/tock/templates/base.html
```diff
@@ -62,9 +62,15 @@
...
 </div>
-  <footer>
-    <p><i>The current release is <code>{{ x_tock_release_name }}</code>, <a href="{{ x_tock_release_url }}" title="{{ x_tock_release_name }}">view this release on GitHub</a></i>.</p>
-  </footer>
-
+    <footer>
+       <div class="usa-footer-secondary_section">
+        <div class="usa-grid">
+          <div class="usa-footer-logo">
+            <p><i>The current release is <code>{{ x_tock_release_name }}</code>, <a href="{{ x_tock_release_url }}" title="{{ x_tock_release_name }}">view this release on GitHub</a></i>.</p>
```
tbaxter-18f: Do we need three wrappers (four counting the original footer element) just to put some side margin on a paragraph? I get that we're reusing some prior art here, and I appreciate that, but is there a lighter-weight alternative?

wslack: I'm taking from the USWDS. Feel free to edit a wrapper out if that's better!

rogeruiz:
screen shot 2018-03-08 at 4 41 09 pm

Thanks @wslack I've tweaked your work a bit, but it looks great now. Thanks for the nudge 👍

PR #746, opened 2018-03-08
Invalid reporting periods should 404. Also some mild view cleaning.

### LINKED ISSUE #735: Reporting period queries should return 404 when they don't find anything
Concerning the `test_reports:ReportingPeriodDetailView`, this view is raising a `DoesNotExist: ReportingPeriod matching query does not exist.`, which suggests that, even if the current user is authenticated, we should be catching this and returning a `404`, instead of propagating the exception.

_Discovered in #729_

### Acceptance Criteria
- [ ] Fix the DoesNotExist: ReportingPeriod matching query does not exist error raised by test_reports:ReportingPeriodDetailView

tbaxter-18f:
## Description

Resolves #735 to 404 if the given start date does not match a ReportingPeriod. Also did some mild clean-up on the view to (hopefully) improve readability.


tbaxter-18f:
I think there's some disagreement between the logged issue and the tests. Per the issue, we want to 404. The Test is looking for a redirect to login:
```
FAIL: test_reports:ReportingPeriodDetailView (tock.tests.test_url_auth.URLAuthTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/circleci/repo/tock/tock/tests/test_url_auth.py", line 169, in test
    self.assertURLIsProtectedByAuth(url)
  File "/home/circleci/repo/tock/tock/tests/test_url_auth.py", line 154, in assertURLIsProtectedByAuth
    f'GET {url} returned HTTP {code}, but should redirect to login or '
AssertionError: GET /reports/2018-01-01/ returned HTTP 404, but should redirect to login or deny access
```

Do we want to update the test or the view behavior? Or is there something else I'm missing?

tock/hours/views.py
```diff
@@ -935,35 +936,50 @@ class ReportingPeriodDetailView(PermissionMixin, ListView):
     template_name = 'hours/reporting_period_detail.html'
     context_object_name = 'timecard_list'
 
+    def dispatch(self, *args, **kwargs):
```
rogeruiz: Love this! didn't know about the [`Method Flowchart` for `django.views.generic.list.ListView`](https://docs.djangoproject.com/en/1.11/ref/class-based-views/generic-display/#listview).

tock/hours/views.py
```diff
@@ -935,35 +936,50 @@ class ReportingPeriodDetailView(PermissionMixin, ListView):
...
-            Timecard.objects.filter(
-                reporting_period=reporting_period,
+        """
+        We need to add the reporting period to context, 
+        as well as users who have not filed timecards, 
+        so long as they are not new hires or recently departed.
+        
+        To do so, we'll first get a quick list of IDs
+        of those who have filed.
+        """
```
rogeruiz: 💯 for documenting this. Thank you!

rogeruiz:
@tbaxter-18f I think the view should be verifying the authentication of the user. we should redirect to a 404 only if they're authenticated, but if they're not authenticated, it should redirect to login, then a 404. We don't want to expose errors to un-authorized users. So it might require _some_ additional changes to the test too. But otherwise the logic in the view looks good. 👍

tbaxter-18f:
Hmm. that's an interesting point. I would think the `PermissionMixin` would be capturing that, but it's not, at least not at `dispatch` where I'm calling `get_object_or_404`. So I'll need to dive into the test and the permissionMixin and figure out why either the order or the auth is not as expected.

tock/hours/views.py
tbaxter-18f: You probably already know this, but https://ccbv.co.uk/ is also excellent for tracing through the CBV steps

tock/hours/views.py
```diff
@@ -935,35 +936,58 @@ class ReportingPeriodDetailView(PermissionMixin, ListView):
...
+    def dispatch(self, *args, **kwargs):
+        """
+        Because PermissionMixin and this view are both subclassing dispatch
+        We'll explicitly check auth and pass to handle_no_permission().
+        
+        See github.com/django/django/blob/master/django/contrib/auth/mixins.py#L47
+        """
+        if not self.request.user.is_authenticated:
+            return self.handle_no_permission()
+        self.report_start_date = dt.datetime.strptime(
```
tbaxter-18f: If someone knows of a more elegant way of handling ^ I'm all ears.

tock/hours/views.py
```diff
@@ -935,35 +936,58 @@ class ReportingPeriodDetailView(PermissionMixin, ListView):
     template_name = 'hours/reporting_period_detail.html'
     context_object_name = 'timecard_list'
 
+    def dispatch(self, *args, **kwargs):
+        """
+        Because PermissionMixin and this view are both subclassing dispatch
+        We'll explicitly check auth and pass to handle_no_permission().
+
```
tbaxter-18f: If anyone knows of a more elegant way of handling this, I'm all ears.

tock/hours/views.py
tbaxter-18f: I live for docs. They're great.

tbaxter-18f:
This should be cleared up with the latest pushes.

tock/hours/views.py
```diff
@@ -935,35 +936,58 @@ class ReportingPeriodDetailView(PermissionMixin, ListView):
...
-            Timecard.objects.filter(
-                reporting_period=reporting_period,
+        """
+        We need to add the reporting period to context,
+        as well as users who have not filed timecards,
+        so long as they are not new hires or recently departed.
+
+        To do so, we'll first get a quick list of IDs
+        of those who have filed.
+        """
```
ccostino: Thank you for the added context!

tock/hours/views.py
```diff
@@ -935,35 +936,58 @@ class ReportingPeriodDetailView(PermissionMixin, ListView):
     template_name = 'hours/reporting_period_detail.html'
     context_object_name = 'timecard_list'
 
+    def dispatch(self, *args, **kwargs):
+        """
+        Because PermissionMixin and this view are both subclassing dispatch
+        We'll explicitly check auth and pass to handle_no_permission().
+
+        See github.com/django/django/blob/master/django/contrib/auth/mixins.py#L47
+        """
```
ccostino: This is very helpful, thank you!

PR #747, opened 2018-03-08
Modify footer link context processor

rogeruiz:
This fixes both the CircleCI deployment of Production and adds better
context to what version is on both Staging and Production whenever it
gets deployed.

cc: @toolness

### DIFF
--- a/tock/tock/context_processors.py
+++ b/tock/tock/context_processors.py
@@ -15,6 +15,6 @@ def version_url(request):
     if re.match('/v20[1-9][0-9][0-9]+\.[0-9]+/', name):
         response['x_tock_release_url'] = '%sreleases/tag/%s' % (base_url, name)
     elif name != 'master':
-        response['x_tock_release_url'] = name[:7]
+        response['x_tock_release_name'] = name[:7]
 
     return response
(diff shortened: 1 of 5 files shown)


rogeruiz:
Once this is merged in, I'll cut a tagged release to get it to deploy on production. Thanks @toolness! :+1:

tock/tock/context_processors.py
```diff
@@ -13,6 +13,8 @@ def version_url(request):
     # Regular expression here matches the regular expression found in
     # .circleci/config.yml for filtering on tags for Production deployments
     if re.match('/v20[1-9][0-9][0-9]+\.[0-9]+/', name):
-        response['release_url'] = '%sreleases/tag/%s' % (base_url, name)
+        response['x_tock_release_url'] = '%sreleases/tag/%s' % (base_url, name)
+    elif name != 'master':
+        response['x_tock_release_name'] = name[:7]
```
rogeruiz: If the default is always `master` and I only ever want to display 7 characters from the `sha`, it might make sense to not have this check here and just do it on line 10, but then we'd need to set the name again if it matches the version RegEx. _just thinking aloud_ 💭

toolness: Does CloudFront require us to configure this header to be included in outbound responses?  it's been a little while since i've dealt with CF, so i forget if it cares about outbound headers as much as it does for inbound ones.

toolness:
This looks good!  (For reference, we used https://stackoverflow.com/a/24725091 to figure out how to change the error handlers.)

Issue #733, opened 2018-03-06
Developer on-boarding / off-boarding documentation


rogeruiz:
See [here](https://github.com/18F/tock/pull/720#discussion_r171552848) for a particular issue being addressed by these changes. But Tock needs betting on-boarding documentation as part of the ATO process (#712) and 18F/infrastructure#685.

_Discovered in #720_

Acceptance Criteria
- [ ] Documentation touching on AC-2, AC-6
    - [ ] On-boarding documentation and process
    - [ ] Off-boarding documentation and process
- [ ] Clear up any confusing or redundant legacy information or code such as the link referenced above or [this comment](https://github.com/18F/tock/pull/720#discussion_r171411049) as much as possible.

toolness:
Ah, yeah, we need to update the docs here. My advice is that before you start up the app, run:

```
python manage.py createsuperuser --username admin@gsa.gov --email admin@gsa.gov --noinput
```

Then start up the app and when you reach fake cloud.gov, log in as admin@gsa.gov.  You will then be logged in as the superuser.

If you ever want to create a non-staff user, just log into fake cloud.gov with any email at gsa.gov, and a non-staff account will automatically be created for you.

PR #755, opened 2018-03-10
Update Getting Started w/ valid username

Jkrzy:
## Description

Small tweak for https://github.com/18F/tock/issues/733#issuecomment-370942684

To create a development admin account, users should call ``createsuperuser`` with a non-email username.

Manually inserting an email username breaks ``utilization.GroupUtilizationView.get_queryset`` as email addresses are not caught by the current regex for the ``reports:ReportingPeriodUserDetailView`` urlpattern.

## Additional information

Not expected to encounter non-conforming usernames outside of development as ``@DOMAIN`` is truncated on user creation. There's only a single authorized domain at the moment,  this approach may lead to issues down the line if/when others are added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ If your team uses Tock and Slack, you might also find the ["angrytock" reminder 
   $ docker-compose build
   $ docker-compose run app python manage.py migrate
   $ docker-compose run app python manage.py loaddata test_data/data-update-deduped.json
-  $ docker-compose run app python manage.py createsuperuser --username admin@gsa.gov --email admin@gsa.gov --noinput
+  $ docker-compose run app python manage.py createsuperuser --username admin --email admin@gsa.gov --noinput
   ```
 
 1. Once the above commands are successful, run:
--- a/README.md
+++ b/README.md
@@ -24,7 +24,12 @@ If your team uses Tock and Slack, you might also find the ["angrytock" reminder 
   $ docker-compose build
   $ docker-compose run app python manage.py migrate
   $ docker-compose run app python manage.py loaddata test_data/data-update-deduped.json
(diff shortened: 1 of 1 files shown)


rogeruiz:
This looks good. I’m curious why we need to pass a username flag at all. An email should be all we need and then have the app handle the username handle for us. Wdyt?

Also `admin.user@gsa.gov` would help us ensure that usernames are being created appropriately too.

Jkrzy:
We're using the [Django provided](https://docs.djangoproject.com/en/1.11/ref/django-admin/#createsuperuser) ``createsuperuser`` which requires both username and email. Pretty standard Django dev workflow.

We could roll our own management command, that routes through the existing user creation functionality, but there's not much to be gained from a testing perspective. The user creation path is already covered by tests.

README.md
```diff
@@ -24,7 +24,7 @@ If your team uses Tock and Slack, you might also find the ["angrytock" reminder
   $ docker-compose build
   $ docker-compose run app python manage.py migrate
   $ docker-compose run app python manage.py loaddata test_data/data-update-deduped.json
-  $ docker-compose run app python manage.py createsuperuser --username admin@gsa.gov --email admin@gsa.gov --noinput
+  $ docker-compose run app python manage.py createsuperuser --username admin --email admin@gsa.gov --noinput
```
rogeruiz: @Jkrzy, What about making it: _new lines **just** for readability here_

```sh
docker-compose run app \
    python manage.py \
    createsuperuser \
    --username admin.user \
    --email admin.user@gsa.gov \
    --noinput
```

Jkrzy: Great thought! That would stick out as abnormal on the employees URL path.

PR #758, opened 2018-03-14
Include 'grade' in timecards.json response.

### LINKED ISSUE #756: Add grade info to /api/timecards.json output
Employee grade (e.g., GS 15) info is stored in the database as of #461, but it's not exported as part of the `/api/timecards.json` output, which I'm using for my [Google Apps Scripts / Tock integration](https://github.com/18F/tock-gas-ts).  It'd be great if it could be included here, so that spreadsheets can accurately calculate budgeting data.

While the output is denormalized, it seems useful to include it since an employee's grade can change over time, which means that at different points in a project, the employee could charge different rates if they are promoted. I think. Will have to make sure of this.

toolness:
Fixes #756.

## To do

- [x] Document this in `api-docs/timecards.md`.
- [x] ~~Should we convert the grade from an integer to an actual grade level string, e.g. `"GS14"`?~~ Decided not to do this; one advantage of keeping it as a number is that it's very easy to compare one grade to another.
- [x] In #461, Patrick says with regard to providing grade information via a superuser-only view, "The purpose of this is to limit the availability of employee-level grade information and protect the privacy of our employees".  Do we only want to provide the grade information if the requester (or the token associated with them) is a superuser? Right now we're just returning the information all the time.

  **Update:**

    I emailed the GSA's Chief Privacy Officer about this and he said:

    > Current GS level is considered "releaseable" under GSA Policy, meaning that it is not confidential/sensitive info:  https://insite.gsa.gov/portal/content/674050

    So we should be good to go!

### DIFF
--- a/api-docs/projects.md
+++ b/api-docs/projects.md
@@ -32,6 +32,7 @@ To fetch a list of projects  information.
         "billable": true,
         "start_date": null,
         "end_date": null,
+        "grade": 16,
         "active": true
     },...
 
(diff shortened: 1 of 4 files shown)


toolness:
Ok I think this is good to go!

PR #761, opened 2018-03-19
Fix 404s on static files we control

### LINKED ISSUE #665: Fix borked static asset references
While looking into an unrelated issue I noticed several errors show up in the logs while fetching static assets:

```
ERR Not Found: /static/img/uswds/close.svg
ERR WARNING:django.request:Not Found: /static/img/uswds/close.svg
ERR Not Found: /tock/static/js/vendor/underscore-min.js
ERR WARNING:django.request:Not Found: /tock/static/js/vendor/underscore-min.js
```

There might be other related ones, but in any case these references need to be fixed to ensure functionality doesn't break anywhere.

### LINKED ISSUE #760: Underscore not loading and potentially unused
I noticed that loading tock throws `Refused to execute script from 'https://tock.18f.gov/tock/static/js/vendor/underscore-min.js' because its MIME type ('text/html') is not executable, and strict MIME type checking is enabled.`. But it's not clear from a quick skim that underscore is actually used. If it *is* used, we probably don't need to both include it in `package.json` and vendor the minified source--which we're doing for a few other assets as well.

rogeruiz:
Closes #665 and closes #760


tock/tock/templates/_navigation.html
```diff
@@ -1,7 +1,7 @@
 <nav role="navigation" class="usa-nav">
   <div class="usa-nav-inner">
     <button class="usa-nav-close">
-      <img src="/static/img/uswds/close.svg" alt="close">
+      <img src="/tock/static/img/uswds/close.svg" alt="close">
```
tbaxter-18f: Wouldn't you just want to use {{ STATIC_URL }} or something like that?

rogeruiz: Thanks. Updated it in a0b98fd

Issue #704, opened 2018-01-22
Update all Tock dependencies


ccostino:
A handful of the Python dependencies in Tock need to be updated to their latest versions.  Some additional testing and verification will be needed to make sure all of the upgrades integrate seamlessly.

Updating Django is a separate issue (#699) as that is going to be required for ATO.

PR #762, opened 2018-03-20
Transition to Pipenv and update python/js dependencies

Jkrzy:
## Description

Per #704, Transition requirements.txt to Pipfile.

Pipfile.lock included with current versions of all dependencies. Django is pinned at 1.11 until we're ready for that transition. DRF is pinned at 3.6, code changes required for >=3.7.

Dockerfile revised to install packages with Pipenv.


rogeruiz:
@jkrzy are these changes going to affect the start commands for the build pack?

http://docs.cloudfoundry.org/buildpacks/python/index.html

rogeruiz:
Also @jkrzy, looks like you’ll need to update `.circleci/config.yml` to use the new commands. Might make sense to have CI leverage a script now rather than raw shell commands.

Jkrzy:
Thanks for the [wip] review @rogeruiz! The CF buildpack generates a ``requirements.txt`` from the Pipfile, no additional changes needed there but deploy docs do need a touch-up.

Doc and CI changes are incoming, need to familiarize myself with the deploy process before I update the documentation on it 😃 🏫

rogeruiz:
Let me know if you need CF credentials. I can give you access to staging.

.circleci/config.yml
```diff
@@ -57,6 +57,7 @@ jobs:
             python manage.py test --noinput --settings=tock.settings.test
             bandit -r . -x docker_entrypoint.py
             flake8
+            pipenv check --dev
```
rogeruiz: Great, I'll add this to the SSP. What does this check do for the codebase? Could you link me to documentation or write up a short summary of the feature here?

Jkrzy: Ref: https://docs.pipenv.org/advanced/#detection-of-security-vulnerabilities

Quick summary: Pipenv leverages the [Safety](https://github.com/pyupio/safety) package to check dependencies against known vulnerabilities. Vulnerabilities are sourced from [Safety-DB](https://github.com/pyupio/safety-db) which are updated once a month by https://pyup.io/.

I would consider this a complement, not a replacement, to active vulnerability scanning a-la gemnasium/snyk/etc.

Jkrzy:
@rogeruiz, js packages updated. This is ready for review at your leisure. 📦 ⬆️ 

In reviewing on my end, I didn't come across any breaking changes between the previous->these package versions. 

Per the earlier [comment](https://github.com/18F/tock/pull/762#issuecomment-374614672), I think the changes required here for the circleci config to resolve #704 are minor enough to not trigger a more broad re-working of `config.yml`. Lets tackle that in a new issue if it is a priority, what do you think???

rogeruiz:
@Jkrzy Looks like Gemnasium is [still saying we're out of date](https://gemnasium.com/github.com/18F/tock), but not sure why since we've updated most of these dependencies before you created this PR. 🤷‍♂️ Worth noting though: [Dependency Status](https://gemnasium.com/github.com/18f/tock)

rogeruiz:
Also, could be related, but it looks like Snyk, which may ultimately be the Gemnasium replacement, doesn't support `Pipfile` to scan for dependencies. Is there a way to still leverage `requirements.txt` with a `Pipfile`?

🐍 [source](https://snyk.io/docs/snyk-for-python)

Jkrzy:
Yeah, looks like Gemnasium does not support Pipfiles either. ☹️ 

Pipenv can generate `requirements.txt` from the `Pipfile.lock`, I'll follow-up with a PR to accommodate.

Pipenv having the [blessings of python.org](https://packaging.python.org/tutorials/managing-dependencies/) will hopefully drive adoption by external tools such as these.

Issue #764, opened 2018-03-22
Refactor GroupUtilization


tbaxter-18f:
Currently, the GroupUtilization view is pretty resource intensive, with calls to `/utilization` racking up some big numbers in Django Debug Toolbar.

This is just one example, but it's pretty typical:
```
Time 
CPU: 11440.00ms (19700.12ms)
SQL: 948 in 1635
```

With some optimization of the queries, we should be able to bring this down to much more reasonable numbers.

rogeruiz:
@tbaxter-18f what do you think about adding this to this body of work too? that way we can catch any other intensive queries in the future by looking at the logs.

https://www.dabapps.com/blog/logging-sql-queries-django-13/

tbaxter-18f:
I'm not averse, but I'm not sure it would help in this particular case, either. I don't think any of the queries here were _that_ intense (although there were a lot of joins), but rather that we had 950 queries for this page.

rogeruiz:
From my understanding of the post, this would display the number of queries for a particular request in the logs. That seems like it's useful information to possibly monitor. Isn't that what the Debug toolbar was showing as well? @tbaxter-18f

tbaxter-18f:
Perhaps -- and I'm just spitballing here -- but maybe it would be good to implement that as middleware? I think it'd be a separate story, but agree it may be worth pursuing.

PR #765, opened 2018-03-22
764 Refactored GroupUtilization for better performance.

tbaxter-18f:
## Description

#764 Refactored GroupUtilization for better performance. Simplifies queries, resulting queryset and view. Please double-check this carefully. There's a lot of changes, and I want to be sure I didn't break anything.

## Additional information
Typical Before: 
CPU: 11440.00ms (19700.12ms)
SQL: 948 in 1635

Typical after:
Time: 1580 (2900)
SQL: 125 in 255

### DIFF
--- a/tock/tock/templates/utilization/group_utilization.html
+++ b/tock/tock/templates/utilization/group_utilization.html
@@ -4,132 +4,129 @@
 {% endblock %}
 
 {% block content %}
+  <h2>Utilization by Unit</h2>
 
-<h2>Utilization by Unit</h2>
-{% if request.user.is_staff %}
-<h3>Notes:</h3>
-<p>
-  The following report contains users who are marked as billable in Tock,
-  organized by their unit as listed in Tock. Both attributes may be easily
-  updated via the <b>Employees</b> page of the Tock <a href="/admin">admin</a>
-  interface. Within each unit, employees are ordered alphabetically by last name.
-</p>
-<p>
-  The contents of this page may only be viewed by Tock users who are marked as
-  "staff" users. This attribute may be updated via the Tock
-  <a href="/admin">admin</a> interface, as well, via the <b>Users</b> page.
-</p>
-<p>
-  Utilization is calculated by dividing the total number of hours submitted on
-  projects that are marked "billable" in Tock, divided by the total number of
-  hours submitted on all projects for the given period.
-</p>
-<h3>Jump to:</h3>
-<ul>
-{% for i in unit_choices %}
-  <li><a href="#{{i.1}}">{{i.1}}</a></li>
-{% endfor %}
-</ul>
+  ---- {{ object_list }}
+  {% if request.user.is_staff %}
+    <h3>Notes:</h3>
+    <p>
+      The following report contains users who are marked as billable in Tock,
+      organized by their unit as listed in Tock. Both attributes may be easily
+      updated via the <b>Employees</b> page of the Tock <a href="/admin">admin</a>
+      interface. Within each unit, employees are ordered alphabetically by last name.
+    </p>
+    <p>
+      The contents of this page may only be viewed by Tock users who are marked as
+      "staff" users. This attribute may be updated via the Tock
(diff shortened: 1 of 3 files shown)


tbaxter-18f:
@toolness how can I work around that list index error in the test?

tock/tock/templates/utilization/group_utilization.html
```diff
@@ -4,132 +4,129 @@
...
+        </tbody>
+      </table>
+    {% endfor %}
 
-{% else %}
+  {% else %}
 
-<h3>This page is viewable by Tock users with an `is_staff` status of True.</h3>
-<p>If you need access, please contact your supervisor.</p>
+    <h3>This page is viewable by Tock users with an `is_staff` status of True.</h3>
```
rogeruiz: I know this was already here previously. But I don't see the value in including the `is_staff` property for users. I think this copy could be cleaned up to read something like.

### You don't have permission to view this information.
If you need access, please contact your supervisor.

tbaxter-18f: I agree, Roger. I actually pondered moving the is_staff check up into the view so we could and redirecting them before making all the queries, but in the end I didn't expand the scope to do that. I'm not averse to doing so, though, though.

I think the main  thing is to decide if you want this page handling both, or if we'd rather redirect to some sort of 403 template. My vote would be the latter, but either will work.

tock/utilization/tests/test_views.py
```diff
@@ -108,30 +108,30 @@ def test_utilization(self):
...
         self.assertTrue(response.context['through_date'])
         self.assertTrue(response.context['recent_start_date'])
-        self.assertEqual(len(response.context['user_list']), 2)
-        self.assertTrue(response.context['user_list'][0].__dict__['last'])
-        self.assertTrue(response.context['user_list'][0].__dict__['fytd'])
-        self.assertTrue(response.context['user_list'][0].__dict__['recent'])
+        self.assertEqual(len(response.context['object_list'][0]['billable_staff']), 2)
+        self.assertTrue(response.context['object_list'][0]['last'])
+        self.assertTrue(response.context['object_list'][0]['fytd'])
+        self.assertTrue(response.context['object_list'][0]['recent'])
```
ccostino: Good call on refactoring the calls to `__dict__`, thank you!

tock/utilization/views.py
```diff
@@ -11,261 +11,241 @@
...
+        """
+        Resolve recent reporting periods.
+
+        Although recent_rps is set to the last four reporting periods,
+        we could accept a form response that allows the user or app to
+        dynamically customize number of periods to include in the queryset.
+
+        Also, if they're not staff, we're going to go ahead and bounce
+        them to 403 so we don't make all these queries.
+        """
```
ccostino: :clap: on the comments!  Good on the 403 redirect as well.

tock/utilization/views.py
```diff
@@ -11,261 +11,241 @@
...
+                    timecard__reporting_period__start_date__gte=self.recent_rps[3]
+                ).select_related(
+                    'timecard__reporting_period',
+                    'project__accounting_code__billable'
+                ).values(
+                    'id',
+                    'hours_spent',
+                    'timecard__reporting_period',
+                    'timecard__reporting_period__start_date',
+                    'project__accounting_code__billable'
```
ccostino: Well done, good use of `values()` here too to cut things down query-wise.

tock/utilization/views.py
```diff
@@ -11,261 +11,241 @@
...
-            ).aggregate(
-                Sum('hours_spent'
+                """
+                We also need to know the billable cards, but
+                we only need the IDs to boil down each reporting period QS
+                and find the intersection below.
+                """
+                user_billable_timecard_ids = user_timecards.filter(
+                    project__accounting_code__billable=True
+                ).values_list('id', flat=True)
```
ccostino: Same, nice use of `values_list()` to get precisely what we need!

PR #752, opened 2018-03-09
ATO Documentation

rogeruiz:
This work encompasses the ATO documentation ~~and any application-level changes~~ related to the SSP.

_update_ see #766 for application level changes.

PR #766, opened 2018-03-23
Update logging for ATO

rogeruiz:
This body of work includes the updated logging Tock needs for the ATO.

You can test this by going to https://tock-logging.app.cloud.gov

You can find the logs in Kibana at https://logs.fr.cloud.gov and search for `@cf.app:"tock-logging" AND @message:"[...]"` with `[...]` being messaging from the #752 docs around logging.

### List of logging needed by the ATO
- [x] successful and unsuccessful account logon events
- [x] account management events
- [x] object access
- [x] policy change
- [x] privilege functions
- [x] process tracking
- [x] system events
- [x] all administrator activity
- [x] authentication checks
- [x] authorization checks
- [x] data deletions
- [x] data access
- [x] data changes
- [x] permission changes


bin/run.sh
```diff
@@ -17,5 +17,32 @@ else
...
+# Adding license key from script $(basename "${0}")
+api_key=${NEW_RELIC_API_KEY}
+EOF
+
+DEPLOYMENT_DESCRIPTION="Recording deployment of ${VERSION}."
+
+echo "${DEPLOYMENT_DESCRIPTION}"
+
+# Record deployment using the New Relic Python Admin CLI
+newrelic-admin record-deploy "${NEW_RELIC_CONFIG_FILE}" "${DEPLOYMENT_DESCRIPTION}"
```
ccostino: Oh, I think I see now... I had asked if this this additional New Relic setup needed or if the initialization in [`wsgi.py`](https://github.com/18F/tock/blob/master/tock/tock/wsgi.py#L16-L25) is enough, but this is being used to log the actual deployment, which is awesome if I'm understanding correctly! :-)

tock/tock/settings/base.py
```diff
@@ -34,13 +34,13 @@
...
-    'organizations',
-    'api',
-    'utilization',
+    'tock.apps.TockAppConfig',
+    'projects.apps.ProjectsAppConfig',
+    'hours.apps.HoursAppConfig',
+    'employees.apps.EmployeesAppConfig',
+    'organizations.apps.OrganizationsAppConfig',
+    'api.apps.ApiAppConfig',
+    'utilization.apps.UtilizationAppConfig',
```
ccostino: Oh 👏 on the better app organization!

rogeruiz:
thanks @ccostino yeah. that's thanks to @toolness work around updating to `1.11` which suggests the `AppConfig` approach for sub-apps. 

This PR is ready to be reviewed. All the events should be logged in Kibana, which is where the accessors will be verifying that the logs are being recorded properly. Let me know if there are any logs I forgot, or if the logging doesn't make sense from a language perspective.
cc @annalee 

[Docs for ATO logging is here](https://github.com/18F/tock/blob/ato-documentation/docs/logging.md).

tock/employees/signals.py
```diff
@@ -0,0 +1,33 @@
+import logging
```
tbaxter-18f: I'm not saying there's anything wrong here, but I definitely like overriding `save()` on the model more than signals. I think it's more transparent, less likely to fail silently, and avoids weird circular dependency problems. Again, not asking for any changes, just throwing it out there.

tock/tock/settings/base.py
tbaxter-18f: Given than employees, organizations, etc, are the apps, this is actually a bit confusing to my eyes.

tbaxter-18f:
I agree with Annalee. I definitely don't want my general opinions about signals to be a blocker, and there _is_ a big advantage in keeping all of those in one place.

rogeruiz:
@ccostino @annalee This is good to be reviewed now against the [updated ATO logging docs located here](https://github.com/18F/tock/blob/ato-documentation/docs/logging.md).

tock/employees/signals.py
ccostino: I'm +1 with on you @tbaxter-18f regarding the signals in general, but I have also run through this with @rogeruiz and we're good to go here!

PR #768, opened 2018-03-28
Pipenv generated requirements.txt and dependency updates

Jkrzy:
## Description

Continuation of PR #762 towards resolution of #704.

Adding `requirements.txt` back into repository to accommodate Gemnasium which does not yet support `Pipfile`.

Updated `Pipfile.lock` and `package-lock.json` with current package versions from the past ~week and included a `README.md` section on dependency management.


README.md
```diff
@@ -60,6 +60,32 @@ docker-compose run app bandit -r . -x manage.py,docker_entrypoint.py
 docker-compose run app flake8
 ```
 
+### Dependency management
```
rogeruiz: @Jkrzy this is great! can you remove this from the main readme and add it to a single file within `/docs`. Perhaps something like `/docs/dependency-management.md`. I'll take care of adding the link in the `/docs/README.md`. Thanks! See #752 for more info on `/docs`.

README.md
```diff
@@ -60,6 +60,32 @@ docker-compose run app bandit -r . -x manage.py,docker_entrypoint.py
...
+In development and continuous integration the python environment is 
+established with `pipenv install --dev`. To acommodate 3rd party tools
+which do not yet support [Pipenv] a `requirements.txt` file, containing 
+only production dependencies, is generated by [Pipenv] and included in 
+the repository.
+
+Development and production dependencies can be updated locally with:
+```sh
+pipenv update --dev
+pipenv lock --requirements > requirements.txt
```
rogeruiz: ❤️ 🐍 Great! do we also need a command for `requirements-dev.txt`?

rogeruiz:
@Jkrzy also, thinking this over a bit more, could you change the base branch from `master` to `ato-documentation`? I have some edits I want to make on top of `/docs/dependency-management.md`.

README.md
Jkrzy: Good question... at the moment I think not.

I don't *believe* the Snyk/Gemnasium integrations can handle separate `requirements.txt` gracefully. If we include dev dependencies within `requirements.txt` either directly or by reference, cloud foundry is going to pick them up and install them on every deploy.

I see that dev dependencies were previously scanned by Gemnasium, and not noted as development,  but I'm not sure when/how that scan was run 😕 . https://gemnasium.com/github.com/18F/tock

Gemnasium toolbelt maybe in February?? cc: @ccostino

rogeruiz: oh strange on the dev scan in Gemnasium.

But, I figured it we could do it like below, but I'm not familiar enough with the tool. `requirements.txt` alone for now is good enough for me. I don't want to lose dependency scanning after #762 😁 

```shell
pipenv lock --requirements > requirements.txt
pipenv lock --requirements --dev > requirements-dev.txt

Jkrzy:
Rebased against `ato-documentation`

PR #772, opened 2018-04-02
Recycle production instances on a schedule

rogeruiz:
This work should be recycling the production instances of Tock on a schedule at 5am every day. This should cause no downtime as it uses [the CF Recycle plugin](https://github.com/rogeruiz/cf-recycle-plugin).


rogeruiz:
This is ready to be reviewed. You can see a working cron recycle job [here](https://circleci.com/gh/18F/tock/541).

PR #776, opened 2018-04-04
Rename credentials service in production

rogeruiz:
This is good to be reviewed and merged, but let's not cut a release for this just yet as we're in the midst of a code-freeze for the ATO assessment.

### DIFF
--- a/manifest-production.yml
+++ b/manifest-production.yml
@@ -17,4 +17,4 @@ applications:
   instances: 4
   services:
   - tockdb
-  - tock-creds-prod
+  - tock-credentials


ccostino:
@rogeruiz once this is merged in, are any manual steps still needed at all, like perhaps once more to get things squared away?  Or will merging be enough?

rogeruiz:
@ccostino merging will be enough, _and cutting the Git tag release_, but the old `tock-creds-production` will need to be manually deleted. but the `tock-credentials` service exists and is bound to the app along with the other credentials service as well.

ccostino:
Got it, thanks!

PR #778, opened 2018-04-10
[work in progress] Add support for pulling TimecardPrefillData from Google Apps Suite

rogeruiz:
This body of work adds API support to Tock for receiving data from the Staffing Snapshot Google Document that is maintained by @abrouilette.

### DIFF
--- a/tock/api/views.py
+++ b/tock/api/views.py
@@ -7,7 +7,12 @@ from django.contrib.auth.models import User
 from django.contrib.auth import get_user_model
 
 from projects.models import Project
-from hours.models import TimecardObject, Timecard, ReportingPeriod
+from hours.models import (
+    TimecardObject,
+    Timecard,
+    ReportingPeriod,
+    TimecardPrefillData
+)
 from employees.models import UserData
 
 from rest_framework import serializers, generics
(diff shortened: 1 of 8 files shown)


rogeruiz:
Running into the N+1 query here when trying to write tests for this. 😞

tock/api/views.py
```diff
@@ -358,3 +387,68 @@ def hours_by_quarter_by_user(request, *args, **kwargs):
...
+        try:
+            userdata = UserData.objects\
+                       .prefetch_related('user')\
+                       .get(user__username=username)
+        except UserData.DoesNotExist:
+            raise Http404
+
+        return TimecardPrefillData.objects\
+               .prefetch_related('employee')\
+               .filter(employee_id=userdata.id)
```
Jkrzy: @rogeruiz, I think the source of the N+1 exceptions is here.  The queries can be tweaked a bit to pass the NPLUSONE checks:

```python
if not UserData.objects.filter(user__username=username).exists():
    raise Http404

return TimecardPrefillData.objects.filter(employee__user__username=username)\
.select_related('employee', 'project')
```

tbaxter-18f:
@rogeruiz what's the status of this?

Issue #653, opened 2017-08-19
Add support for late Tock entry notifications


ccostino:
It'd be a very nice thing for our Ops team to add support for notifications to be sent out to late Tock entrants automatically as opposed to someone having to manually do this.  There are a few ways we could do this, listed below:

* [ ] Email notifications via a tool like ~Mandrill~ and scheduled jobs through Celery (Mandrill isn't an option anymore)
* [ ] Possible Slack integration via a bot like [angrytock](https://github.com/18F/angrytock) (depends on ATO status and procedure)
* [x] Client-side browser notifications, e.g., PR #647 (thanks, @mgwalker!)

More ideas and details to come in the future as we have time to work through the user needs of this.

mgwalker:
For Slack integration, it could also be done with [Charlie](https://github.com/18f/18f-bot).  The pro of using Charlie is one less bot to maintain.  The con is that angrytock already has integration code, so that would have to be reinvented.  😛

ccostino:
Good to note, thanks!  If it wouldn't take much to get Angrytock back up and running that might the path of least resistance here. :-)  That said, this is much lower on the priority list compared to other things, but we at least have your browser notifications in place! 🎉

ccostino:
We'll investigate including Angrytock in the ATO; otherwise, a note that notifications should show up at about 9:30 - 10:00 AM local time Monday morning for users who are late.

tbaxter-18f:
Some effort has been put into this in the related 18F/tock#779 [WIP] - Notifications integration PR, but someone will need to pick it up and bring it the rest of the way in.

PR #779, opened 2018-04-15
[WIP] - Notifications integration

jontours:
## Description
This is a possible integration for adding notifications to Tock.  Some integration work still to go.  In fact, perfectly ok for this feature to never see the light of day, but to maybe get the conservation going.

This work was to demonstrate feasibility for the Notifications-as-a-Service 10X.

I'm not a Django programmer but this was the simplest way to get a simple webform started with limited impact.

image

### DIFF
--- a/tock/notifications/views.py
+++ b/tock/notifications/views.py
@@ -6,7 +6,12 @@ from django.views.generic.edit import FormView
 def index(request):
     return HttpResponse("Hello, world. You're at the polls index.")
 
-class NotificationsListView(ListView):
+class Notification():
+    notification_label = "Friday alert"
+    notification_text = "Tock your Time!"
+
+class NotificationsListView(DetailView):
     template_name = 'notifications/notifications_list.html'
-    def get_queryset(self):
-        return
+
+    def get_object(self, **kwargs):
+        return Notification()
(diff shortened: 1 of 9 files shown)


tock/notifications/views.py
```diff
@@ -0,0 +1,17 @@
...
+from django.views.generic.edit import FormView
+
+def index(request):
+    return HttpResponse("Hello, world. You're at the polls index.")
+
+class Notification():
+    notification_label = "Friday alert"
+    notification_text = "Tock your Time!"
+
+class NotificationsListView(DetailView):
```
tbaxter-18f: Just curious: Is there a reason to call it ListView but subclass DetailView?

tbaxter-18f:
I think there's the start of a good idea here, but I have two extra thoughts:

1. Could we leverage Django's built-in messages framework for the on-site messages themselves?
2.  Is this a rare good use of push notifications?

tbaxter-18f:
I'm relating this to #653 so we can take a holistic approach to notifications.

jontours:
This work has been discussed a bit lately.  But more in the context of maybe utilizing it as a cloud.gov broker.  The UK GDS has rolled out a shared notifications service that is mostly similar to the ideas @amoose and I threw around.

tbaxter-18f:
I'm definitely intrigued, and when I -- or whoever -- picks this up again, let's definitely meet up and walk through all these options.

tbaxter-18f:
Given the age of this and the work @mgwalker has been doing with Slack integration and notifications, I'm going to go ahead and close this.

Issue #804, opened 2018-05-07
'Save Timecard' and 'Submit Timecard' should be visually distinguishable


pburkholder:
I have on two or three occasions 'Saved' my timecard when I meant to 'Submit'

It might help to have them visual distinct to indicate that Submit is the critical element to act on.  Or, 'Save' could pull up an dialog element to say 

> Timecard saved. Be sure to submit hours by 08:00 EST [CONTINUE]

Thanks, Peter

ccostino:
Good call, @pburkholder!  Will take a look and adjust this. :-)

tbaxter-18f:
Given the improvements done in the related #807 I think this can be closed. If anyone objects, I'll happily reopen.

PR #807, opened 2018-05-14
Improve timecard form UX

ccostino:
Address #804 

This changeset is an attempt to improve the user experience a bit around the buttons and info/error messages when saving or submitting the form.  Specifically, it addresses the following:

* Swaps the order of the save and submit buttons
* Makes the submit button noticeably larger
* Changes the style of the save button to be a secondary button
* Shifts form info and error messages to appear at the top and bottom of the timecard for improved visibility
* Provides alert titles to the messages displayed
* Fixes the display of most alerts
* Adds a submission success mesage

This is just a start; more improvements can certainly be made!

### DIFF
--- a/tock/hours/views.py
+++ b/tock/hours/views.py
@@ -899,12 +899,20 @@ class TimecardView(PermissionMixin, UpdateView):
 
     def get_success_url(self):
         if self.object.submitted:
-            return reverse("ListReportingPeriods")
+            messages.add_message(
+                self.request,
+                messages.INFO,
+                ('Time card for {timecard} submitted successfully.'.format(
+                    timecard=str(self.object.reporting_period.start_date)
+                ))
+            )
+
+            return reverse('ListReportingPeriods')
         else:
             messages.add_message(
                 self.request,
                 messages.INFO,
-                ("Timesheet saved. "
+                ("Time card saved. "
                  "Please remember to submit your timesheet when finished.")
             )
 
--- a/tock/hours/views.py
+++ b/tock/hours/views.py
@@ -901,10 +901,11 @@ class TimecardView(PermissionMixin, UpdateView):
         if self.object.submitted:
             messages.add_message(
                 self.request,
-                messages.INFO,
-                ('Time card for {timecard} submitted successfully.'.format(
+                messages.SUCCESS,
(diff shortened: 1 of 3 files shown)


ccostino:
Hopefully this will help some with the user experience folks have when submitting time cards!  Here are some screenshots of the changes:

New buttons
screen shot 2018-05-14 at 12 03 58 pm

New notifications
screen shot 2018-05-14 at 11 52 05 am
screen shot 2018-05-14 at 12 04 25 pm

New notification on reporting period list after submitting a time card
screen shot 2018-05-14 at 12 22 03 pm

ccostino:
I've made some additional improvements to this!  Brought to you while listening to The Final Countdown and Eye of the Tiger among other things of that nature...

Saving the time card:
screen shot 2018-05-14 at 11 07 15 pm

Submitting the time card:
screen shot 2018-05-14 at 11 07 34 pm

:-)

ccostino:
Last improvement and then it's off to bed for me, because Sweet Dreams (Are Made of This)...

Improved error alert!
screen shot 2018-05-14 at 11 17 01 pm

line47:
These changes are great @ccostino I love the better alerts and different button styles! ❤️ 

One thing to note is that some folks were previously having issue of accidentally submitting their timecard which is why the save action was first and the submit action was second. Ideally i think some sort of way for folks not to be able to submit until appropriate hours are met would be a great addition to help stop some of those accidental submissions.

ccostino:
Thanks so much for taking a look, @line47! :-)

I know what you mean about the buttons and folks still accidentally submitting when they're not done; this was the best I could do given what I was seeing and some of the feedback given.  Tock will prevent you from submitting if you have too few or too many hours, but that's about it.

Do you think it'd be better to keep the order of the buttons as they were but with the new styles and text?  Or does it make more sense with them swapped again?

Our time is limited on this and I caught a bit of a break while at PyCon to work through these things, but if you'd like to expand on your thoughts a bit about restricting submissions and file an issue, we'd appreciate it! :-)

Thanks again for taking a look!

line47:
Maybe leave it for now and see if the issue comes back up about accidental submissions. If it does then switch the order to save first with submit second 👍

ccostino:
Awesome, sounds good to me! :-)

PR #813, opened 2018-05-31
[wip] Initial draft for Alert markdown extension

rogeruiz:
This PR is to be merged onto #808. This work adds an extension for the Markdown library to allow for class on heading elements to contain the `usa-alert-heading` class in order to render properly in the `<div class="usa-alert-body">`.

### DIFF
--- a/tock/markdown_extensions/add_alert_header_class.py
+++ b/tock/markdown_extensions/add_alert_header_class.py
@@ -1,0 +1,8 @@
+import re
+
+
+def set_alert_heading_class(self, element):
+    for child in element:
+        if re.search('h\d', child.tag, re.IGNORECASE):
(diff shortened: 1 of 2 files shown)


rogeruiz:
Documentation about this exists here. https://github.com/Python-Markdown/markdown/wiki/Tutorial:-Writing-Extensions-for-Python-Markdown and here https://python-markdown.github.io/extensions/api/

tbaxter-18f:
I don't think I understand. If the parent element already has `usa-alert-body` why does the child need a class? And if it's a consistent pattern, can we just mix `usa-alert-heading` into `usa-alert-body h1` in the scss? i'm pretty sure there's context I'm missing, but this feels kinda heavy-handed just to get styling on a child.

Issue #810, opened 2018-05-21
[high] Logout Function does not Terminate Sessions


rogeruiz:
**Location**:  https://tock.18f.gov/logout

**Description**:  When using the logout function of the application it was identified that user session are not appropriately terminated. After a user clicks logout they are able to re-enter the application as the same authenticated user without re-authenticating.

**Recommendation**:  When a user logs off, or when any other session termination event occurs, the application must terminate the user session to minimize the potential for an attacker to hijack that particular user session.

**References**:
- https://www.owasp.org/index.php/Testing_for_logout_functionality_(OTG-SESS-006)

rogeruiz:
closed by #814

PR #814, opened 2018-06-13
Terminate all UAA sessions at the server level

rogeruiz:
This addresses #810 with the following solution.

```
If the user is authenticated
  Clear Django user session
  Redirect to UAA server with Client ID and Redirect parameters
  Redirect to application logout page
If the user isn't authenticated
  Show the application logout page
```

### DIFF
--- a/tock/tock/signals.py
+++ b/tock/tock/signals.py
@@ -12,7 +12,7 @@ def successful_login(sender, request, user, **kwargs):
 
 
 def successful_logout(sender, request, user, **kwargs):
-    logger.info(f'Successful logout event for {user.username}.')
+    logger.info(f'Successful logout event for {user}.')
 
 
 def failed_login(sender, credentials, request, **kwargs):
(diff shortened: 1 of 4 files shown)


rogeruiz:
In order to deploy this to production, we need to add the `https://tock.18f.gov/logout` URL to the redirects array in the client.

tock/tock/signals.py
```diff
@@ -12,7 +12,7 @@ def successful_login(sender, request, user, **kwargs):
 
 
 def successful_logout(sender, request, user, **kwargs):
-    logger.info(f'Successful logout event for {user.username}.')
+    logger.info(f'Successful logout event for {user}.')
```
ccostino: Wouldn't it be better to leave this as `user.username`, or are you going for the `repr`/`str` defined on the model instead by doing this?

rogeruiz: The value that got sent over after the redirect was no longer a Django user but rather the string of the user. This was occurring after the redirect I believe. when I user `user.username` it was hitting an error. I figured it's also easier to just rely on the `str` defined on the model anyway here. It never felt right to drill down into when the string representation of the `User` model could change and be more useful.

rogeruiz:
Whoops, gotta fix the tests here! 🐠

Issue #812, opened 2018-05-21
[medium] Sessions do not timeout


rogeruiz:
**Location**:  https:/tock.18f.gov/ 

**Description**:  User sessions do not timeout .

**Evidence**:  The user session timeout was tested in 10 minute, 1 hour, 4 hour, and 24 hour intervals to which re-authentication was not required to continue on the current user session.

**Recommendation**:  All applications should implement an idle or inactivity timeout for sessions. This timeout defines the amount of time a session will remain active in case there is no activity by the user, closing and invalidating the session upon the defined idle period since the last HTTP request received by the web application for a given session ID.

**References**:
- https://www.owasp.org/index.php/Test_Session_Timeout_(OTG-SESS-007)

PR #815, opened 2018-06-15
Add Auto Logout functionality

rogeruiz:
This PR closes and address #812

Right now the timeout is for 60 minutes. But for testing purposes I've set it to 3 seconds by overriding it in the test specifically. I tested this locally too with a setting of 1 minute and it worked beautifully. swedishchef-myspace

### DIFF
--- a/tock/tock/middleware.py
+++ b/tock/tock/middleware.py
@@ -3,25 +3,34 @@ from django.contrib import auth
 from datetime import datetime, timedelta
 
 
-class AutoLogout():
-    def process_request(self, request):
-        # Check if user is logged in at all
-        if not request.user.is_authenticated():
-            return
+class AutoLogout(object):
 
-        fmt = '%Y%m%d%H%M%S'
-        # Compare the time of the last activity with the logout delay
-        try:
-            session_time = datetime.strptime(
-                request.session['tock_last_activity'],
-                fmt
-            )
-            if datetime.now() - session_time > \
-               timedelta(0, settings.AUTO_LOGOUT_DELAY_MINUTES * 60, 0):
-                auth.logout(request)
-                del request.session['tock_last_activity']
-                return
-        except KeyError:
-            pass
+    def __init__(self, get_response):
+        self.get_response = get_response
 
-        request.session['tock_last_activity'] = datetime.now().strftime(fmt)
+    def __call__(self, request):
+
+        # Check if user exists and is logged in
+        if request.user and request.user.is_authenticated():
+
+            logout_time_in_seconds = settings.AUTO_LOGOUT_DELAY_MINUTES * 60
+            fmt = '%Y%m%d%H%M%S'
+
+            # Compare the time of the last activity with the logout delay
+            try:
+                session_time = datetime.strptime(
+                    request.session['tock_last_activity'],
+                    fmt
+                )
+                if datetime.now() - session_time > \
+                   timedelta(seconds=logout_time_in_seconds):
+                    auth.logout(request)
+                    del request.session['tock_last_activity']
+                    return self.get_response(request)
+            except KeyError:
+                pass
+
+        request.session['tock_last_activity'] = \
+            datetime.now().strftime(fmt)
+
+        return self.get_response(request)
--- a/tock/tock/middleware.py
+++ b/tock/tock/middleware.py
@@ -9,12 +9,12 @@ class AutoLogout(object):
         self.get_response = get_response
 
     def __call__(self, request):
+        fmt = '%Y%m%d%H%M%S'
 
         # Check if user exists and is logged in
         if request.user and request.user.is_authenticated():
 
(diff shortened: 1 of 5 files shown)


tock/tock/middleware.py
```diff
@@ -0,0 +1,27 @@
+from django.conf import settings
+from django.contrib import auth
+from datetime import datetime, timedelta
+
+
+class AutoLogout():
```
ccostino: Well done!  Nice and succinct. :-)

However, this does appear to be constructed with the [older style of middleware classes](https://docs.djangoproject.com/en/1.11/topics/http/middleware/#upgrading-pre-django-1-10-style-middleware), and I wonder if it might be worth getting this in shape with the [current middleware style](https://docs.djangoproject.com/en/1.11/topics/http/middleware/#writing-your-own-middleware) to help prep with a Django 2.0 update (technically this can be done now for 1.11, too).

This may or may not make for more work though given the other pieces of middleware, so there's an argument for not doing that right now, too.

tock/tock/tests/test_middleware.py
```diff
@@ -0,0 +1,25 @@
...
+        self.assertIn('tock_last_activity', response_initial.client.session)
+
+        # Sleep for an arbirary five seconds
+        time.sleep(5)
+
+        response_after_expiry = self.client.get(
+            reverse('ListReportingPeriods')
+        )
+        self.assertEqual(response_after_expiry.status_code, 302)
+        self.assertIn('tock_last_activity', response_initial.client.session)
```
ccostino: ❤️ this test, thank you!

tock/tock/middleware.py
```diff
@@ -0,0 +1,27 @@
+from django.conf import settings
+from django.contrib import auth
+from datetime import datetime, timedelta
```
tbaxter-18f: Minor, but following convention the datetime import should be above the django imports, shouldn't it?

tock/tock/middleware.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class AutoLogout():
+    def process_request(self, request):
+        # Check if user is logged in at all
+        if not request.user.is_authenticated():
+            return
```
tbaxter-18f: I would consider either 

```
if request.user and request.user.is_authenticated():
    do stuff
```
or
```
if not request.user or request.user.is_anonymous():
    do stuff
````
I may be over-guarding, but I'd be watching for any weird cases where request.user wasn't as expected.

File: tock/tock/middleware.py [hunk 5]
```diff
@@ -0,0 +1,27 @@
...
+
+        fmt = '%Y%m%d%H%M%S'
+        # Compare the time of the last activity with the logout delay
+        try:
+            session_time = datetime.strptime(
+                request.session['tock_last_activity'],
+                fmt
+            )
+            if datetime.now() - session_time > \
+               timedelta(0, settings.AUTO_LOGOUT_DELAY_MINUTES * 60, 0):
```
tbaxter-18f: You could maybe make this more readable by changing the import to 
```
from django.conf.settings import AUTO_LOGOUT_DELAY_MINUTES as LOGOUT_TIME
```
or, alternately, doing the math earlier:
```
LOGOUT_TIME = settings.AUTO_LOGOUT_DELAY_MINUTES * 60
```

Which would in turn let you get this cleanly on one line:
```
if datetime.now - session_time >  timedelta(0, LOGOUT_TIME):
```
Also, I don't think the trailing 0 arg is needed is it?

tbaxter-18f:
Everything I see is either really minor formatting stuff or just personal preference. I think it looks good.

rogeruiz:
alrighty, this is ready for review and merge. Thanks @ccostino @tbaxter-18f

tock/tock/middleware.py
```diff
@@ -0,0 +1,36 @@
...
+            fmt = '%Y%m%d%H%M%S'
+
+            # Compare the time of the last activity with the logout delay
+            try:
+                session_time = datetime.strptime(
+                    request.session['tock_last_activity'],
+                    fmt
+                )
+                if datetime.now() - session_time > \
+                   timedelta(seconds=logout_time_in_seconds):
```
tbaxter-18f: Still doesn't fit on one line? Those dang line lengths :-D

tock/tock/middleware.py
```diff
@@ -0,0 +1,36 @@
...
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+
+        # Check if user exists and is logged in
+        if request.user and request.user.is_authenticated():
+
+            logout_time_in_seconds = settings.AUTO_LOGOUT_DELAY_MINUTES * 60
+            fmt = '%Y%m%d%H%M%S'
```
tbaxter-18f: You need to move this up and out of the `if` block, or L34 fails. That looks like what's making Circle unhappy.

PR #818, opened 2018-08-13
Update docs

tbaxter-18f:
## Description

Clarifies how API users can get started and adds a README for additional clarity.


docs/onboarding.md
```diff
@@ -21,6 +21,7 @@ checklist. Replace {TockDeveloper} with the name of the onboarding developer.
 - [ ] {TockDeveloper} has been invited to #tock-dev in Slack.
 - [ ] {TockDeveloper} is familiar with the [deployment process of
   Tock](https://github.com/18f/tock/tree/master/docs/deployment-process.md).
+- [ ] {TockDeveloper} has access to `GSATETockLibrary` and `GSATEUtilitiesLibrary` scripts.
```
rogeruiz: Can we explain how someone would get access to these scripts? I had Angela look through the Google Apps Script UI interface via search and asked her to grant me permission. I think a line along the thread of “ask the person who previously had access or the product owner of Tock for access”.

api-docs/README.md
```diff
@@ -0,0 +1,24 @@
+# Tock API documentation
+
+## Before you start
+You'll need to have an API key, and you'll need to have it added to the `GSA_TE_Tock_Library` script. Contact a Tock developer or ping the `#tock-dev` slack channel for help.
```
rogeruiz: I’m not sure this makes sense here. As the API documentation is for our API, but as Tock devs, we don’t take ownership of external libraries using it. Perhaps we should? Then I’d say that the source for those libs should be in this directory under, GAS examples or something. Wdyt?

api-docs/README.md
```diff
@@ -0,0 +1,24 @@
...
+- [Timecards](timecards.md): Fetches a list of all submitted timecards and related information.
+- [User data](user-data.md): Fetches a list of all users, along with organizational information for each.
+- [Users](users.md): Fetches a list of all users, along with basic information about them.
+
+--------
+
+The Tock API documentation is written in
+[GitHub-flavored markdown][gh-md] and is best read using the GitHub interface.
+
+[gh-md]: https://guides.github.com/features/mastering-markdown/#GitHub-flavored-markdown
```
rogeruiz: Where is this documentation coming from? Is it tangential to this work? It’s okay if it is. I’m just unclear on what it’s doing in the pr.

api-docs/README.md
tbaxter-18f: Well, the docs were already in this folder, just lacking an index/readme. So by creating one we improve the navigability and usefulness of those docs, provide some brief summary AND have a handy spot to say, "Hey, API user, you're gonna need a key."

That was my thinking anyway.

api-docs/README.md
tbaxter-18f: I put it here to keep it parallel with the dev docs where it was already noted, but I don't have strong opinions about it one way or the other.

docs/README.md
```diff
@@ -16,4 +13,11 @@ Welcome to the Tock documentation! This documentation is written in
 - [Logging](logging.md)
 - [Backup and Restore Database](backup-and-restore-database.md)
 
+If you are a consumer of Tock APIs and data, you may want to consult the [API documentation](../api-docs)
+
+-----------
+
+The Tock documentation is written in
+[GitHub-flavored markdown][gh-md] and is best read using the GitHub interface.
```
tbaxter-18f: I did move this down. It's just not important enough to be the top-most thing on the page.

docs/onboarding.md
tbaxter-18f: The instructions are for the onboarder to load into a GH issue, but I have added a line to clarify that they have to do that for the new, onboarding developer. .

tbaxter-18f:
@rogeruiz could you take another look at this and see if we can close it out?

docs/google-script-integration.md
```diff
@@ -0,0 +1,11 @@
+## Google Script Integration
+
```
rogeruiz: Can we link to this page from `docs/README.md`

tbaxter-18f: Done now.

Issue #825, opened 2018-09-26
Fix inactive user login redirect loop


tbaxter-18f:
Currently, if someone is not an active user in tock, they get thrown into an nasty redirect loop. We should catch the user-is-not-active case and instead direct them to a simple unauthenticated page that says something like _"you are inactive in Tock, please reach out to the Tock team in the #tock Slack channel for help."_

*We'll know we're done when*
* Inactive users logging in resolve to a helpful page
* Tests are written

tbaxter-18f:
Resolved with merge of #826

PR #826, opened 2018-09-27
Raise PermissionDenied for inactive user accounts

Jkrzy:
## Description

For #825, deny access by and provide a help message to users with inactive accounts.

## Additional information

Prevents inactive users from completing authentication w/ tock backend.

### DIFF
--- a/tock/tock/remote_user_auth.py
+++ b/tock/tock/remote_user_auth.py
@@ -1,15 +1,16 @@
 import datetime
 import logging
 
-from uaa_client.authentication import UaaBackend
-from django.core.exceptions import ValidationError
 from django.conf import settings
 from django.contrib.auth.models import User
+from django.core.exceptions import PermissionDenied, ValidationError
+from uaa_client.authentication import UaaBackend
 
 from employees.models import UserData
 
 logger = logging.getLogger('tock-auth')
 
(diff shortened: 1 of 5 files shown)


tock/tock/remote_user_auth.py
```diff
@@ -5,11 +5,13 @@
 from django.core.exceptions import ValidationError
 from django.conf import settings
 from django.contrib.auth.models import User
+from django.core.exceptions import PermissionDenied
```
tbaxter-18f: Let's combine this with the ValidationError imported on L5 (and while we're at it, we can better match up our imports with best conventions)

Jkrzy: Whoops, missed that one thanks!

tock/tock/remote_user_auth.py
```diff
@@ -43,6 +44,12 @@ def verify_userdata(user):
 
 
 class TockUserBackend(UaaBackend):
+
+    def user_can_authenticate(self, user):
+        if not user.is_active:
+            raise PermissionDenied(ACCOUNT_INACTIVE_MSG)
```
tbaxter-18f: This is an interesting way of handling this. In my head, I think I saw either redirecting to a simple templateView or maybe leveraging the messages framework, but I think I like piggybacking on the 403.

PR #843, opened 2018-10-30
Fixing Local Development Environment

amymok:
## Description

Following [instructions](https://github.com/18F/tock/blob/master/docs/local-development.md) to bring up local development environment is currently broken.

## Additional information
Issues I faced when following the instructions:

- [x] `docker-compose run app python manage.py loaddata test_data/data-update-deduped.json`: This gave me a duplicate key issue.  See [output here](https://gist.github.com/amymok/6571999d32b81c90cc49561547d33e4a)
I am fixing it by removing all the `auth.permission` entries in `tock/test_data/data-update-deduped.json`, since they conflict with what was already in the database.

- [x] `docker-compose up`: This gave me an exception on `WhiteNoise`.  See [output here](https://gist.github.com/amymok/ab25109adef8e7546041597207f6e5b6)
I am fixing that by what was suggested from the [upgrade instructions here](http://whitenoise.evans.io/en/stable/changelog.html#v4-0)

- [x] Logging in with admin.user@gsa.gov: This gave me an authentication failure.  "cloud.gov Login Error: Authentication failed: authenticate_failed", Screenshot below:
screen shot 2018-10-30 at 7 10 39 pm

This is fixed by regenerating the Pipenv.lock file because the older version has two versions of Django installed which they conflict with each other.   (Thanks @Jkrzy for discovering that)

- [x] Did not comply with Flake8 rules.  Fixed.

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -3,19 +3,6 @@
         "hash": {
             "sha256": "565fdc5d1f74b8ad4e2435c6f0cd041df300651cf6e1c16f4b669718163e5d05"
         },
-        "host-environment-markers": {
-            "implementation_name": "cpython",
-            "implementation_version": "3.6.2",
-            "os_name": "posix",
-            "platform_machine": "x86_64",
-            "platform_python_implementation": "CPython",
-            "platform_release": "16.7.0",
-            "platform_system": "Darwin",
-            "platform_version": "Darwin Kernel Version 16.7.0: Thu Jun 21 20:07:39 PDT 2018; root:xnu-3789.73.14~1/RELEASE_X86_64",
-            "python_full_version": "3.6.2",
-            "python_version": "3.6",
-            "sys_platform": "darwin"
-        },
         "pipfile-spec": 6,
         "requires": {
             "python_version": "3.6"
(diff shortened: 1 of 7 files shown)


tbaxter-18f:
Amy, you said 
"I am fixing it by removing all the auth.permission entries in tock/test_data/data-update-deduped.json, since they conflict with what was already in the database."

But I think when you loaddata it should be on an empty DB. Did you run it twice? Is there another step that could have populated your DB?

amymok:
Yes, I am still trying to figure out that part.  I try to go into the database before any migration happens, and checked that `auth.permission` was empty, but somewhere during the migration, it is filled out.  I wasn't able to track how that happens yet (it could also mean something Django does that I am not understanding yet).  So the `auth.permission` table is filled completely with 80 something entries when migrations finished.

amymok:
@tbaxter-18f Okay, so it looks like that `auth.permission` table is loaded automatically by django.  As you create more models such as `organization`, `Project`, and everything else, it will create 4 entries per new model to view, add, change, delete.  So it makes sense to remove them from `data-update-deduped.json` bc it conflicts with what was already created.

rogeruiz:
We're cherry picking your work from 747c68b @amymok to fix `master` because it's currently breaking the deployment of staging. thanks for the fix! 👍

amymok:
@rogeruiz Please do.  I am still having problems with the authentication and not so sure what is happening yet.  I will try to look at it some more in a bit.

tbaxter-18f:
Amy, I'm seeing changes to pipfile.lock but not pipfile. Is that right?

Also, do we want to check in the deduped.json changes? I'm not sure.

amymok:
@tbaxter-18f 
> Amy, I'm seeing changes to pipfile.lock but not pipfile. Is that right?

That's correct.  I regenerated the Pipfile.lock locally by doing `pipenv install --python 3.6.5 --dev` at the base folder, and this time it did not create two different versions of Django as appears in Pipfile.lock right now.  So I left it as that, just change the Pipefile.lock.

> Also, do we want to check in the deduped.json changes? I'm not sure.

So I double check here.  It would be fine to leave it as-is for now with Django 2.0.x, but when we upgrade to >=2.1, it will create a problem because of this new changes on [model view permission](https://docs.djangoproject.com/en/2.1/releases/2.1/#model-view-permission).  Django will create an extra entry per model for view-only.  Before 2.1, there are only 3 entries created (add, change, delete). 
 If we upgrade, this will throw off all of the permission entries here with the ID which I also described [above](https://github.com/18F/tock/pull/843#issuecomment-434521393).

Those entries are unnecessary anyway because they already exist in the database as we `migrate` and were created by Django, so there's no need to duplicate them.  So it is probably better to get rid of them.  But if we wish to keep it there, we need to keep a note to revisit when we upgrade to 2.1.  Let me know what else may be the concern.

rogeruiz:
Thanks for that explanation @amymok! Good work on this. thank you!

Issue #680, opened 2017-10-13
Update reports to include FY18-specific reports


ccostino:
The reports page in Tock only allows you to pull all information or just a single FY specifically, but there is leakage between the reports. It appears they pass a single `after` param, so they can pick up reports from after the desired date. 

There is also a desire to add the user organization to the tables (both html and csv) as an additional datapoint.

*We'll know we're done when*
* FY reports only pull the desired FY
* user.organization shows in the output table
* Any related tests are written

ccostino:
This is done with #684 in terms of adding the reports for FY18, but more work needs to be done to fix all of the FY reports a bit more so that they only include the data for their respective fiscal years.

tbaxter-18f:
@ccostino Can you elaborate on what still needs to be done? Are we getting leakage or... ?

And I suspect we need to add FY19, right? @abrouilette is that a need yet or is it premature?

I'm wondering if we ought to just output these years automatically, as a range between 2017 and datetime.date.today.year or something (although that would capture calendar year, not fiscal so... :shrug:)

tbaxter-18f:
I think as we're rolling into 2019 this is going to become an issue again, and Amy has expressed an interest in handling it, so let's do it!

PR #847, opened 2018-11-07
Issue 680: Fix FY reports to only include specific years

amymok:
## Description

See Issue #680 

- [x] Added `before` parameter for `get_timecards` function to allow getting a list of timecards before and include a specific date.  This can now be used with `after` parameter to get a specific date range of timecards
- [x] Added corresponding tests for `before` parameter as well as using it in conjunction with `after parameter
- [x] Added `before` parameters to the current reports so it is filtered with a range
- [x] Iterate through fiscal years to create reports instead of hardcoding every single year
- [x] user.organization shows in the output table

I also added a few static methods to figure out the actual fiscal year start and end date based on the discussion I have with Jackie.
>During the week that spans two fiscal years, with the five work days, charging will happen to the year that has more days in it. So, if there are three or more September work days, the week will belong to the year September is in, else, if there are more October work days (three or more), this week will belong to the year October is in.

## Additional information

* Screenshot images
* Notes
  * ~~There seems to be an inconsistency with the date in the report for FY2016.  The superuser report starts with 10/1/2016, but other reports starts with 10/2/2016.  10/1/2016 is a Saturday so I am not sure if this is an exception or not for fiscal years.  I changed it back to FY starts on 10/1 and ends with 9/30 unless there are exceptions, please let me know.~~
  * I hardcoded "2017" as the first fiscal year for reporting because that was what's already there, if I should use something else, let me know.

### DIFF
--- a/tock/hours/models.py
+++ b/tock/hours/models.py
@@ -179,13 +179,13 @@ class ReportingPeriod(ValidateOnSaveMixin, models.Model):
             # this week belongs to this fiscal year
             # so the Sunday is the start of the fiscal year
             # i.e. if Tuesday, this will subtract 2 days to get the Sunday date
-            return first_date - datetime.timedelta(days=normalized_weekday)
+            return first_date - datetime.timedelta(days = normalized_weekday)
         else:
             # this week belongs to previous fiscal year
             # so the coming Sunday is the start of the fiscal year
             # i.e. if Thursday, this will add 7 - 4 days (3 days) to get the
             # Sunday date
-            return first_date + datetime.timedelta(days=7-normalized_weekday)
+            return first_date + datetime.timedelta(days = 7 - normalized_weekday)
 
     @staticmethod
     def get_fiscal_year_end_date(fiscal_year):
@@ -210,13 +210,13 @@ class ReportingPeriod(ValidateOnSaveMixin, models.Model):
             # so the Saturday before is the end of the fiscal year
             # i.e. if Tuesday, this will subtract 2+1 = 3 days to get the Saturday
             # date
-            return last_date - datetime.timedelta(days=normalized_weekday + 1)
+            return last_date - datetime.timedelta(days = normalized_weekday + 1)
         else:
             # this week belongs to this fiscal year
             # so the coming Saturday is the end of the fiscal year
             # i.e. if Thursday, this will add 6 - 4 days (2 days) to get the
             # Saturday date
-            return last_date + datetime.timedelta(days=6-normalized_weekday)
+            return last_date + datetime.timedelta(days = 6 - normalized_weekday)
 
     @staticmethod
     def get_fiscal_year_from_date(date):
--- a/tock/hours/models.py
+++ b/tock/hours/models.py
@@ -141,11 +141,91 @@ class ReportingPeriod(ValidateOnSaveMixin, models.Model):
         """Determines the Fiscal Year (Oct 1 - Sept 30) of a given
             ReportingPeriod. Oct, Nov, Dec are part of the *next* FY """
         next_calendar_year_months = [10, 11, 12]
-        if self.start_date.month in next_calendar_year_months:
+        # When this reporting period spans two fiscal years
+        if self.start_date.month == 9 and self.end_date.month == 10:
+            # if start date is 9/27 or earlier, it means there's more September
+            # than October in this period, so it belongs to FY of September
+            if self.start_date.day <= 27:
+                return self.start_date.year
+            # Else there's more October days, it belongs to the FY of October
+            else:
+                return self.start_date.year + 1
+
+        elif self.start_date.month in next_calendar_year_months:
             fiscal_year = self.start_date.year + 1
             return fiscal_year
         else:
             return self.start_date.year
+
+
+    @staticmethod
+    def get_fiscal_year_start_date(fiscal_year):
+        """
+        Normally the FY start date is 10/1 of the previous year, but since our
+        system makes use of reporting period, we cannot just easily use this
+        date.  So in this system, during the week that spans two fiscal years,
+        that week will belong to the previous year if there's more September
+        days than October. And more October days means it belongs to this year.
+        i.e. 10/1/2015 is on a Thursday, this means it belongs to FY 2015 since
+        there is more September days that week (Sun to Wed), so the start date
+        of FY 2016 is 10/04/2015 (that coming Sunday).
+        """
+        first_date = datetime.date(fiscal_year - 1, 10, 1)
+        # converting definition of weekday for easier calculation,
+        # so Sun = 0, Mon = 1, ... Sat = 6
+        normalized_weekday = (first_date.weekday() + 1) % 7
+        # if 10/1 is Sun=0, Mon=1, Tues=2, or Wed=3
+        if normalized_weekday in range(4):
+            # this week belongs to this fiscal year
+            # so the Sunday is the start of the fiscal year
+            # i.e. if Tuesday, this will subtract 2 days to get the Sunday date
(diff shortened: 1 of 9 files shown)


tock/hours/views.py
```diff
@@ -942,6 +943,14 @@ def get_queryset(self, queryset=None):
         sorted_fiscal_years = sorted(fiscal_years.items(), reverse=True)
         return sorted_fiscal_years
 
+    def get_context_data(self, **kwargs):
+        # Call the base implementation first to get a context
+        context = super().get_context_data(**kwargs)
+        today = localdate()
+        year = today.year + 2 if today.month >= 10 else today.year + 1
```
tbaxter-18f: this line could maybe use a comment explaining the logic behind it?

tock/tock/templates/hours/reports_list.html
```diff
@@ -12,28 +12,33 @@ <h2>Regular Reports</h2>
...
+			{% for year in fiscal_years %}
+				<a href="{% url 'reports:BulkTimecardList' %}?after{{ year|add:"-1" }}-10-01&before={{ year }}-09-30">FY{{ year }}</a>
+			{% endfor %}
 		</li>
 		<li>Complete timecard data with fewer fields:
 			<a href="{% url 'reports:SlimBulkTimecardList' %}">All</a>
-			<a href="{% url 'reports:SlimBulkTimecardList' %}?after=2016-10-02">FY2017</a>
-            <a href="{% url 'reports:SlimBulkTimecardList' %}?after=2017-10-01">FY2018</a>
+			{% for year in fiscal_years %}
+				<a href="{% url 'reports:SlimBulkTimecardList' %}?after{{ year|add:"-1" }}-10-01&before={{ year }}-09-30">FY{{ year }}</a>
```
tbaxter-18f: I wonder if it'd be cleaner to add `prior_year` into context as well, in the view, instead of calcing it multiple places in the template.

tbaxter-18f:
@amymok Is there anything outstanding you need to finish this up?

amymok:
@tbaxter-18f yes, I am waiting for a conversation with Jackie about the FY, because it will alter how that is being calculated, let me ping her again.

amymok:
Just had a conversation with @HuixianXu, here's the definition:

During the week that spans two fiscal years, with the five work days, charging will happen to the year that has more days in it.  So, if there are three or more September work days, the week will belong to the year September is in, else, if there are more October work days (three or more), this week will belong to the year October is in.

So I will use this to continue with the work.

tbaxter-18f:
Fantastic. Thanks for following up, Amy. Will you be sure to note that in the code so we have it documented for posterity? Thanks!

tock/hours/views.py
amymok: The logic has changed, but I tried to put as much doc as I can.

tock/tock/templates/hours/reports_list.html
amymok: The logic has changed as well, so this is no longer relevant.

tbaxter-18f:
Amy, I believe the last acceptance criteria was referring to https://github.com/18F/tock/blob/master/tock/tock/templates/hours/reporting_period_detail.html, which currently shows users who have and haven't tocked, with their name and email. The desire was to show their org, too, if I recall correctly. @abrouilette does that sound right to you?

tock/hours/models.py
```diff
@@ -138,15 +138,95 @@ def get_message_enabled_display(self):
...
+        that week will belong to the previous year if there's more September
+        days than October. And more October days means it belongs to this year.
+        i.e. 10/1/2015 is on a Thursday, this means it belongs to FY 2015 since
+        there is more September days that week (Sun to Wed), so the start date
+        of FY 2016 is 10/04/2015 (that coming Sunday).
+        """
+        first_date = datetime.date(fiscal_year - 1, 10, 1)
+        # converting definition of weekday for easier calculation,
+        # so Sun = 0, Mon = 1, ... Sat = 6
+        normalized_weekday = (first_date.weekday() + 1) % 7
```
tbaxter-18f: There's no need to change anything, and I'm not even sure I'm right, but I think there may be bits in the `calendar` module that may make some of this easier: https://docs.python.org/2/library/calendar.html

Only mentioning it in case it's helpful in the future :-)

tock/hours/views.py
```diff
@@ -943,6 +944,23 @@ def get_queryset(self, queryset=None):
...
+    def get_context_data(self, **kwargs):
+        # Call the base implementation first to get a context
+        context = super().get_context_data(**kwargs)
+        # Get today's date to figure out the fiscal year it is in
+        today = localdate()
+        # The year will be used to determine which year reporting to end with
+        year = ReportingPeriod.get_fiscal_year_from_date(today)
+        context['fiscal_years'] = []
+        for year in range(2017, year + 1):
+            year_dates = dict()
```
tbaxter-18f: I probably would have written all of this to the dict in one shot in order to (hopefully) be a bit more readable...
```
context['fiscal_years'].append({
     'year': year,
     'start_date': blahblahblah,
    'end_date': BobLoblaw
})
```
But it's fine the way you have it here too

tock/tock/templates/hours/reports_list.html
```diff
@@ -12,28 +12,33 @@ <h2>Regular Reports</h2>
 	<ul>
 		<li>Complete timecard data:
 			<a href="{% url 'reports:BulkTimecardList' %}">All</a>
-			<a href="{% url 'reports:BulkTimecardList' %}?after=2016-10-02">FY2017</a>
-            <a href="{% url 'reports:BulkTimecardList' %}?after=2017-10-01">FY2018</a>
+			{% for fy in fiscal_years %}
+				<a href="{% url 'reports:BulkTimecardList' %}?after={{ fy.start_date|date:'Y-m-d' }}&before={{ fy.end_date|date:'Y-m-d' }}">FY{{ fy.year }}</a>
```
tbaxter-18f: It's a minor thing, but inconsistency in our indentation hurts readability. Not asking for a change, just something to watch out for.

tbaxter-18f:
I think this looks really great, Amy. Thanks for really thinking it through. The only thing I want to be sure about before merging it is that organization in the table. Thanks!

tbaxter-18f:
Confirmed in Slack, it's that report html page. Thanks!

amymok:
> Confirmed in Slack, it's that report html page. Thanks!

Thanks I will take a look later today.

tock/hours/models.py
amymok: I didn't know about the calendar, thanks!  I think still some work using the calendar method.  I will leave this as-is.

tock/hours/views.py
amymok: This is actually a really confusing piece of code now that you pointed out.  I also was using `year` as each instance in the for loop as well as a variable for the end of range.  I am rewriting it now.

amymok:
@tbaxter-18f This is ready for review again.  Thanks!

tock/hours/views.py
```diff
@@ -943,6 +944,23 @@ def get_queryset(self, queryset=None):
...
+        # Get today's date to figure out the fiscal year it is in
+        today = localdate()
+        # The year will be used to determine which year reporting to end with
+        year = ReportingPeriod.get_fiscal_year_from_date(today)
+        context['fiscal_years'] = [
+            {
+                'year': fy,
+                'start_date': ReportingPeriod.get_fiscal_year_start_date(fy),
+                'end_date': ReportingPeriod.get_fiscal_year_end_date(fy)
+            } for fy in range(2017, year + 1)
```
tbaxter-18f: This is interesting how you're building out the list of dicts in a list comprehension. I wouldn't have thought of that, but it's very readable this way!

PR #850, opened 2018-11-14
Round Total Hours sum to 2 decimal precision.

### LINKED ISSUE #848: Calculating sums for decimal hours isn’t correct.
When calculating the total hours tracked for a week, the Total Hours calculation exposes a classic JavaScript floating math bug:

`0.2 + 0.2 + 0.2 = 0.6000000000000001`

Screenshot of Tock, showing three entries with 0.2 entered, and a sum of 0.6000000000000001

@tbaxter-18f advises starting [in the frontend code](https://github.com/18F/tock/tree/master/tock/tock/static/js) to investigate. 

Likely the fix will involve working around using floating numbers in JavaScript, which are notoriously imprecise.

adunkman:
Fixes #848.

We shouldn’t need anything more than 2 digits of decimal precision to calculate Total Hours. This fixes common floating point precision issues in JavaScript, most easily seen by adding `0.2 + 0.2 + 0.2` (which should result in `0.6`, as shown below, but results in `0.6000000000000001` without rounding.

Screenshot showing 0.2 added three times results in 0.6

This is my first time contributing to a python project, so I’m unsure of how to write a test to cover this — let me know how, and I’d be more than happy to add (pun intended) one!

### DIFF
--- a/tock/tock/templates/hours/timecard_form.html
+++ b/tock/tock/templates/hours/timecard_form.html
@@ -164,7 +164,7 @@
 
       if (!is_checked) hoursTotal += ($(this).val() / 1);
     });
-    return {'hoursTotal': hoursTotal};
+    return {'hoursTotal': parseFloat(hoursTotal.toFixed(2))};
   }
 
   function populateHourTotals(){


rogeruiz:
Thanks @adunkman. Writing tests for the JS in this project would require that the JavaScript be extracted outside of the HTML templates and loaded on the page as a file. Then we could have unit tests test this JS code separate of any python code. 

Most of the JS in this project has been added quickly and to HTML templates. We don’t have an asset management system for external JS in the project. But this would require a refactor of most of the HTML pages among other things. Since Tock isn’t a billable project, this is probably just a good idea that may never get implemented.

Issue #862, opened 2019-01-08
Remove operations dashboard reports


tbaxter-18f:
Currently the operations dashboard reports do not work and are not used. They can be removed.

**We'll know we're done when**
* Nav item is removed and extra space distributed
* URLs removed
* Views removed
* templates removed
* tests removed

PR #863, opened 2019-01-08
Remove the Operations Dashboard

Jkrzy:
## Description

Per #862, The operations dashboard is no longer of use. Removing all associated code, tests, and models.


tock/hours/models.py
```diff
@@ -24,51 +24,6 @@ class Meta:
         unique_together = ['project', 'hours_per_period']
         ordering = ['project__name']
 
-class Targets(models.Model):
```
tbaxter-18f: An excellent argument for why models should always be commented. Congrats on figuring out this was the relevant model

Jkrzy:
Wanted to take a second pass through looking for loose ends. Ready for review.

Issue #881, opened 2019-01-14
Enforce project line submission at timecard submission


tbaxter-18f:
In some cases, somehow, when a user submits a timecard the project lines will not be marked as submitted. This will present to the user as an undefined error, and can be seen in the admin by the project lines having red "unsubmitted" checks when the timecard itself has been entered.

We need to determine why the project lines have their own "submitted' independent of the timecard "submitted'. If they're not needed, remove them. If they are, be sure that they're in sync and don't throw an error.

Jkrzy:
I've been unable to replicate inconsistency between `Timecard.submitted` and TimeCardObject.submitted`. 

However I don't believe a discrepancy there would prevent submission of a Timecard.

`TimeCardObject.submitted` is functionally a copy of `Timecard.submitted` and any instance where it is evaluated should be replaced with `timecard__submitted`. It looks like the only place its used at the moment is in the utilization reports.

tbaxter-18f:
Yes, timecard.submitted is definitely used. You can see the inconsistency if you force an error on a tock submission  -- apparently, judging from Slack conversations, leaving out required notes is the easiest way -- then view the submission in the Tock admin. You'll get a "submitted" timecard with unsubmitted timecardobjects.

That said, I understand you'd prefer to pull that work into a separate issue, which I'm totally fine with.

tbaxter-18f:
@Jkrzy is this resolved with the merge of #886 above?

Jkrzy:
No, thanks for the reminder @tbaxter-18f  that didn't address an assessment/refactoring of `TimeCardObject.submitted`. I'll follow-up later this week with a plan to close this issue.

PR #886, opened 2019-01-23
Allow users to swap project IDs across timecard entries

Jkrzy:
Recreating the `TimeCardObject` unique constraint as deferrable and deferring when saving TimeCardObject formsets thereby allowing users to swap project IDs between
existing entries on a timecard.

## Description

Prior to this PR, attempts to swap project IDs between existing entries, either intentional or inadvertent, prevented users from submitting their timecard without providing any useful error message or pathway to resolve.


tock/hours/forms.py
```diff
@@ -261,6 +261,17 @@ def clean(self):
 
         return getattr(self, 'cleaned_data', None)
 
+    def save(self, commit=True):
+        """
+        Save with deferred constraints
+        Allowing users to swap project IDs between TimeCardObjects
+        """
```
tbaxter-18f: Since this gets a bit down into the gears of things, can we either write up more on what we're doing or is there a link we can offer that gets into the details?

tbaxter-18f:
This resolves #881 right?

tock/hours/models.py
```diff
@@ -400,6 +400,21 @@ def save(self, *args, **kwargs):
         super(TimecardObject, self).save(*args, **kwargs)
```
tbaxter-18f: I still think we should remove L376 from the save() and `submitted` from the model while we're doing this. They're not doing any good, right?

Jkrzy:
Thanks @tbaxter-18f, We'll address #881 separately, this PR doesn't include any modification/assessment of `TimeCardOjbect.submitted`

Jkrzy:
Working towards a resolution of #887 with this PR.

tock/hours/forms.py
Jkrzy: Wrote this up more comprehensively in #887

tbaxter-18f: Thats great, but I'm thinking of future person reading forms.py who's confused by what this is and why we did it. I want to leave them a bit more context here (or link to more context)

Jkrzy: Gotcha! Added details clarifying the custom save behavior here.

tbaxter-18f: This looks great. Thank you!

PR #894, opened 2019-03-18
Only recycle based on tags instead of all branches

rogeruiz:
## Description

We're having an issue with #772 where it's consuming too many containers because it gets stuck. This tries to limit recycling production instances based on changes to the Tock release tag. So instead of running on all branches, it'll only target one tag at a time.

## Additional information

- n/a

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -182,5 +182,7 @@ workflows:
     - schedule:
         cron: "0 10 * * *"  # Roughly 5am ET
         filters:
+          tags:
+            only: /v20[1-9][0-9][0-9]+\.[0-9]+/
           branches:
-            only: /.*/
+            ignore: /.*/


.circleci/config.yml
```diff
@@ -182,5 +182,7 @@ workflows:
     - schedule:
         cron: "0 10 * * *"  # Roughly 5am ET
         filters:
+          tags:
+            only: /v20[1-9][0-9][0-9]+\.[0-9]+/
           branches:
-            only: /.*/
+            ignore: /.*/
```
tbaxter-18f: This looks pretty straightforward and makes sense, but I do wonder, would it make more sense to not fool with the tag and just use the `master` branch?  Thoughts?

rogeruiz: I think that may put us in the same situation as before. Essentially, we want to only trigger on one thing. Not multiple commits to that same thing. This stanza attaches a cron job to whatever branch/tag we point it to. So two commits to master in a 24h period will cause two recycle jobs to run. I think it being pointed just to a tag should be good unless of course we release multiple times in 24h, then lol.

tbaxter-18f: Makes sense. Thanks!

Issue #896, opened 2019-05-14
Update CF stack to `cflinuxfs3` and other dependencies


Jkrzy:
We need to migrate production and staging instances to `stack: cflinuxfs3`

Cloud.gov is dropping support for `cflinuxfs2` per https://cloud.gov/updates/2019-05-13-cflinuxfs2-final-warning/

### We'll know we're done when
* Staging and production manifests have been updated to `stack: cflinuxfs3`
* Any unforeseen compatibility issues are resolved

PR #897, opened 2019-05-14
Dependency and cloud.gov stack updates

Jkrzy:
## Description

Updating stack per to `cflinuxfs3` per #896. 

Also using this opportunity to update dependencies which includes a new version of Bandit with additional checks. Some of these new checks are false-positives and have been ignored using `# nosec` in-line where appropriate.

Also updating `.circleci` config here, tags filter no longer support in scheduler trigger and address unexpected behavior we're seeing  where `recycle_prod` is being triggered from multiple sources/branches.


amymok:
It all makes sense, and LGTM!

Issue #902, opened 2019-06-20
Upgrade Django version


Jkrzy:
Lets upgrade Tock to the latest Django LTS version, 2.2.2.

Jkrzy:
Depends on: https://github.com/18F/cg-django-uaa/issues/48

PR #903, opened 2019-06-25
Updating Django and other dependencies

Jkrzy:
## Description

For #902, we're jumping up to the latest LTS version of Django. Taking this opportunity to bump up our other python and js dependencies as well.


Pipfile
```diff
@@ -14,7 +14,7 @@ djangorestframework-csv = "*"
 bleach = "*"
 newrelic = "*"
 cfenv = "*"
-cg-django-uaa = "*"
+cg-django-uaa = {git = "https://github.com/18F/cg-django-uaa",ref = "v2.0.0"}
```
tbaxter-18f: What do you think about logging an issue to ourselves to update this when cg-django-uaa gets a proper release?

Issue #905, opened 2019-06-26
Update vendored javascript dependencies and documentation


Jkrzy:
Lets update the version of Jquery being served by Django and update the documentation to reflect that we're vendoring the front end dependencies when deployed.

PR #906, opened 2019-06-26
Update jquery

Jkrzy:
## Description

Per #905, Updating our vendored version of `jquery.min.js` to the latest version. Also updating documentation to reflect our use of NPM alongside vendored dependencies.


tock/tock/templates/projects/project_detail.html
```diff
@@ -96,19 +96,4 @@ <h3 class="chart__title">Latest Periods with Hours Logged</h3>
...
 
-<link rel="stylesheet" href="{% static 'css/vendor/jquery.dataTables.min.css' %}" />
-<script src="{% static 'js/vendor/jquery.dataTables.min.js' %}"></script>
-<script src="{% static 'js/vendor/dataTables.fixedColumns.min.js' %}"></script>
-<script type="text/javascript">
-$(document).ready(function() {
-  var $table = $('#dataTable').dataTable({
-    scrollX: true,
-    paging: false,
-    lengthChange: false
```
Jkrzy: This line contained a syntax error which has prevented datatables from running. It doesn't look like this has worked for quite some time as resolving the syntax breaks styling on the page. Herein we remove in favor of fixing.

Jkrzy:
Hey @amymok, just pushed another update here for latest dependenecies as well as a general clean-up of our vendored things being served out of `static/`. Please take a look when you have a moment. Thanks!

An additional note on the removal of datatables.js here: https://github.com/18F/tock/pull/906/commits/dcebc5b9adc679f5daac32d4564933296c597c2d#r312072292

tock/tock/templates/projects/project_detail.html
tbaxter-18f: maybe this is an opportunity to remove the datatable dependency altogether?

I'd love it if we could get to a point where we didn't have a jquery dependency, either. This might be one step in that direction?

Jkrzy: We are 😄, that code snippet and the source files associated with datatables are removed w/ this PR.

Issue #926, opened 2019-11-25
Admin Tab for newly added admins


saraheckert:
As a newly added admin, I have the rights to update various features, but the admin button does not appear in my main tock menu.  I am able to use a workaround by copying the admin URL.  Thanks!
<img width="840" alt="Screen Shot 2019-11-25 at 8 53 53 AM" src="https://user-images.githubusercontent.com/42154564/69556035-53c6c000-0f61-11ea-9e07-e01d72e0ccf6.png">

tbaxter-18f:
Thanks for logging this @saraheckert 

For anyone reading, in some very cursory looking, I think this is related to the permissions check we do to build the nav. We're checking for `is_superuser` but I think we maybe/probably should be checking for `is_staff` and then sort it out with permissions, since `is_staff` is usually the determining factor for getting into the admin. 

Totally open to other options. At any rate, see https://github.com/18F/tock/blob/master/tock/tock/templates/_navigation.html#L33

Jkrzy:
This one is a bit of a rabbit hole...since I don't think its clearly defined anywhere when someone should be a superuser as opposed to just having admin access and fine grained permissions. Though there's definitely a divide in functionality and a handful of users where `is_staff and not is_superuser`.

For now, let's provide a nav option for _everyone_ that does have admin access.

tbaxter-18f:
My opinions, for whatever they're worth, based on my understanding of general django concepts:

1. No one should be a superuser except Tock core devs
2. is_staff is conventionally the expected pathway to the admin

Not saying that's how we have to do it that way, just clarifying the expectations I'm coming to the conversation with.

Jkrzy:
Agreed with that being the desired state, will take some audit/documenting/role updating to get there and this effort should also close out #873.

Jkrzy:
Nav provided in #930, will continue audit/documentation/updating work in #873

Issue #868, opened 2019-01-09
Update nav to differentiate user and administrator actions


tbaxter-18f:
To make the Tock nav more logical, we'd like to differentiate between user actions (reporting time, finding the override form, logging out) and reports and admin actions that users are generally not so interested in. 

Discussion took place in this google doc:  https://docs.google.com/document/d/1lWFJggjjp7Ue2FbJGaxswGiXhuD4f6tUq37NqfJOEWE/edit?ts=5c353be2#

Where we landed on moving the administrative actions to the upper right, as the attached (very rough) mockup shows:

![screen shot 2019-01-08 at 4 39 04 pm](https://user-images.githubusercontent.com/32077682/50902598-efe6e780-13e0-11e9-8a55-a423236a14e6.png)

We should also add a link for “file time report” -- currently, if you follow another nav item you have to click on the 18f logo to get back to the ability to file your timesheet. This doesn’t make a ton of sense since the app doesn’t really have a homepage per se; it has different sets of actions available in different places. All those actions should be enumerated on every page.

Note that mockup is just a mockup and should not necessarily be construed to represent the desired final state.

Note that there are some Nav items that aren't visible to all users, and we'll need to be sure to account for those.

**We'll know we're done when**
* "File time report" link is added
* Navigation is modified similar to depiction

tbaxter-18f:
It may make a lot of sense to tackle this and #867 in parallel

tbaxter-18f:
Also, if we've got some writers with some spare time on their hands, getting a bit clearer and more consistent with the language in these nav links might be nice.

jenniferthibault:
@tbaxter-18f what are the nav items under Admin that aren't visible to all users? I took a crack at this with some content designers, but missed that one

tbaxter-18f:
Assuming I have full permissions (and I think I do), I have 

* A link into the site admin
* Create reporting period
* Bulk timecard CSV (may be available elsewhere, I'm not sure)

There's also a "supervisor" tab with a single item under it to "utilization reports"

I think the other links are all public.

Personal opinion: I don't think we have enough links that we necessarily need to put any of them behind dropdowns.

jenniferthibault:
Here's an initial direction that was even ✨blessed✨ by a few of our NYC-based content designers in an informal feedback session. The goal was to use matching format & point of view:

**Nav item microcopy**
[...]

PR #930, opened 2019-11-25
Render link to admin for those w access

Jkrzy:
## Description

Working towards resolving #926.

Proposed first step, let's make sure that _everyone_ who currently has `Admin` access it able to navigate there from Tock.

Consolidating navigation items which require elevated privileges into a single nav item, `Administrative`,  wherein links to all functionality accessible by the current user are rendered.

### is_staff AND is_superuser
Screen Shot 2019-11-25 at 3 43 40 PM


Jkrzy:
Whoops, we need to incorporate the good works of https://github.com/18F/tock/issues/868 here.

Issue #864, opened 2019-01-08
Update reports page for FY19


tbaxter-18f:
On the reports page, we show FY17 and FY18, we can remove those, and need to add FY19

https://tock.18f.gov/reports/

**We'll know we're done when**
* FY17 and FY18 are removed
* FY19 is added

amymok:
https://github.com/18F/tock/pull/847 should have added the FY19

tbaxter-18f:
Thanks Amy! I forgot about that. So we can just remove the older years, which in turn means this is less urgent, and if we don't do it before our next deploy it's no big deal.

Jkrzy:
@tbaxter-18f, can you re-confirm which FY should be available on the reports page?

Is it FY20  and FY19 now?

tbaxter-18f:
I'm gonna go ahead and say yes, and if at some future date @saraheckert or anyone else wants to see historical data we can pull it manually or add it to the page.

PR #932, opened 2019-12-03
Limit FYs displayed on reports page

Jkrzy:
## Description
For #864, add setting, `STARTING_FY_FOR_REPORTS_PAGE` allowing
for configuration of which reports are directly linked from reports page.

In this initial configuration we restrict the linked reports to FY19 and beyond.

Custom/other reports are still directly accessible via the API and query params.

e.g. Need complete timecard data from FY18?

> https://tock.18f.gov/reports/timecards_bulk.csv?after=2017-10-01&before=2018-09-29

The above will still function and generate the expected data, but its no longer rendered as direct link on the reports page.


tock/tock/settings/base.py
```diff
@@ -134,4 +134,5 @@
 SECURE_HSTS_INCLUDE_SUBDOMAINS = True
 SECURE_HSTS_PRELOAD = True
 
-RECENT_TOCKS_TO_REPORT = 5
\ No newline at end of file
+RECENT_TOCKS_TO_REPORT = 5
+STARTING_FY_FOR_REPORTS_PAGE = 2019
```
tbaxter-18f: Wondering if it would make sense to set the year dynamically from datetime.date. It wouldn't align with FY, but it would just overlap for few months and then update automatically. Not necessary, just thinking out loud.

Jkrzy: Good idea, that I think could extend to abstracting out the Fiscal Year math from `ReportingPeriod`.

Will open a new issue for that effort

PR #933, opened 2019-12-04
Remove TimeCardObject.submitted

Jkrzy:
## Description

Resolves: #881 by removing `TimeCardObject.submitted` and replacing references in queries/tests to the associated TimeCard object's `submitted` field.


requirements.txt
```diff
@@ -1,22 +1,22 @@
...
 bleach==3.1.0
-certifi==2019.9.11
+certifi==2019.11.28
 cfenv==0.5.3
 cg-django-uaa==2.0.0
 chardet==3.0.4
 dj-database-url==0.5.0
 django-webtest==1.9.7
-django==2.2.7
+django==2.2.8
```
cantsin: Seems like https://github.com/18F/tock/pull/935 is in this PR as well? I'm curious as to if these updates are necessary here.

tock/projects/tests.py
```diff
@@ -494,13 +494,11 @@ def test_total_hours_billed(self):
         TimecardObject.objects.create(
             timecard = Timecard.objects.get(pk=1),
             project=Project.objects.get(pk=1),
-            submitted = True,
             hours_spent= 10
         )
         TimecardObject.objects.create(
             timecard = Timecard.objects.get(pk=2),
             project=Project.objects.get(pk=1),
-            submitted = False,
```
cantsin: I don't understand why these `submitted` values were set in the first place, as far as I can tell, this doesn't impact the test in any way. (To be clear, not sure why one is True, and the other is False)

Jkrzy: Me either, I suspect a well intentioned but abandoned functionality path

PR #940, opened 2019-12-19
Calculate and display details of billable hour expectations for the week

Jkrzy:
## Description

Working towards resolution of #880 

1. For each timecard calculate and show progress towards the expected billable hours for the week.

## Additional information

This PR gets us to a point where the information needed to calculate progress towards expected hours to be billed for each time card.

Three values are passed to the `timecard_form.html` template.
```javascript
 const excludedFromBillability = {{excluded_from_billability|safe }};
 const billableExpectation = {{billable_expectation|safe }};
 const nonBillableProjects = {{non_billable_projects|safe}};

```

Which will allow the frontend to resolve:

The target billable hours for the week as:
`( total_hours_billed - hours_billed_to_excluded_projects) * billable_expectation)`

The total billable hours currently entered as:
`Sum of hours tocked to projects with IDs (value of selected option) NOT in nonBillableProjects`

PR #955, opened 2020-01-10
Add billable hours indicator to timecard.

### LINKED ISSUE #949: Render billable time expectations on timecard form
Let's bring together the work of https://github.com/18F/design-lab/issues/52, Issue #880, and PR #940 to enhance the Timecard submission form to communicate expectations and progress towards billable time.

https://github.com/18F/design-lab/issues/52 contains potential visual representations of this information.

PR #940 makes available, in javascript, the data necessary to calculate target billable hours.

## We'll know we're done when

- [ ] Users are presented with the information necessary to understand how many hours they're expected to Tock to `billable` projects for a given week and how that number is calculated.
- [ ] Progress towards the target billable hours is updated as tock lines are completed on the form.
- [ ] Users are directed to the #tock and the [handbook](https://handbook.18f.gov/tock/#weekly-billable-hour-expectations) if they have questions re: target billable hours.

adunkman:
Closes #949, includes #940. Best reviewed when [ignoring whitespace](https://github.com/18F/tock/pull/955/files?w=1).

This PR implements two circular graphs to replace to **Total hours** section of the timecard, one showing total hours and the other showing billable hours. 

Billable hours uses the implementation in #940 to determine what tock entries are billable, non-billable, or excluded from calculation (out of office). 

<details>
<summary>Expand to see how I got to that design, or see the final product below.</summary>
<br>

I originally started with the design proposed in https://github.com/18F/design-lab/issues/52, but I kept coming back to how complicated the visualization ends up, due to the different data items a single bar was showing: 

- What number of hours are expected for this timecard? (40)
- What number of billable hours are expected for this timecard? (Maybe 32)
- Have I entered the correct number of hours for this timecard? (Used to be shown with red/green on the Total Hours counter)
- Have I entered enough billable hours for this timecard? 
- How many billable hours have I entered?
- How many total hours have I entered? 

I started with adaptations of this using a bar graph, and a simplification of it:

|Original|Simplified|
|---|---|
|image | image_1

It ended up being a lot of data in a very small space though, and it seemed too complicated. I circled around the idea that although the numbers were changing, the _percentages_ are what are important — but humans are thinking in number of hours, so the percentages themselves as numbers aren’t super valuable. 

Focusing in on “what does it take to get to 100%" allowed me to simplify the graphic quite a bit — and resulted in the approach below. 

(I toyed around with even simpler ideas, like making a smiley face happier, but it didn’t convey enough information to be useful, despite it being charming).

image_3

</details>

Billable hours and total hours graphs

As requested in https://github.com/18F/design-lab/issues/52, this does not require a visualization library — it’s simple SVG stroke manipulation. Additionally, the **Billable hours** header links to [the associated section of the Handbook](https://handbook.18f.gov/tock/#weekly-billable-hour-expectations) describing billable hours expectations.


tock/employees/models.py
```diff
@@ -105,6 +103,9 @@ class UserData(models.Model):
     is_18f_employee = models.BooleanField(default=True, verbose_name='Is 18F Employee')
     is_billable = models.BooleanField(default=True, verbose_name="Is 18F Billable Employee")
     unit = models.IntegerField(null=True, choices=UNIT_CHOICES, verbose_name="Select 18F unit", blank=True)
+    billable_expectation = models.DecimalField(validators=[MaxValueValidator(limit_value=1)],
+                                              default=0.80, decimal_places=2, max_digits=3,
+                                              verbose_name="Percentage of hours which are expected to be billable each week")
```
tbaxter-18f: Is it a percentage, or a number? I had _thought_ a number?

tbaxter-18f: Maybe say "Percentage of hours (expressed as a decimal) expected to be billable"?

tock/hours/forms.py
```diff
@@ -36,7 +36,7 @@ class TimecardForm(forms.ModelForm):
 
     class Meta:
         model = Timecard
-        exclude = ['time_spent', 'reporting_period', 'user']
+        exclude = ['time_spent', 'reporting_period', 'user', 'billable_expectation']
```
tbaxter-18f: This is out of scope for this review, but we should change this to explicitly include rather than include, per Django recommendations.

tock/hours/views.py
```diff
@@ -504,6 +505,9 @@ def get_context_data(self, **kwargs):
             'timecard_notes': TimecardNote.objects.enabled(),
             'unsubmitted': not self.object.submitted,
             'reporting_period': reporting_period,
+            'excluded_from_billability': json.dumps(list(Project.objects.excluded_from_billability().values_list('id', flat=True))),
```
tbaxter-18f: Really minor, but I don't think you should need to convert the values_list to list. Given flat=True, I think it should be a list already.

tock/projects/models.py
```diff
@@ -185,6 +185,12 @@ def active(self):
     def inactive(self):
         return self.get_queryset().filter(active=False)
 
+    def excluded_from_billability(self):
```
tbaxter-18f: What's the distinction between excluded from billability and non-billable? Is it time to add comments to the methods?

tock/projects/models.py
```diff
@@ -185,6 +185,12 @@ def active(self):
...
+        return self.get_queryset().filter(exclude_from_billability=True)
+
+    def non_billable(self):
+        return self.get_queryset().filter(exclude_from_billability=False, accounting_code__billable=False)
```
tbaxter-18f: I wonder -- not advocating, just wondering -- if we should add some managers for these things.

tock/tock/static/sass/_entries.scss
```diff
@@ -154,26 +154,58 @@
...
+  }
+
+  .circular-graph {
+    transform: rotate(-90deg);
+    position: absolute;
+    left: 0;
+
+    .background {
+      fill: none;
+      stroke: $color-primary-alt-lightest;
```
tbaxter-18f: I think this is a really clever CSS-only solution to transform the decimal. Nice!

tock/tock/templates/hours/timecard_form.html
```diff
@@ -174,15 +232,34 @@ <h3 class="usa-alert-heading">{{ timecard_note.title }}</h3>
...
+
+    if (totals.billableHours === 0) {
+      billableElement.removeClass('valid invalid');
+    }
+    else if (totals.billableHours >= totals.billableHoursTarget) {
+      billableElement.addClass('valid').removeClass('invalid');
     }
     else {
-      $('.entries-total-reported-amount').removeClass('invalid');
+      billableElement.addClass('invalid').removeClass('valid');
```
tbaxter-18f: consider maybe toggleClass?

tbaxter-18f:
One additional thought: can we add some indication of the period this is for? Eventually, we'd like to show if you're on target for the current time period, for some larger time period (maybe 3 months) and for the year. I understand that may not be doable right now, but I do think it would help to say what period the graph is representing

igorkorenfeld:
Great work! I like how you broke this down into simple, digestible visualization. It did take me a little bit to figure out why the billable hours circle gets filled and becomes green when you add the 8 hours of OOO, since you there is no indication of the scale for the circles. I am curious to see if other understand this more easily. There are just a couple of picky little visual design things that could be adjusted, and perhaps we can pair on those.

tbaxter-18f:
@adunkman can we pick this up again?

adunkman:
@igorkorenfeld would be happy to pair! Would you prefer that to happen before this PR is merged, or would you be okay to merge this first and follow-up with a tweaks PR as timing permits?

igorkorenfeld:
@adunkman totally OK with having the adjustments happen in a follow up PR

tock/hours/forms.py
amymok: Adding the recommendation reference here: https://docs.djangoproject.com/en/3.0/topics/forms/modelforms/#selecting-the-fields-to-use

@tbaxter-18f Since this is out of scope, I recommend adding a new issue to be resolved later.  Does that work?  If so, I can add the issue.

tock/hours/views.py
amymok: `values_list` does not return a `list`, it returns a `ValuesListQuerySet`.  `flat=True`  only means to return results as single values rather than one-tuples.
https://docs.djangoproject.com/en/3.0/ref/models/querysets/#values-list

tock/hours/forms.py
tbaxter-18f: Works great for me! Thank you!

tock/hours/views.py
tbaxter-18f: I still think `json.dumps(ProjectProject.objects.excluded_from_billability().values_list('id', flat=True))` should be fine, but it's so minor it's not even worth the time I've spent typing this. My point was that when using `flat=True` it's list-like enough to not need to cast to `list` but really, it's not important.

tock/projects/models.py
tbaxter-18f: @amymok I think if this is a really simple clarification to deal with here, let's go ahead, but if it takes any real digging, let's lot a separate issue to better document and clarify it.

amymok:
I have a co-working session with @adunkman today.  I have a few things I am looking at right now before merging this in.  Will let you know once it is ready for re-review.  Thanks.

tock/tock/templates/hours/timecard_form.html
tbaxter-18f: Just a suggestion. I'm totally OK with  leaving it as-is.

tock/hours/forms.py
amymok: https://github.com/18F/tock/issues/973

tock/projects/models.py
amymok: Can you elaborate more here?  Probably will work on it separately from this PR.

amymok:
@tbaxter-18f We have made a few updates, can you take a quick look to make sure they address your comments?

tock/projects/models.py
tbaxter-18f: Just wondering if it would be useful to have `Projects.non_billable.all()` and such. Not looking for any change, just thinking about it.

tock/projects/models.py
```diff
@@ -219,7 +231,10 @@ class Project(models.Model):
     )
     project_lead = models.ForeignKey(User, null=True, on_delete=models.CASCADE)
     organization = models.ForeignKey(Organization, blank=True, null=True, on_delete=models.CASCADE)
-
+    exclude_from_billability = models.BooleanField(
```
tbaxter-18f: I want to clarify this a little more. I *think* anything non-billable should be excluded. @saraheckert does that seem right to you?

If that's the case, then I'm not sure I understand the need =for the new field and check.

@amymok can you help explain it?

amymok: I think this should probably read, `check if this project should be excluded from calculations of total hours`.  Because this check is really to make sure we accurately calculate the number of nonbillable hours for a reporting period.

adunkman: From the original #940, I think this explains what "excluded" was intended to mean:

> The target billable hours for the week as:
>
> `(total_hours_billed - hours_billed_to_excluded_projects) * billable_expectation)`

So "excluded" hours is the category to which Out of Office lives in.

amymok:
@tbaxter-18f I plan to merge this PR in at the end of the day Thursday 2/20.  I do not think there's any outstanding conversations we need to resolve here.  We can also open new issues after this is merged as well.

PR #966, opened 2020-02-07
Moved utilization reports

tbaxter-18f:
Moves utilization reports from behind a menu only superusers can see and into a location where they can be viewed by anyone. Transparency!


amymok:
@tbaxter-18f Can you update the PR description above: https://github.com/18F/tock/pull/966#issue-372537668

tock/tock/templates/_navigation.html
```diff
@@ -10,11 +10,24 @@
           <span>Projects</span>
         </a>
       </li>
-      <li>
-        <a href="{% url 'reports:ListReports' %}" class="usa-nav-link">
-          <span>Reports</span>
-        </a>
-      </li>
+      {% if user.is_staff %}
```
amymok: This should be `request.user.is_staff`

tock/tock/templates/_navigation.html
```diff
@@ -10,11 +10,24 @@
...
+        <li>
```
amymok: This doesn't render for me.  No menu actually shows up.

tock/tock/templates/_navigation.html
tbaxter-18f: Huh. I thought we passed user into context, but it appears you're right!

tock/tock/templates/_navigation.html
tbaxter-18f: We may need to be sure you're set up as superuser and staff in the admin.

tock/tock/templates/_navigation.html
```diff
@@ -10,11 +10,24 @@
...
       </li>
-      <li>
-        <a href="{% url 'reports:ListReports' %}" class="usa-nav-link">
-          <span>Reports</span>
-        </a>
-      </li>
+      {% if request.user.is_staff %}
+        <li>
+          <button class="
+            usa-accordion-button usa-nav-link" aria-expanded="false" aria-controls="side-nav-2">
```
amymok: The `side-nav-2` is conflicting with the one under Administrative, this is causing the submenu not rendering.  So choose something else here.

tock/tock/templates/_navigation.html
```diff
@@ -10,11 +10,24 @@
...
+            <span>Reports</span>
+          </button>
+          <ul id="side-nav-2" class="usa-nav-submenu" aria-hidden="true">
```
amymok: Same comment as above for the `side-nav-2`.  Choose another id.

tbaxter-18f: Good catch, Amy. Those are also bad IDs, so I've made them a bit more descriptive.

amymok: @tbaxter-18f Did you push the commit?

tbaxter-18f: I did now :-)

amymok:
There's no changes in dependency, Snyk is being funky here for pending.  Will bypass and merge.

Issue #968, opened 2020-02-11
Filter reports by division (18F, COE, Login.gov, etc)


allisonnorman:
From @saraheckert: As Chief of Projects, I want to view only users who have not submitted timecards **within my division** (e.g. 18F) so that I can follow up with them to tock. 

Steps to pull this information today:
- Select Reports by Weekly Reporting Period and select relevant week
- See Users who have not filed a timecard -> All divisions are present, including COE, Login, etc.
- Visually filter out team members from other divisions (this is the problem)

![Screen Shot 2020-02-11 at 3 10 48 PM](https://user-images.githubusercontent.com/6900926/74288327-fb41cc00-4ce0-11ea-957e-f8d6eb8f5c5f.png)

## We'll know we're done when
- [x] Users have the ability to view timecards by division (18F, COE, Login.gov) on the Weekly reporting period reports (https://tock.18f.gov/reports/)

saraheckert:
This looks great, thank you

PR #974, opened 2020-02-14
First take at sorting reporting by org.

tbaxter-18f:
## Description

First take on #968, attempting to sort by org name. Because there's a fairly gnarly relationship out to org, I'm kinda hesitant to do more sophisticated reporting.

This is hard to test as you have to find a reporting period:  http://localhost:8000/reports/2015-08-16/ 
And you also have to apply org data to multiple users, which don't currently have it. If you can pick it up and fully test it, I'd appreciate it.


Jkrzy:
Thanks @tbaxter-18f!! I'll pick this one up!

Jkrzy:
Tiny step in the right direction for #968 ahead of adding inputs and true sorting/filtering UX interactivity. 

We're adding Organization name to the sorting for all displays on the `Reports` page. With this change, each table is now sorted by Organization Name, Last Name, and then First Name.

`18F` will render as the first organization in the table.

tbaxter-18f:
LGTM, but since I started it I'll wait for @amymok to approve. Thanks for picking it up, Joe!

amymok:
There's no changes in dependency, Snyk is being funky here for pending.  Will bypass and merge.

PR #984, opened 2020-03-03
Added some migrations

tbaxter-18f:
## Description
Missing migrations from prior code commits


tbaxter-18f:
Weird behavior I'm seeing locally... it's trying to create a new migration for hours to show the TimecardObject unique_together, but that was added two years ago. 

And then, even more puzzlingly, it claims there's a violation of the constraint in the database: https://gist.github.com/tbaxter-18f/a3ec20b31a97b9110313cf3170961073

I've deleted that migration rather than merging it in, but I can't explain the behavior, and it does trouble me.

Jkrzy:
Interesting, let's pair tomorrow or later this week and take a look together.

tbaxter-18f:
Came at this morning with fresh eyes, blew away everything in Docker, started from scratch and got the last migration working. So this one is ready for review.

Jkrzy:
The fix required here was a bit more nuanced than expected. Passing back to you for a review of the new commit @tbaxter-18f.

tbaxter-18f:
Looks good to me!

Issue #981, opened 2020-02-27
Update utilization reports by business unit


Jkrzy:
For #980 

> As an 18F tock user, For each business unit I want to be able to see accurate and complete summaries of billable hours tocked to the target number of billable hours for last week, the last month, and FYTD by staff member and overall.

The current implementation for this story, [GroupUtilizationView](https://github.com/18F/tock/blob/master/tock/utilization/views.py#L12) and the [Utilization by Unit page](tock.18f.gov/utilization/), are not consistently displaying results for `last_week` or `fiscal_year_to_date`. It also does not exclude projects as desired, like `out of office`.

We'll decompose this view to extract the filtering and aggregation logic into smaller components.
We'll also update that logic working towards the goals defined in #980.

## We'll know we're done when

- [ ] All active 18F employees are present on the [Utilization by unit page](http://127.0.0.1:8000/utilization/)
- [ ] Utilization, billable hours, and total hours are displayed for each employee for `last_week`, `last_month`, and `fiscal_year_to_date`
- [ ] Utilization, billable hours, and total hours are displayed for overall business unit
- [ ] Projects with `exclude_from_billability==True` are excluded from all calculations, e.g. `670-Out of office`
- [ ] `GroupUtilizationView` does not contain _all_ of the aggregation and filtering logic
- [ ] New functions are established to isolate aggregation and filtering logic
- [ ] Where appropriate, we've leverage [Django's aggregations](https://docs.djangoproject.com/en/3.0/topics/db/aggregation/)

Jkrzy:
I've got a `proof of concept` in the works for this one...

Jkrzy:
To be resolved by release of #985

tbaxter-18f:
This is done.

Issue #980, opened 2020-02-26
Utilization reporting


Jkrzy:
As an 18F tock user, I want to be able to see accurate and complete summaries of  billable hours tocked to the target number of billable hours for last week, the last month, and FYTD.

For this initial effort we're going to focus only on 18F use cases and specifically:

1. Displaying utilization for an individual on their details page
2. Displaying utilization for all staff by business unit 

Below describes our destination for aggregating utilization. We'll break this down into smaller pieces and work incrementally in this direction.

Considerations for 18F employees:

 - Employees enter or depart 18F at any time
 - Employees utilization targets will vary w/ time ((80% is default, engineering supervisors are 40%, for example))
 - Employees may move across 18F business units at any time
 - Employees `billable` status may change at any time

###  Timecards
To accurately aggregate utilization information across time periods we need to be able to include only timecards which "count" towards utilization.

A timecard should be included when:

- Employee was a `current employee`
- Employee was in 18F 
- Employee was a `Billable employee`

To achieve this we need to write more attributes to each Timecard when created:

- Employee's organization - `UserData.organization`
- Employee status - `UserData.current_employee`
- Billable status - `UserData.is_billable`
- 18F status - `is_18f_employee`
- 18F unit - `UserData.unit`

### Calculating Utilization

For a given timecard we calculate utilization as follows:
```python
# `total_hours_tocked_to_excluded_projects` is the sum of 
# hours tocked to projects with `exclude_from_billability == True` 
# e.g. Out of Office
expected_hours = ( 
  (total_hours_tocked - total_hours_tocked_to_excluded_projects) 
  * timecard.billable_expectation
)

utilization = (total_hours_tocked_to_billable_projects / expected_hours)
```

### Aggregate utilization at these intervals

A `ReportingPeriod` is considered complete if it's `end_date` is today or in the past.
Timecards will only be included if they meet the criteria specified above.

#### Last week
For a given user, the timecard set is defined as:

> The Timecard, if any, associated with the most recently completed `ReportingPeriod`

`ReportingPeriod.objects.filter(end_date__lte=today).get()`

#### Last month
For a given user, the timecard set is defined as:

> The Timecards, if any, associated with the 4 most recently completed `ReportingPeriod`

`ReportingPeriod.objects.filter(end_date__lte=today).order_by('-start_date')[:4]`

#### Fiscal year to date
For a given user, the timecard set is defined as:

[...]

PR #985, opened 2020-03-04
Utilization refactor and update

Jkrzy:
## Description

For #981 and #980

This PR re-works how Tock calculates and makes available utilization data.

1. Decompose existing `GroupUtilizationView` and leverage Django aggregations where possible
2. Centralize all utilization pieces in the `utilization` app
3. Calculate/store utilization and components _for each timecard_
    a. This allows us to sum `target billable hours` across reporting periods
    b. `target billable hours` is the denominator for all utilization calculations
4. Calculate/show 18F wide utilization
5. Add Utilization to User details view
6. We default existing timecards and users to an 80% utilization target.

## Local Testing
Note -- we need updated test fixtures and/or a better alternative for generating test/demo data.
1. Run migrations
2. Update test data to exclude "Out of Office" project instances w/ `exclude_from_billability=True`
2. Calculate utilization for each timecard
    ```shell
    docker-compose run app python manage.py calculate_timecard_utilization
    ```
3. View user details pages and [utilization by unit](http://127.0.0.1:8000/utilization/)


tbaxter-18f:
> Note -- we need updated test fixtures and/or a better alternative for generating test/demo data.

lets spin that out into a fresh issue.

tock/employees/views.py
```diff
@@ -51,6 +54,7 @@ def get_context_data(self, **kwargs):
         kwargs['username'] = self.kwargs['username']
         user = User.objects.get(username=kwargs['username'])
         context = super(UserFormView, self).get_context_data(**kwargs)
+        context.update(user_billing_context(user))
```
tbaxter-18f: Just wondering, should we consider a context processor and just do this everywhere? Is that overkill?

tock/hours/models.py
```diff
@@ -222,6 +232,18 @@ class Timecard(models.Model):
                                             default=0.80, decimal_places=2, max_digits=3,
                                             verbose_name="Percentage of hours which are expected to be billable this week")
 
+    # Utilization reporting
```
tbaxter-18f: I love how well these are differentiated. Should they be non-editable though?

tock/hours/models.py
```diff
@@ -236,8 +258,47 @@ def save(self, *args, **kwargs):
         """
         if not self.id and self.user:
             self.billable_expectation = self.user.user_data.billable_expectation
+        self.calculate_hours()
         super().save(*args, **kwargs)
 
+    def calculate_utilization_denominator(self):
+        return self.billable_hours + self.non_billable_hours
```
tbaxter-18f: Wondering if we could maybe leverage notifications framework to throw a message if this is > 40.

tock/hours/models.py
```diff
@@ -236,8 +258,47 @@ def save(self, *args, **kwargs):
...
+        timecard = Timecard.objects.filter(id=self.id).annotate(billable=billable).annotate(non_billable=non_billable).annotate(excluded=excluded)[0]
+
+        self.billable_hours = round(timecard.billable, 2)
+        self.non_billable_hours = round(timecard.non_billable, 2)
+        self.excluded_hours = round(timecard.excluded, 2)
+
+        self.target_hours = self.calculate_target_hours()
+        self.utilization = self.calculate_utilization()
+
+    def _update_utilization_fields(self):
```
tbaxter-18f: This one could maybe use a comment, and I think `calc_target_hours` above could just call this for less repetition

tock/tock/templates/utilization/unit_utilization_cell.html
```diff
@@ -0,0 +1,6 @@
+<td>
```
tbaxter-18f: Super minor, but for tidiness, maybe we should put these includes in a `utlization/includes/` folder?

tock/utilization/employee.py
```diff
@@ -0,0 +1,38 @@
+from django.conf import settings
```
tbaxter-18f: another minor thing, but to avoid confusion with the employee app, what if we called this "individual"?

tock/utilization/org.py
```diff
@@ -0,0 +1,39 @@
+from django.conf import settings
```
tbaxter-18f: I assume that when we get the business units into a db table, we'll have a lot easier time with them, right?

tock/utilization/unit.py
```diff
@@ -0,0 +1,60 @@
+from django.conf import settings
```
tbaxter-18f: Is this blocked on getting the units into a database? Pretty much all the current info will break when we do that, I expect.

tock/hours/models.py
Jkrzy: Makes sense since the intention is to derive all of these attributes from other fields!

tock/hours/models.py
Jkrzy: Whoops, good catch. Vestigial and not being called anywhere -- removing altogether.

tock/utilization/employee.py
Jkrzy: I'd like to leave as-is for now `utilization.employee` feels like a specific enough scoping to avoid confusion unless we start using this piece of work beyond utilization

tock/utilization/unit.py
Jkrzy: Yeah, we'll hold off on major work here until that issue is wrapped and merged. Shouldn't be too tough to accommodate

tock/employees/views.py
Jkrzy: I think _yes_ if/when we want to display utilization details on pages beyond this specific view.

tock/hours/models.py
Jkrzy: Definitely possible, let's tackle that in a separate issue though if desired.

PR #992, opened 2020-03-10
Convert business units to DB table

### LINKED ISSUE #965: Convert business units to DB table
Currently the business units are coded as choices in employees/models.py. We need to change to (and migrate the data to) a DB table. We'd like to support sub-disciplines and easier changes.

tbaxter-18f:
## Description

Resolves #965 and converts business units to a DB table instead of choices, along with updated tests and fixtures. Also cleans up some tech debt around how orgs are defined and used.


tock/api/views.py
```diff
@@ -54,7 +54,7 @@ class UserDataSerializer(serializers.Serializer):
     organization = serializers.StringRelatedField()
 
     def get_unit(self,obj):
```
Jkrzy: I think there's an opportunity here to drop this altogether and let the `StringRelatedField` handle the work like we're doing for `organization`

tbaxter-18f: Initially I had some concerns about scope creep, but after searching I don't see where we actually call `get_unit` anywhere, so may as well drop it.

tbaxter-18f: Correction... without this one we throw a serializer error that looks weird enough that I don't want to track it down for this story. Let's log a separate issue

tock/hours/views.py
```diff
@@ -94,21 +94,13 @@ class GeneralSnippetsTimecardSerializer(serializers.Serializer):
     )
     hours_spent = serializers.DecimalField(max_digits=5, decimal_places=2)
     notes = serializers.CharField()
-    unit = serializers.SerializerMethodField()
```
tbaxter-18f: We had already defined `unit` on L88

tock/employees/admin.py
```diff
@@ -65,7 +65,7 @@ class UserDataAdmin(admin.ModelAdmin):
     search_fields = ('user__username',)
 
     def unit_info(self, obj):
-        return obj.get_unit_display()
+        return obj.unit.name
```
amymok: This needs to check for `None` object.  It is currently breaking the admin interface.

tbaxter-18f: Nice catch Amy! Updated, and added a comment.

amymok:
Merging this PR, something is not quite working in Snyk.  There's no manifest changes.

PR #1023, opened 2020-04-09
Utilization should use target hours as denominator

### LINKED ISSUE #1022: GroupUtilization report should show target hours as denominator
The group utilization report is currently using `total` reported hours for aggregation and calculation of utilization. It should be `target` hours.

Need to update logic and the test which was testing to make sure our implementation was incorrect 😅

Jkrzy:
## Description

Resolves #1022, correcting utilization aggregations and display to render `target` hours as the denominator and for use in utilization calculations.

### DIFF
--- a/tock/utilization/unit.py
+++ b/tock/utilization/unit.py
@@ -39,22 +39,22 @@ def _get_unit_billing_data(unit, recent_periods=None, fiscal_year=False):
     else:
         start, end, utilization = utilization_report(users)
 
-    data = utilization.values('username', 'billable', 'total')
+    data = utilization.values('username', 'billable', 'target')
 
     # Build context for each employee
     output_data = [{
             'username': employee['username'],
-            'total': employee['total'],
+            'denominator': employee['target'],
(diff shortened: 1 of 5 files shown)


tock/utilization/unit.py
```diff
@@ -39,22 +39,22 @@ def _get_unit_billing_data(unit, recent_periods=None, fiscal_year=False):
...
         start, end, utilization = utilization_report(users)
 
-    data = utilization.values('username', 'billable', 'total')
+    data = utilization.values('username', 'billable', 'target')
 
     # Build context for each employee
     output_data = [{
             'username': employee['username'],
-            'total': employee['total'],
+            'denominator': employee['target'],
```
tbaxter-18f, line 47: super minor, but I wonder if in this context it would be better to call it `target` rather than denominator. Your call, though.

Jkrzy: 🤷  I started with `target` but then imagined a world next week where we wanted to render `total` again 😄

PR #1028, opened 2020-04-09
Remove is_staff check for GroupUtilizationView

### LINKED ISSUE #1026: Remove `is_staff` check from utilization view
We intended this to be open and transparent. Missed a check in the view: https://github.com/18F/tock/blob/master/tock/utilization/views.py#L31

If someone has perms to log in, they should be able to see it.

Jkrzy:
## Description

Resolves #1026 by removing the `is_staff` check and updating test to access this view as a non-staff user.

### DIFF
--- a/tock/utilization/tests/test_views.py
+++ b/tock/utilization/tests/test_views.py
@@ -52,8 +52,8 @@ class TestGroupUtilizationView(WebTest):
 
         self.user = User.objects.create(
             username='regular.user',
-            is_staff=True
         )
+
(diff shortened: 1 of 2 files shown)


tock/utilization/tests/test_views.py
```diff
@@ -52,8 +52,8 @@ def setUp(self):
 
         self.user = User.objects.create(
             username='regular.user',
-            is_staff=True
```
tbaxter-18f, line 55: nice catch

Issue #1034, opened 2020-04-13
Deprecate `UserData.is_billable`


Jkrzy:
As a tock admin, I wanted to describe an employees billability solely by defining their billing targets for a given week. 

We'll know we're done when:
- [ ] `UserData.is_billable` labeling has been updated to reflect it's deprecation
- [x] `is_billable` has been removed from utilization checks (#1037)

PR #1037, opened 2020-04-13
Include all active users in utilization reports

### LINKED ISSUE #1033: Cannot access user details page of non-billable users
Tock is attempting to display utilization data for each user on the `UserDetails` page.

When attempting to calculate this for `non-billable` users, we're encountering a `NotImplementedError` because `non-billable` users are excluded from utilization data. This exception isn't handled gracefully by the backend and results in a 500 error to the user.

We'll know we're done when:
- [ ] UserDetail view for non-billable users loads as expected

Jkrzy:
## Description

Resolves #1033 and part of #1034 by removing the check for `is_billable=True` when generating utilization data.

Also includes a minor tweak to reduce number of database queries when rendering UserDetail view.

### DIFF
--- a/tock/employees/views.py
+++ b/tock/employees/views.py
@@ -7,10 +7,12 @@ from django.views.generic import DetailView, ListView
 from django.views.generic.edit import FormView
 from rest_framework.permissions import IsAuthenticated
 from tock.utils import IsSuperUserOrSelf, PermissionMixin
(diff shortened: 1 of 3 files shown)


tbaxter-18f:
Suggestion: Add help_text to is_billable in the employee model to note that it's deprecated.

Issue #1030, opened 2020-04-10
Set a ceiling for target billable hours


Jkrzy:
The current approach calculates target hours by multiplying total hours worked and billing expectation. When a user tocks more than 40 hours in a week, their target hours may exceed 40.

There should be a maximum applied to target_hours.

## We'll know we're done when
- [ ] We've researched and defined a maximum value for `target_hours`
- [ ] We've implemented the maximum in the codebase

tbaxter-18f:
I'm not sure I understand. Your target hours as a Supe is 16. That shouldn't change. What am I missing?

Jkrzy:
Current calculation is:
```python
target_hours = billable_expectation * (billable_hours + non_billable_hours)
```

If someone works more than 40 hours, target hours can go to the 🌔.

e.g.
Bob's billing expectation is 80%.
On a normal week, Bob's target hours would be 32.
Last week Bob worked 65 hours, so tock calculates his target hours as 52:

```python
target_hours = billable_expectation * (billable_hours + non_billable_hours)
# 0.8 * (65.0 + 0.0)
# target_hours = 52
```

tbaxter-18f:
I see. it seems like it's maybe related to https://github.com/18F/tock/issues/1020
 or at least having that would enable reuse as the cap.

Issue #1020, opened 2020-04-09
Allow input of target hours as a number


tbaxter-18f:
Because everyone thinks of target hours as a hole number -- "I'm expected to bill 16|24|32 hours a week" and because the percentage can twist ones' brain around as you ponder what exactly it's a percentage of, we should allow the entry as a whole number, then figure out the percentage.

One possible way this could work is to set the percentage as noneditable, create the new "expected billable hours" field, then on `save()` do the math to compute the percentage.

tbaxter-18f:
Marking as low priority because if the percentage doesn't bother folks, maybe this isn't necessary. Let's see what people say.

tbaxter-18f:
upping the priority of this.

FWIW, here's how I see this working without us having to revisit everything:
* Make the existing percentage-based target field non-editable. We'll continue to use it behind the scenes, since the math works best with a percentage.
* Add a new field to allow expressing the target as hours.
* on save() convert hours to the percentage
* Add a migration to backfill the field for everyone by reversing the current % in the system.

Thoughts?

PR #1038, opened 2020-04-13
Establishing a max number of target billable hours per timecard

Jkrzy:
## Description

Resolves https://github.com/18F/tock/issues/1030 by establishing a max number of target billable hours per timecard.

## Additional information

Calculated for each timecard as ```(BILLABLE_EXPECTATION * HOURS_IN_A_REGULAR_WORK_WEEK)```


tbaxter-18f:
Joe, the issue is that it's a percentage, right? Won't https://github.com/18F/tock/issues/1020 save us some math here, by leveraging the expectation-as-an-integer?

Jkrzy:
The specific issue here isn't related to how we're defining `billable_expectation`, it is how we're calculating `target_hours` for a given timecard.

 #1030 is about setting a cap on `target_hours` as determined for each timecard, this gets us there.

Resolving #1020 first would dictate a slightly different implementation here, moving the math a bit and adding some conditional branches.

tbaxter-18f:
Right, what I'm saying is that if we have expectation, then we _should_ (I think) be able to simply say that target_hours cannot be > than hours_expected.

Nevertheless, if there is an easier/cleaner way, we'll get to it sooner or later, and this looks like it works to me.

PR #1039, opened 2020-04-13
Management command to backfill timecard data

### LINKED ISSUE #1031: Migrate billable expectations for past time cards
If we correct someone's billable expectations it does not work for past time cards, only future ones. In order to have accurate historical data we will have to know, for each person:

Name
Expected hours
When that expectation began

We will then have to convert that to a data structure, and write a script to migrate the data.

We'll know we're done when:
[ x ] Correct info is collected
[ x ] Usable data structure written
[  ] Migration script written
[  ] migration script run

Jkrzy:
## Description

First pass at a management command to resolve #1031 

## Additional information

Provided a csv file like:
```csv
user.name, 10/1/2019, 32
user.name2, 02/02/2020, 0
...
```

Iterate through each row grabbing timecards for the provided user and starting from the provided date. For each of those timecards, calculate and assign a new billable expectation with the provided `target_value` and save, triggering an update of all calculated utilization fields.

There's some hardcoding here around the start of FY20 so that we capture the first tock reporting period of the fiscal year which starts on 9/29/2019.


Jkrzy:
Rebased against master and updated to correct linting failure.

Jkrzy:
@tbaxter-18f, I propose we test this in the `staging` environment against a database backup.

tbaxter-18f:
I think that sounds like an excellent plan

Issue #986, opened 2020-03-04
Update test fixtures and/or create a better alternative for generating test/demo data.


tbaxter-18f:
Our test data is getting stale and should be updated. Among deficiencies noted:

1. Does not account for billable target
2. Does not account for organization (or soon, business unit) models.

tbaxter-18f:
@Jkrzy started this so we could start tracking this.


We may also need some new tests to go with the new data.

amymok:
@tbaxter-18f I have a few questions here:

Are we talking about updating `data-update-deduped.json` file?

What would be some `organizations` that would be useful?  I am seeing the list of 18F Unit available right now, so what would be the difference between that and "organizations"?  Unless random ones are okay, I would like to use something that make sense for our case.

Can you also elaborate more on the billable target?  Are you talking about `billable_expectation` being set at 0.8?  Do you want to see some `employees` having a different number as well?

tbaxter-18f:
> Are we talking about updating data-update-deduped.json file?
Yes, although there are some objects created in tests. I'm not sure yet if we need updates there or not. I suspect we may have some test gaps though, so it's likely.

>What would be some organizations that would be useful? 

Because we have organization objects, and we do things with them, but currently don't really test for them?

> I am seeing the list of 18F Unit available right now, so what would be the difference between that and "organizations"?

Unit is a business unit within an org. So, like Engineering Chapter is within 18F.

Some amount of organizations (and probably more units) will be added with the issue I'm working. I don't think that will necessarily mean we shouldn't add more, but I do mention it because it may change how we look at this one. I hope to get that one out for PR today (fingers crossed).

Also note that this issue is just a scratch pad for where we'd already noted gaps. There may very well be other gaps, these are just ones Joe and I had noted while chatting one day.

amymok:
I have a few more questions before I can proceed:
- How was the data originally created?
- Are we trying to now link up employees with unit and organization?
- Carter already included Units and Organizations in this data, do we want more or this is the “official” list we want to work from?
- Do we want to fill in Org/Business Unit for everyone?
- Everyone is currently an 18F employee, if they are PIF/Solutions, we need to create some employees with those orgs and they are not 18F Employee
- What other data do we want to fill out?
- How do we want to fill out billable target/expectation?
- We are only focusing on data-update-deduped.json? There are other fixtures, do we need to worry about them? What are the usage difference between data-update-deduped.json and other fixtures?
- Anything else?

Jkrzy:
[...]

PR #1040, opened 2020-04-13
(WIP) Generating Demo User Data

amymok:
## Description
This is a partial implementation of issue #986.  The command is used to generate users and can adapt and configure.

Right now this only support generating number of users for each unit, their start dates, very basic user_data.

This needs further development to include other aspects such as reporting period creations.  This command can be run after the units are created in the demo data.

I have only been manually testing locally so far.


cryptofilegsa:
wow, this is great!  

Quick question: can we specify faker 4+ rather than any version?  Apparently 2.x broke testing things.

PR #1048, opened 2020-04-15
Removed user form

tbaxter-18f:
Removed user form since it requires custom CSS and can be done in the admin anyway.


cryptofilegsa:
circleci says this fails at ` Creating UserData for tock.tock` so it looks like all that's needed here is to remove the corresponding test.

cryptofilegsa:
also seems worth noting this is within `Creating test database for alias 'default'...`

Issue #1021, opened 2020-04-09
Better business unit admin


tbaxter-18f:
As a Tock admin, I want to see the org associated to each BU in the listview so that I can scan easier.

We'll know we're done when
* related org is displayed in listview
* we throw in an admin filter to just select for current org.
* should also add visibility and filter to "active"

PR #1061, opened 2020-04-21
Update Unit admin to include and filter on Organization

timoballard:
## Description

Improves business unit admin view by including each BU's Organization in the list view. Units can also be filtered by Organization and `active` status.

https://github.com/18F/tock/issues/1021


hbillings:
Does this also close out #968?

timoballard:
I don't believe so. #968 looks like a change to the reporting view(s), whereas this just affects the administration panel:

image

Issue #967, opened 2020-02-11
Schedule tock timecard notes in advance


allisonnorman:
From @saraheckert: As Chief of Projects, I want to be able to schedule tock timecard notes in advance so that I do not have to manually enable and disable them.

For federal holidays, Sarah manually adds a tock timecard note to remind the team to include the OOO when submitting their time. As OOO federal holidays are traditionally scheduled in advanced, we should support the ability to schedule these to enable and disable based on a date range, rather than relying on someone to manually do so, which is tedious and takes time.

Examples:
- An OOO timecard note could be scheduled to display the notification for President's Day for the period 2/16-2/22. (which will translate to "show between these dates" with a date range)
- This note, along with other ones for other federal holidays (Christmas, Thanksgiving, etc) could be set up in advance earlier in the year and will be enabled for the respective reporting period set up as tock timecard notes.

## We'll know we're done when
- [ ] Admins have the ability to schedule time card notes for all federal holidays in a calendar year so that the note shows during a given time range
- [ ] Admins no longer need to manually enable/disable time card notes that are pre-scheduled
- [ ] Ad hoc time card notes are still supported (for example, federal employees are given an unscheduled federal holiday, or a one-off message needs to be displayed).

saraheckert:
Thank you, this captures the request.

tbaxter-18f:
@Jkrzy I set this as "to-do" just based on @cryptofilegsa being already assigned, but if there are more pressing issues, I'm OK with "soon-ish" too

PR #1062, opened 2020-04-22
Add support for scheduled Timecard Notes

timoballard:
## Description

Adds support for scheduling Timecard Notes so that they display during a specified date range.

#967 

## Additional information

- Adding start/end dates to Timecard Notes is optional.
- Manually `Enabling` a Timecard Note will cause it to display even if the scheduled date range is not active.

image


tock/hours/models.py
```diff
@@ -314,6 +316,10 @@ class TimecardNoteManager(models.Manager):
     def enabled(self):
         return super(TimecardNoteManager, self).get_queryset().filter(enabled=True)
 
+    def active(self):
+        now = timezone.now()
+        return super(TimecardNoteManager, self).get_queryset().filter(Q(enabled=True) | Q(display_period_start__lte=now, display_period_end__gt=now))
```
tbaxter-18f, line 321: Wouldn't this blow up if start and end were left blank?

tock/hours/models.py
```diff
@@ -349,6 +355,16 @@ class TimecardNote(models.Model):
         default=True,
         help_text='Toggle whether or not the note is displayed in a timecard.'
```
tbaxter-18f, line 356:
```suggestion
        help_text='Toggle whether or not the note is displayed in a timecard. Note that when this is checked any start and end date defined below will be ignored.'
```

tock/hours/models.py
```diff
@@ -349,6 +355,16 @@ class TimecardNote(models.Model):
...
     )
+    display_period_start = models.DateField(
+        null=True,
+        blank=True,
+        help_text='The start date for displaying this note. Note: manually enabling will override this setting.'
+    )
+    display_period_end = models.DateField(
```
tbaxter-18f, line 363: What if you wanted to start a note on May 1 and leave it up for an undetermined amount of time? How would that work?

timoballard: Right - this implementation doesn't support that, as it requires you to specify either both the start and end date, or neither. It would be pretty straightforward to modify this to support that use case if that's something we want to do.

tock/hours/models.py
timoballard: Based on my testing and understanding, we should be okay here. The field lookups (`lt`, `gt`, etc) resolve to false in cases where the value is `null`

tock/hours/models.py
tbaxter-18f: I suppose one could just put in a far-future end date and call it good, too.

Jkrzy:
Thanks @timoballard!

I _think_ we missed a migration here, can you double check and submit a follow-up PR if necessary?

timoballard:
Good catch @Jkrzy - I sure did. I overlooked that @tbaxter-18f's suggested change on the help text would imply a migration. 

#1067

Issue #993, opened 2020-03-11
investigate viability of postgres update


cryptofilegsa:
using 9.6.11 but most recent 9.6 of postgres is 9.6.17 see: https://www.postgresql.org/about/news/2011/

Would be nice to catch those changes.

tbaxter-18f:
Is there any update on this?

What is the advantage of upgrading for a few point releases?
What are the potential downsides?

I'm wondering if this is actually worth doing, especially after 28 days of not doing it.

cryptofilegsa:
Well, this is a lot longer than 28 days of not upgrading involved here.
Changelogs to be reviewed for blockers:
https://www.postgresql.org/docs/release/9.6.12/
https://www.postgresql.org/docs/release/9.6.13/
https://www.postgresql.org/docs/release/9.6.14/
https://www.postgresql.org/docs/release/9.6.15/
https://www.postgresql.org/docs/release/9.6.16/
https://www.postgresql.org/docs/release/9.6.17/

tbaxter-18f:
I mean we've been looking at it for 28 days.

hbillings:
@cryptofilegsa Rather than combing through the changelogs, I'd recommend the following:
- [ ] Try upgrading locally.
- [ ] If local upgrade works, open a PR with the changes.

We only need to worry about reviewing changelogs for large breaking changes, which shouldn't happen in a patch release (assuming the versioning system follows [semantic versioning](https://semver.org/), which Postgres does). Patch releases are for backwards-compatible bug fixes and shouldn't require deep dives.

cryptofilegsa:
Actually per @Jkrzy this is most likely a cloud.gov piece, so looking into how they manage db versioning.  

See https://cloud.gov/docs/services/relational-database/

> The PostgreSQL and MySQL plans are configured to automatically upgrade currently-running dedicated instances to the most recent compatible minor version available via AWS RDS GovCloud (US).

May be handled via AWS management console see: [docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_UpgradeDBInstance.Maintenance.html](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_UpgradeDBInstance.Maintenance.html)

Perhaps @tammersaleh or @bengerman13 have guidance, per: [github.com/cloud-gov/aws-broker](https://github.com/cloud-gov/aws-broker)

Continuing to poke around as the most cloud.gov way to get this done.

tbaxter-18f:
[...]

PR #1063, opened 2020-04-23
upgrade db version only for development branch

cryptofilegsa:
postgresql 9.6.17 incorporates fixes referenced here: https://github.com/18F/tock/issues/993
purpose of this is to stick closer to current db fixes for 9.6 rather than falling too far behind


cryptofilegsa:
replaced by https://github.com/18F/tock/pull/1065

Issue #1024, opened 2020-04-09
Make "organization" a required field


tbaxter-18f:
With us sorting and filtering and reporting utilization on it, we need this data available for each user in Tock.

## We'll know we're done when
- [x] The `organization` field on the `UserData` model has been updated to be a required field for form validation
- [x] Migrations have been generated to update the database
- [x] `UserDataAdmin` has been updated so that `organization` is sortable
- [x] `UserDataAdmin` has been updated allowing filtering for instances with no values for `organization`

tbaxter-18f:
@cryptofilegsa can you give an update on how this is coming? Are there blockers?

cryptofilegsa:
can sensible data mean "18F" is default?

tbaxter-18f:
I don't see a problem with that.

cryptofilegsa:
change at https://github.com/18F/tock/commit/777a73fca228f7e6c6524c03c4d9c3a67a6830bb , more may be needed once I give migrate.py a shot.

cryptofilegsa:
OK, I have the change made locally that works, the migration has been produced, running the migration works locally and puts the database in a state that is operational and I've tested the fix.  However, I don't think this will work on CircleCI or with testing on other devices without changes to the default data set.

Jkrzy:
Updated original issue `We'll know we're done when` from convo in :slack:

cryptofilegsa:
@tbaxter-18f changes here are ongoing discussion in #tock-dev.   @Jkrzy question re: github etiquette: given **all** of the criterion (definition of done) have changed here, can we open a new issue to track the new work?  happy to change focus to new issue as you suggest.  Where possible would like to avoid anything that could be intentionally confusing to anyone following the trail of thought in our work.

tbaxter-18f:
If we were to create a new issue it would duplicate this one and require cross-linking the two, so I'm not seeing an advantage in doing so. The issue we're solving is unchanged, only our approach to solving it.

cryptofilegsa:
for the new approach with four steps from June 3, late yesterday, tutorials pointed out to me were inadequate but I found needed documentation in the #modeladmin-options section of the django documents here: https://docs.djangoproject.com/en/3.0/ref/contrib/admin/#modeladmin-options

PR #1064, opened 2020-04-23
organization required and given default

cryptofilegsa:
organization field is made a requirement and given a default value of '18F'

## Description
to address  https://github.com/18F/tock/issues/1024

## Additional information
still need to run makemigrations


cryptofilegsa:
replaced by PR1065

PR #1065, opened 2020-04-23
default for Organization set with DEFAULT_ORG_ID

cryptofilegsa:
default for Organization set with constant

## Description
makes Organization have a required default value, set with a constant now

## Additional information
to address https://github.com/18F/tock/issues/1024
in need of custom function to retrieve foreign key value for '18F'


cryptofilegsa:
fails with  Key (organization_id)=(1) is not present in table "organizations_organization".

Issue #1071, opened 2020-04-27
UserData API returns some incorrect data


mgwalker:
Some of the user data returned by the UserData API is incorrect. E.g., some people who are no longer with 18F still have the `current_employee` flag set to true. In other cases, non-18F employees have the `is_18f_employee` flag.

You can see the list that popped up for me when testing Angry Tock here: https://gsa-tts.slack.com/archives/C1JFYCX3P/p1588016935306000?thread_ts=1588016905.305900&cid=C1JFYCX3P

Jkrzy:
Proposing we close this one out by:
- [x] Merging #1074 
- [ ] Deprecating `current_employee` and `is_18f_employee` attributes, replacing them with properties that evaluate based on other data. I _think_ we can calculate these and feels duplicative to track this data in multiple attributes if so.

`current_employee` -  True if `user__is_active`
`is_18f_employee` -  True if `organization.name == "18f"`

PR #1074, opened 2020-04-30
Add is_active property to UserData model

mgwalker:
## Description

Copies the `is_active` property from an associated `User` model into the `UserData` model, and updates the UserData API serializer to include that property. This could help with #1071, per @Jkrzy's comment in Slack.

### DIFF
--- a/tock/employees/models.py
+++ b/tock/employees/models.py
@@ -111,6 +111,10 @@ class UserData(models.Model):
         return '{0}'.format(self.user)
 
     @property
+    def is_active(self):
+        return self.user.is_active
+
+    @property
     def is_billable(self):
         """
         True if user is currently expected to bill more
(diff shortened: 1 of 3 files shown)


tock/employees/models.py
```diff
@@ -110,6 +110,10 @@ class Meta:
     def __str__(self):
         return '{0}'.format(self.user)
 
+    @property
+    def is_active(self):
```
tbaxter-18f, line 114: I wonder if there's a good argument for us to use `@cached_property` here, thinking this is relatively expensive to compute (requires a db traversal) and changes infrequently. 

I wonder about this with a lot of our properties, tbh.

https://docs.djangoproject.com/en/3.0/ref/utils/#django.utils.functional.cached_property

Jkrzy:
Thanks @mgwalker!!!

PR #1078, opened 2020-05-05
Replace if_18f_employee field with property

Jkrzy:
## Description
For #1034, dropping `is_18f_employee` from UserData model.

As we move towards Organization being a required field, we drop this independent field and replace with a property whose value is derived from Organization.

We'll no longer have to maintain a separate field in addition to Organization in order to serve data re: a user being in 18F or not.


tock/employees/tests/test_models.py
```diff
@@ -212,3 +208,24 @@ def test_token_is_delete_on_active_is_false(self):
...
+            description='not_18f',
+            active=True
+        )
+        self.regular_user_userdata.organization = not_18f
+        self.regular_user_userdata.save()
+        self.assertFalse(self.regular_user_userdata.is_18f_employee)
+
+    def test_is_18f_employee_true_if_18f(self):
+        """True if org is named 18F"""
+        self.assertTrue(self.regular_user_userdata.is_18f_employee)
```
mgwalker, line 231: This one doesn't need an org set because it's set in the test setup method?

tbaxter-18f: Correct. Up on L65

Jkrzy: Yaas, but that's not clear from this one line test alone--- :enhance:

PR #1079, opened 2020-05-05
org change with migration

cryptofilegsa:
PR submitted prior to call to discuss organization model strategy with JK


Jkrzy:
Thanks @cryptofilegsa, closing this one out after our convo.

PR #1080, opened 2020-05-06
Tock 1057 unit utilization

### LINKED ISSUE #1057: When aggregating utilization data better account for movement throughout TTS
As a tock user reporting on utilization I need complete and accurate data for the fiscal year therefore hours billed by staff no longer at 18F should be included in business unit and organizational aggregates.

The current approach to utilization reporting only aggregates data from users which are current and active 18F employees. If a user departs TTS, their billable and target hours within a reporting period should still count towards business unit and organization levels.

## We'll know we're done when
- [x] We've defined the conditions when the timecards of employees no longer with 18F  should be included in utilization aggregates
- [ ] We've implemented timecard level attributes to allow inclusion of said timecards
- [ ] We've updated the documentation

timoballard:
## Description

Adds `Organization` and `Unit` attributes to `Timecard` objects, which are copied from the user when the `Timecard` is saved. Migration includes script to backfill `Organization/Unit` data using the owning user's current assignment.

Updates `GroupUtilizationView` to include timecard data for users that are no longer assigned to the `Organization/Unit`.

resolves #1057

## Additional information
- The backfilling of `Org/Unit` data naively applies the user's current `Org/Unit` to all of their past `Timecards`, which is not correct in all cases.


tock/hours/models.py
```diff
@@ -237,6 +238,9 @@ class Timecard(models.Model):
                                             default=Decimal(settings.DEFAULT_BILLABLE_EXPECTATION), decimal_places=2, max_digits=3,
                                             verbose_name="Percentage of hours which are expected to be billable this week")
 
+    organization = models.ForeignKey(Organization, null=True, blank=True, on_delete=models.CASCADE)
```
tbaxter-18f, line 241: Wondering two things:
1. Should they be required?
2. Should they be non-editable, since they're auto-set?

timoballard: 1. It makes sense to me for them to be required, but at this level our hands are somewhat tied by the fact that they're not required at the `User` level
2. Yes, I think so

Jkrzy: Ooo, I think these **need** to be editable so that timecards can be corrected via the admin when necessary.

tock/hours/models.py
```diff
@@ -265,6 +269,8 @@ def save(self, *args, **kwargs):
             self.billable_expectation = self.user.user_data.billable_expectation
         if self.id:
             self.calculate_hours()
+            self.organization = self.user.user_data.organization
+            self.unit = self.user.user_data.unit
```
Jkrzy, line 273: We should set these when a timecard is first created to avoid overwriting these fields if/when a timecard is updated at a later time.

tock/hours/tests/test_models.py
```diff
@@ -185,6 +196,24 @@ def test_timecard_max_target_hours(self):
 
         self.assertEqual(self.timecard.target_hours, self.timecard.max_target_hours())
 
+    def test_timecard_saves_user_organization(self):
```
Jkrzy, line 199: Related to above, let's tweak these tests to confirm both the application of the org/unit attributes but that they're only effect on creation not every save.

tock/utilization/unit.py
```diff
@@ -10,51 +10,77 @@
...
+    # all users that have ever created a timecard for this unit
+    users = User.objects.filter(timecards__unit=unit).distinct()
+
+    last_week = _get_unit_billing_data(users, unit)
+    last_month = _get_unit_billing_data(users, unit, recent_periods=settings.RECENT_TIMECARDS_FOR_BILLABILITY)
+    this_fy = _get_unit_billing_data(users, unit, fiscal_year=True)
 
     context = _build_utilization_context(last_week, last_month, this_fy)
 
+    usernames_without_hours = [e.username for e in _employees_without_hours(users, last_week['data'] + last_month['data'] + this_fy['data'])]
```
Jkrzy, line 23: Could this be simplified if we assume that `this_fy` is the boundary of the reporting for `unit_billing_context`, then doing that filtering in the `User` Queryset above?

tock/utilization/utils.py
```diff
@@ -33,19 +33,22 @@ def _build_utilization_query(users=None,recent_periods=None, fiscal_year=False):
     """
 
     if not fiscal_year:
-        period_filter =  _limit_to_recent_periods(recent_periods)
+        filter =  _limit_to_recent_periods(recent_periods)
```
Jkrzy, line 36: Opportunity here to tweak just a _bit_ more and append an underscore to differentiate this from the python built-in [filter](https://docs.python.org/3/library/functions.html#filter)

tock/utilization/unit.py
timoballard: I think I see what you mean - `last_week` and `last_month` timecards are subsets of `this_fy` timecards for a given user, so if we additionally filter the QS on line 15 to timecards dated within this FY, that would give us the set of `employees_with_hours`

I think... 🤔

timoballard: Yes - this works, and is simpler 👍 

I have to update some of the tests around this to make them a bit more date-resilient (they were setup with hardcoded RP dates, but the FY is determined by looking at the system clock). Should be ready for review by tomorrow AM.

Jkrzy:
We'll need to update this one and re-align with the migrations merged in #1082

Jkrzy:
Looks great, thanks @timoballard!

We'll want a method by which we can facilitate backfilling data for existing timecards, but can tackle that data gathering/application in another issue.

PR #1082, opened 2020-05-08
Input target hours as a number

### LINKED ISSUE #1020: Allow input of target hours as a number

timoballard:
## Description
Adds support for managing `Employee` target hours as a whole number, rather than a percentage.

Adds `expected_billable_hours` as a new field on `Employees`, and changes `billable_expectation` to a property, which derives its value from `expected_billable_hours`.  

resolves #1020 

## Additional Information
- Migration includes script to backfill `expected_billable_hours` for existing `Employees`.


tock/employees/models.py
```diff
@@ -130,6 +131,14 @@ def is_billable(self):
         """
         return self.billable_expectation > Decimal(0.0)
 
+    @cached_property
+    def billable_expectation(self):
```
tbaxter-18f, line 135: Interesting. In my head we'd have used a non-editable field, but i like this approach too!

tock/employees/models.py
```diff
@@ -130,6 +131,14 @@ def is_billable(self):
...
+        """
+        Returns the employee's billable expectation, expressed as a percentage
+        of their hours worked
+        """
+        return Decimal(self.expected_billable_hours / settings.HOURS_IN_A_REGULAR_WORK_WEEK)
```
tbaxter-18f, line 140: Should we take advantage of the ability to set precision on Decimal in order to avoid repeating decimals? Or, since it's hidden from users, is the precision no problem?

tock/employees/models.py
timoballard: Yeah, could go either way here. I started with a non-editable field, but making it a property seemed sort of self-documenting, and we avoid storing redundant data (`billable_expectation` should always just be a function of `expected_billable_hours`). 🤷‍♂️

tock/employees/models.py
timoballard: Yeah, I think we should 👍 . On that note - would it be useful to still display the `billable_expectation` in the admin pages (alongside the `expected_billable_hours`)?

tbaxter-18f: I think just the expectation in hours. That's how everyone thinks of them.

tock/employees/models.py
```diff
@@ -82,9 +82,10 @@ class UserData(models.Model):
     start_date = models.DateField(blank=True, null=True, verbose_name='Employee start date')
     end_date = models.DateField(blank=True, null=True, verbose_name='Employee end date')
     current_employee = models.BooleanField(default=True, verbose_name='Is Current Employee')
-    billable_expectation = models.DecimalField(validators=[MaxValueValidator(limit_value=1)],
-                                             default=Decimal(settings.DEFAULT_BILLABLE_EXPECTATION), decimal_places=2, max_digits=3,
-                                             help_text="Percentage of hours (expressed as a decimal) expected to be billable each week")
+    expected_billable_hours = models.IntegerField(validators=[MaxValueValidator(limit_value=settings.HOURS_IN_A_REGULAR_WORK_WEEK)],
+                                                default=settings.DEFAULT_EXPECTED_BILLABLE_HOURS,
+                                                help_text="Number of hours expected to be billable each week")
```
Jkrzy, line 87:
```suggestion
                                                help_text="Number of hours expected to be billable in a 40 hour work week")
```

tock/employees/migrations/0031_userdata_expected_billable_hours.py
```diff
@@ -0,0 +1,26 @@
+# Generated by Django 2.2.12 on 2020-05-07 20:04
+
+from django.conf import settings
+import django.core.validators
+from django.db import migrations, models
+
+def backfill_expected_billable_hours(apps, schema_editor):
```
Jkrzy, line 7: This doesn't look like it's called anywhere, was it intended to run as part of this migration?

timoballard: 😲 

Yes - good catch! I was briefly toying with combining the two migrations into one and must have mistakenly removed the call to this

PR #1088, opened 2020-05-20
User/Project organization validation

### LINKED ISSUE #1060: Data validation request
Would it be possible to add data validation by organization. For example, there are non 18F-ers using 18F non-billable GSA mandated.

For the folks who are non 18F using these codes, I am unsure of how this is calculating into our numbers or if it matters. I think Matt Spencer may be pulling the GSA mandated hours when calculating overhead costs, but this may be something we want to run by him. 

Ex. 
 https://tock.18f.gov/employees/e/anil.chaudhry/ has had 16 hrs/week for several weeks, https://tock.18f.gov/employees/e/brandon.kirby/ has a couple of hours,
https://tock.18f.gov/employees/e/matthew.montano/ was 20 hours last week,
https://tock.18f.gov/employees/e/edward.golden/ had 15 hours,
https://tock.18f.gov/employees/e/kathy.mcneill/ had 40 hours last week (looks new)
https://tock.18f.gov/employees/e/tamieca.hamlin/ had 15 hours, also looks new,
https://tock.18f.gov/employees/e/ambuj.neupane/ had 12, also looks new.)

## ([update](https://github.com/18F/tock/issues/1060#issuecomment-629375775))We'll know we're done when:
- [ ] When creating a new billable Project, Organization will be a required field
- [ ] When creating a new non-billable Project, Organization will be an optional field
- [ ] Employees are allowed to tock to:
  - [ ] Projects associated with their Organization
  - [ ] Projects not associated with any Organization
- [ ] Existing Projects will have their Organization backfilled, which Matt offered to take the lead on

timoballard:
## Description

Filter the list of Tockable projects to those assigned to the employee's organization and those that are not assigned to any organization.

Newly created billable projects must have an assigned organization.

resolves #1060 

## Additional information
* Assigns existing projects to organizations based on project name (mapping confirmed with Matt Spencer)
* Leaves `Project.Organization` null in cases where it cannot be inferred from the project name


tock/projects/admin.py
```diff
@@ -115,6 +116,9 @@ class ProjectAdmin(admin.ModelAdmin):
         'notes_required',
         'organization__name'
     )
+    list_editable = (
```
tbaxter-18f, line 119: nice

tock/hours/views.py
```diff
@@ -479,7 +479,8 @@ def get_context_data(self, **kwargs):
         # TODO: This is inefficient because we're writing over the
         # already-generated choices. Ideally we should be passing these
         # into the formset constructor.
-        choices = projects_as_choices(reporting_period.get_projects())
+        projects = reporting_period.get_projects().filter(Q(organization=self.request.user.user_data.organization) | Q(organization=None))
```
tbaxter-18f, line 482: Without actually looking at get_projects, this makes me wonder if we should shift some of this filtering into it. It's already a manager, isn't it? Not saying we _should_ , since I haven't looked at it. But wondering.

timoballard: Yeah. I stopped short of pushing the filtering into `get_projects` because `ReportingPeriods` have no notion of users, so we'd have to pass either the user or the user's org into `get_projects`. I didn't love any of the options so I went with the least invasive one 😁

Jkrzy:
We'll need some extra release notes around this to let people know their list of tockable projects will be changing.

PR #1093, opened 2020-05-29
Iteration for organizations migration

cryptofilegsa:
Also includes orgchange.py to update the test data


tock/employees/migrations/0034_auto_20200529_1230.py
```diff
@@ -0,0 +1,34 @@
...
+    for orgs in Employees.all():
+        try:
+            i = 1
+            org_data = Employees.objects.get(organization=null)
+            orgs.organization = org_data.organization
+            orgs.organization = i
+            orgs.save
+            i++
+        except ObjectDoesNotExist:
+            pass
```
Jkrzy, line 19: Thanks @cryptofilegsa!

Let's focus on defining and refining the iteration for now. We want to iterate through employees which have no organization assigned.

The `orgs` variable assigned name in the `for` statement is a bit confusing since we're iterating through instances of the `Employee`  model not `Organization`. The usage of the `Employees` queryset is also missing a piece, this would fail with an `AttributeError`

Django provides handy filtering functionality to return a QuerySet containing only records meetings that criteria. Take a look at how they're using `.filter()` in the [second part of the Django tutorial](https://docs.djangoproject.com/en/3.0/intro/tutorial02/#playing-with-the-api)

PR #1101, opened 2020-06-04
Test for submitted cards in utilization views

tbaxter-18f:
Makes sure only submitted timecards are used in utilization reports.


Jkrzy:
Good as-is, this covers excluding non-submitted timecards from utilization reporting. Thanks  @tbaxter-18f! We'll still need to update the `Recent Tocks` table on the `UserDetailView`

tbaxter-18f:
Joe, would you like me to take care of that real quick before we merge this?

Jkrzy:
That would be great, thanks!

tbaxter-18f:
Ready for re-review.

PR #1102, opened 2020-06-04
1024 make organization required

cryptofilegsa:
https://github.com/18F/tock/issues/1024


cryptofilegsa:
two ways to past the failing test, have gotten both to work, either creating organization object or sourcing default data from organizations.json

tock/employees/tests/test_admin.py
```diff
@@ -23,11 +25,13 @@ def setUp(self):
             as_end_date=datetime.date.today() + datetime.timedelta(days=20),
             account_type='Expense'
         )
+        # Organization.objects.create(name="18F")
```
Jkrzy, line 28: Don't need to leave this handing around if it's commented out.
```suggestion
```

cryptofilegsa: you mean hanging around?

Issue #1105, opened 2020-06-09
Make updating existing timecards easier


Jkrzy:
As a tock administrator I need to update existing/previously submitted timecards so that they accurately reflect the state of world at the time of their submission. Currently this can be done via the admin on a per timecard basis, but modifying multiple timecards at once would greatly facilitate this work and make it easier to remedy incorrect utilization data.

Tuning the Tock admin feels like the most reasonable approach here.

## We'll know we're done when the the Timecard admin page has been updated so that

- [x] Tock Admins can perform in-line edits for fields used in utilization calculations (billable_expectation, organization, ???)
- [x] Tock Admins can filter timecards by Employee

cryptofilegsa:
I took a look and the current /admin/hours/timecard/ is filterable by reporting period only.
Apparently employee is labelled as _User_ in that form.
In any case, a filter by User (employee) can be added.

billable_expectation is editable as a text field.  
organization is editable as a pop-up menu
Can you clarify what is meant by "in-line edits" ?

While doing this, any others to be added?

Jkrzy:
In-line edits being, we want the ability to edit multiple timecards at once from the listing page of the admin. There are a few other models for which we've provided this functionality, like `ProfitLossAccount`

PR #1118, opened 2020-06-19
Filter by employee 1105

cryptofilegsa:
## Description

Admin for timecards list is now filterable per employee in addition to timecards.

## Additional information

addresses second half of https://github.com/18F/tock/issues/1105


tock/hours/admin.py
```diff
@@ -24,6 +24,22 @@ def queryset(self, request, queryset):
         return queryset
 
 
+class EmployeeListFilter(admin.SimpleListFilter):
```
Jkrzy, line 27: Do we need this? Doesn't look like it's being used anywhere.

cryptofilegsa: not needed, removing...

cryptofilegsa:
ready for re-review

Issue #1084, opened 2020-05-11
Hours tocked to `microrequest` projects should count towards IC billable hours


Jkrzy:
A handful of project lines tracking `microrequest` are not being included as billable hours in utilization reports.  e.g., `1149`

This is because
> We don't bill the partner for them, which is why the tock line isn't listed as billable, but they should count towards the IC's billable hours

We'll need to work with the account management team on this one to ensure they can continue to pull the data they need.

This may take the form of being able to differentiate `Accounting Codes` which are or are not invoiced in addition to their `billable` status.
 
## We'll know we're done when:
- [ ] Microrequest tock lines are considered billable and included in utilization reporting
- [ ] <s>AM</s> Finance team is able to pull data required for invoicing without including unecessary data from these same `microrequest` projects.

tbaxter-18f:
We'll need a way to differentiate "billable" microrequests (with a partner) vs "non-billable" (usually internal).

@hannahkane & @rrefoy do y'all have thoughts?

hannahkane:
@tbaxter-18f - the partner-billable microrequests would show up on the project tock line. Non-billable microrequests would show up as the general non-billable. The four TTS-specific microrequests are the only exception (that I know of). It's Matt Spencer's (gh handle?) opinion that matters, but I'm wondering what havoc it would wreak if we marked them as billable in tock?

rrefoy:
Can i see a use case for what @Jkrzy is describing? Why wouldn't we bill the partner for those hours?

tbaxter-18f:
Thanks Hannah. I wasn't even thinking abou those. So we have three types of microrequests

1. small help on a project. These are already billable to the project and it's up to the IC to do so.
2. truly non-billable. These are internal writing lab requests and such. There's really no one to bill them too.
3. TTS microrequests that fall under the agreement. These are the ones @Jkrzy and @hannahkane were referring to. They're not marked as billable because they've already been billed, but I they should count toward utilization because they are money-making work, right?

Tapping in @martyspencer for visibility

hannahkane:
@tbaxter-18f - that's right. The first two are straightforward. The third is an edge case. They're billable from the perspective of hours utilization, but the partner doesn't receive a bill as it was worked out in an overhead costs sharing conversation at the beginning of the fiscal year (I believe).

tbaxter-18f:
Right. And currently they are not being counted toward the individual's utilization total. Since they are doing work and creating impact for the partner, we think that should count as utilization.
 
[...]

PR #1121, opened 2020-06-24
include_in_utilization in model, filter, and editable to support microrequests issue #1084

cryptofilegsa:
## Description

This pull request provides underlying database and admin filtering to support project level overrides to both include or exclude from utilization reporting.  Reference issue:  https://github.com/18F/tock/issues/1084

For each Project/Tock line:

Added an include_in_utilization boolean field, defaults to that project's billable status

include_in_utilization defaults to False but can be custom-set to True in the admin interface.

(can be set for the 4 edge cases we run into every week)

Further work:

Look into how utilization reports could filter on the new Project.include_in_utilization field

## Additional information

yellow highlighting of the enhancements in the projects admin page:

* Screenshot images
image


tock/projects/models.py
```diff
@@ -247,6 +247,11 @@ class Project(models.Model):
         default=False,
         help_text='Check if this project should be excluded from calculations of billable hours, e.g. Out of Office'
     )
+    include_in_utilization = models.BooleanField(
+            default=False,
```
Jkrzy, line 251: For all existing Projects, we want to set the value of this to the project's billable status. This sets _all_ of them to `False`

tock/projects/models.py
```diff
@@ -247,6 +247,11 @@ class Project(models.Model):
...
+            help_text="For handling microrequests, per github.com/18F/tock/issues/1084"
```
Jkrzy, line 252: There's an opportunity here to use the `help_text` attribute to describe the functionality that setting this field to `True` provides. Noting the origin in the code does provide much utility.

tock/hours/admin.py
```diff
@@ -86,7 +87,7 @@ class TimecardAdmin(admin.ModelAdmin):
     inlines = (TimecardObjectInline,)
     list_display = ('user', 'reporting_period', 'billable_expectation', 'organization', 'submitted')
     list_editable = ('organization', 'billable_expectation')
-    list_filter = (ReportingPeriodListFilter, 'reporting_period',)
+    list_filter = ( 'user', ReportingPeriodListFilter, 'reporting_period')
```
Jkrzy, line 90: I believe these changes are already in `main`

tock/projects/models.py
cryptofilegsa: tactically, starting with the most common value and can update in separate PR

tock/projects/models.py
cryptofilegsa: addressed in a9259f2

tock/projects/models.py
timoballard: I think there are a couple of different ways to handle this, but I tend to extend the auto-generated migration to set the defaults. Here's an example:

https://github.com/18F/tock/blob/1b8b88039d4b52e8c184281e392c1fde65c890c6/tock/hours/migrations/0055_auto_20200430_1742.py#L7-L40

timoballard:
Left a comment about setting the `include_in_utilization` default value based on billable status as part of the migration. Happy to pair on this if you'd like, @cryptofilegsa

tock/projects/migrations/0027_project_include_in_utilization.py
```diff
@@ -0,0 +1,26 @@
+# Generated by Django 2.2.13 on 2020-06-23 16:29
+
+from django.db import migrations, models
+
+def make_default_utilization(apps, schema_editor):
+    Projects = apps.get_model('projects','Project')
+
+    for project in Projects.objects.all():
+        project.include_in_utilization = project.is_billable
```
timoballard, line 9: turns out `is_billable` isn't accessible in this context - if you switch this to...

`project.include_in_utilization = project.accounting_code.billable`

...you should be good.

tbaxter-18f:
👋  would this be useful to finish up?

Jkrzy:
@tbaxter-18f, if you're still up for it -- absolutely this would be good to wrap up!

kfoley-18F:
No longer needed with weekly billing.

PR #1136, opened 2020-07-31
Utilization Analytics MVP

neilmb:
## Description

There is lots of valuable information about 18F operations in the Tock database. This is a very early step in making that data visible. It adds a page under "Reports" called "Analytics" that currently has just a single graph of the total hours billed over time broken down by billable and non-billable.

Screen Shot 2020-07-31 at 3 54 20 PM

There are many more relevant graphs that could be added here in the future, but this is an MVP to establish that we can include these kinds of graphs.

## Additional information

This adds a dependency on the python `plotly` library for creating the plots in a way that can be included in the template and the template for the page then includes the Plotly.js library on-page from their CDN. If we need to ship a fixed version of that JS library, we could figure out how to do that instead.


tock/utilization/views.py
```diff
@@ -50,3 +57,92 @@ def get_context_data(self, **kwargs):
...
+        yaxis_title="Hours",
+        title="Total Hours recorded vs. Time",
+    )
+
+    plot_div = plot(fig, output_type='div', include_plotlyjs=False)
+    return plot_div
+
+
+def _utilization_data(start_date, end_date):
+    Timecard = apps.get_model("hours", "Timecard")
```
Jkrzy, line 102: We're[ storing sums for billable and non-billable on each Timecard](https://github.com/18F/tock/blob/main/tock/hours/models.py#L247) can we leverage those here instead of aggregating across all `timecardobjects`?

tock/utilization/views.py
```diff
@@ -50,3 +57,92 @@ def get_context_data(self, **kwargs):
...
+    def dispatch(self, *args, **kwargs):
+        """
+        Resolve recent reporting periods.
+
+        Although recent_rps is set to the last four reporting periods,
+        we could accept a form response that allows the user or app to
+        dynamically customize number of periods to include in the queryset.
+        """
+        if not self.request.user.is_authenticated:
+            return self.handle_no_permission()
```
Jkrzy, line 135: doctring here looks like a bit of duplication from `GroupUtilizationView` that should be removed.  And in looking at this I think both of these views probably don't need a customized `dispatch` function 😄. I'll file a separate issue to take a closer look at that.

tock/utilization/views.py
neilmb: I don't see any timecards in the test database that have non-null values for `billable_hours` and `non_billable_hours`. 

Is that a problem with the `test_data/`? Given that the migration for those fields comes from 2020-04-03, it seems likely that the test database just hasn't been updated to match the presence of the new fields...which makes it hard to test the implementation of this feature.

I'd rather not block this on waiting for #986 to be done first, but that would be a possible way forward.

tock/tock/templates/utilization/utilization_analytics.html
```diff
@@ -0,0 +1,13 @@
+{% extends "base.html" %}
+{% block navigation %}
+  {% include "_navigation.html" %}
+{% endblock %}
+
+{% block content %}
+  <script src="https://cdn.plot.ly/plotly-latest.min.js"></script>
```
Jkrzy, line 7: Let's bring this in house and serve ourselves for consistency by pulling the version we need into `tock/tock/static/vendor`

https://github.com/18F/tock/blob/main/docs/dependency-management.md#production

tock/utilization/views.py
Jkrzy: @neilmb wrangled up some test data and I believe has a WIP commit to refactor this a bit. Resolving for now.

tock/utilization/views.py
```diff
@@ -50,3 +57,82 @@ def get_context_data(self, **kwargs):
...
+    )
+
+    plot_div = plot(fig, output_type='div', include_plotlyjs=False)
+    return plot_div
+
+
+def _utilization_data(start_date, end_date):
+    Timecard = apps.get_model("hours", "Timecard")
+    billable_filter = Q(timecardobjects__project__accounting_code__billable=True)
+    data = (Timecard.objects.filter(reporting_period__start_date__gte=start_date,
```
Jkrzy, line 104: Opportunity here to limit this to submitted timecards by filtering on `Timecard.submitted`

Jkrzy:
Looks :amaze: @neilmb, thank you!!
 Made a few comments inline where there are some opportunities for changes.

neilmb:
@Jkrzy Comments addressed: filter on submitted, vendor plotly into static, sum using existing billable/non-billable totals on timecard.

PR #1141, opened 2020-08-07
Include raw data table and tweak graph

neilmb:
## Description

This builds on the new utilization graph with a textual data table, improved hover labels, and CSV download of the raw data table.

Screen Shot 2020-08-07 at 14 24 56

## Additional information

There is an added Python dependency on `pandas` which is very, very useful for handling data and graphics. We use a custom template tag to render the data frame into the HTML table and pass the data frame as an argument to make the plot.

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "b1fc1ec32ebaa8f52b4b4af000715645e972802aeb1c88093c37b73778befec8"
+            "sha256": "42ad89926313f60ad1399dc8492181975f76ab55e515d338b17307f44634bfc1"
         },
         "pipfile-spec": 6,
         "requires": {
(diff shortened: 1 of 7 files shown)


Pipfile.lock
```diff
@@ -516,11 +578,11 @@
...
         "django": {
             "hashes": [
-                "sha256:edf0ecf6657713b0435b6757e6069466925cae70d634a3283c96b80c01e06191",
-                "sha256:f2250bd35d0f6c23e930c544629934144e5dd39a4c06092e1050c731c1712ba8"
+                "sha256:3e2f5d172215862abf2bac3138d8a04229d34dbd2d0dab42c6bf33876cc22323",
+                "sha256:91f540000227eace0504a24f508de26daa756353aa7376c6972d7920bc339a3a"
             ],
             "index": "pypi",
-            "version": "==2.2.14"
+            "version": "==2.2.15"
```
Jkrzy, line 585: Catching a django update here, bonus! Thanks!

Issue #1119, opened 2020-06-22
Adjust billable hour expectations


tbaxter-18f:
Per https://gsa-tts.slack.com/archives/C0XUF8XL1/p1592837760105000 we need to adjust billable hour expectations for almost everyone. 

Generally speaking, most everyone will be +4 higher, for 36 hours. Exceptions are noted here: https://docs.google.com/spreadsheets/d/1QkvP3IfwuUow9PmHR1gD_FCGYQ12NngOQB3FbVsrfgs/edit#gid=0

Jkrzy:
Let's create an updated sheet and use the command introduced in d15043a1ce39c9140f1fc972072390cda84463f0 to apply changes the week of July 6th.

tbaxter-18f:
Sheet is noted above. I'll post here when it's been fully reviewed.

Jkrzy:
If possible I'd like to run through an audit of 18F folks across the board as part of this effort. There's a few guild leads that I already see need corrections and backfills.

tbaxter-18f:
I could put that on the SMT. I'm already asking them to note _any_ exceptions in the spreadsheet above.

tbaxter-18f:
@Jkrzy I don't have final confirmation from Elizabeth, but I believe the sheet is done and ready to use now. Thanks for your additions to it!

Jkrzy:
Thanks! We'll roll this out the week of 7/6/2020!

Jkrzy:
Realizing now that the existing management command only effects changes to existing Timecards. We need to update billable_expectations on the `UserData` model here.

We'll need another approach

Jkrzy:
Given the short week and need to get this in place for next week, let's make these changes via the admin.

Each employee on this sheet needs to have their `Employee` billable expectation value confirmed and updated if necessary.
https://docs.google.com/spreadsheets/d/1QkvP3IfwuUow9PmHR1gD_FCGYQ12NngOQB3FbVsrfgs/edit#gid=32063559

From [this starting point (current 18F employees)](https://tock.18f.gov/admin/employees/userdata/?current_employee__exact=1&organization__id__exact=1), we can filter by business unit and divide the work amongst a few administrators.

Jkrzy:
As of today this backfill hasn't happened.

We'll now also need to update previous timecards for the weeks starting:
* Aug. 2, 2020
* July 26, 2020
* July 19, 2020
* July 12, 2020
* July 5, 2020

[...]

PR #1143, opened 2020-08-11
Enhance billable_expectation backfill management command

Jkrzy:
## Description

For #1119, we expand our management command previously used to backfill billing targets for existing timecards with the ability to update `UserData.billable_expectation`. Also renamed the command to `update_billable_expectations` which is both a bit shorter and I think clearer as to it's purpose.

Example call:
```bash
python manage.py update_billable_expectations  aug20-backfill.csv -u
```

With a csv file structured as below:
```csv
user.name, 36, 07/04/2020
```

Results in all timecards for reporting periods after 7/4/20, as well as UserData.billable_expectation, being set to 36 hours for each `username` provided in the csv.

### DIFF
--- a/tock/tock/management/commands/update_billable_expectations.py
+++ b/tock/tock/management/commands/update_billable_expectations.py
@@ -11,7 +11,7 @@ from hours.models import ReportingPeriod, Timecard
 
 @transaction.atomic
 class Command(BaseCommand):
-    help = 'Load CSV file of (username, target_hours, from_date) to bulk update timecard billing expectation values'
+    help = 'Load CSV file of (username, target_hours, from_date) to bulk update timecard, and if desired UserData, billing expectation values'
 
     def add_arguments(self, parser):
         parser.add_argument('csv_path', nargs='+', type=str)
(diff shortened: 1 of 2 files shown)


tock/tock/management/commands/update_billable_expectations.py
```diff
@@ -40,7 +47,15 @@ def handle(self, *args, **options):
                         timecards = Timecard.objects.filter(user__username=username, reporting_period__start_date__gte=start_date)
                         self.stdout.write(f'Updating {len(timecards)} timecards for {username} to target hours {target}')
 
+                        target_billable_expectation = self._get_target_billable_expectation(target)
                         # Update all identified timecards
-                        [self._update_timecard(timecard, float(target)) for timecard in timecards]
+                        [self._update_timecard(timecard, target_billable_expectation) for timecard in timecards]
+
+                        if update_user_data:
```
neilmb, line 54: I think it might be worth pulling this out into an `_update_userdata` function to parallel the existing `update_timecard`.

Jkrzy: Thanks for the nudge!!! Added a commit with this tweak

PR #1151, opened 2020-08-20
Utilization analytics by organization

neilmb:
## Description

This adds the capability to look at Tock data at the level of a single organization. This is especially valuable because utilization rate makes the most sense for the 18F organization without mixing in cloud.gov, login.gov, CoE, etc.

Screen Shot 2020-08-20 at 12 15 55

## Additional information

There is a layout re-design for the form that selects orgs and dates to make it less intrusive on the space available for the graphs.  There is also a pesky `NULL` organization in the oldest tock data that needed special handling.

### DIFF
--- a/tock/utilization/analytics.py
+++ b/tock/utilization/analytics.py
@@ -1,4 +1,3 @@
-from django.apps import apps
 from django.db.models import Count, F, Q, Sum
 
 import pandas as pd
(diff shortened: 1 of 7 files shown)


tock/utilization/analytics.py
```diff
@@ -82,29 +92,41 @@ def utilization_plot(data_frame):
...
+            x=1,
+        ),
     )
 
     plot_div = plot(fig, output_type="div", include_plotlyjs=False)
     return plot_div
 
 
-def utilization_data(start_date, end_date):
+def utilization_data(start_date, end_date, org_id):
```
Jkrzy, line 114: This works but it feels like a great opportunity to re-use some of the other code we have a Tock -- there's a few places where we take URLParameters and return a filtered (or not) Queryset of Timecards and it would be awesome to have a single source for that work.

Could we expand/re-use maybe `api.views.get_timecards`? https://github.com/18F/tock/blob/64710825fadd6b8ade3c3de1290114815798d35a/tock/api/views.py#L192

neilmb: I hear you on code re-use, but in fact `get_timecards` is actually `get_timecardobjects` which fetches rows for the items on timecards rather than filtering the individual timecards which is what we're using here. 

I can definitely DRY up the first part of the queries in `utiltization_data` and `headcount_data`. Can we pair for more detail on what else you might have in mind?

neilmb:
@Jkrzy I refactored the timecard filtering from `api/views.py` and hooked it up with the new utilization graphs. Can you re-review now?

Issue #1138, opened 2020-08-04
Use or Lose PermissionMixin


Jkrzy:
We should refactor our usage of `PermisisonMixin`, removing it or replacing with [LoginRequiredMixin](https://docs.djangoproject.com/en/2.2/topics/auth/default/#the-loginrequired-mixin) and defining `permissions_classes` where needed.

In most of our class based views we're inheriting from `tock.utils.PermissionMixin`.

A handful of those views do not then define a `permission_classes` attribute which results in the mixin having no effect. 

Yet another subset of that group then overrides the `dispatch` function to handle permission checking like looking for unauthenticated users. 

e.g. `tock.utilization.views.GroupUtilizationView`

_Originally posted by @Jkrzy in https://github.com/18F/tock/pull/1136/files/6e5cbf42e9d368baad6db6aa60a162e3569c672b_

tbaxter-18f:
We should probably think about writing a test to check for missing permission_classes, too. 

Or just rethink how we do perms altogether.

PR #1153, opened 2020-08-21
Fix our use of PermissionMixin

neilmb:
## Description

In  #1138  @jkrzy noticed that we were mis-using the `PermissionMixin` class with no `permission_classes` which restricts nothing. This corrects that by using `LoginRequiredMixin` in those cases and prevents it from happening again by making it an exception to use `PermissionMixin` without defining the `permission_classes` attribute.

## Additional information

We could go further with this and replace all of the correct uses of `PermissionMixin` with `permission_classes = (IsAuthenticated, )` by `LoginRequiredMixin` but I haven't done that here. I'm willing though if people would prefer it.


Jkrzy:
Thanks @neilmb, this is nice move in the right direction. I don't think we need to address `PermissionMixin with permission_classes = (IsAuthenticated, )` in this PR.

Issue #1044, opened 2020-04-14
Investigate jquery removal


tbaxter-18f:
We don't use jquery a lot, but we do use it, which leads to the eventual Snyk unhappiness as our version falls out of date. Let's look at dropping the dependency in favor of vanilla JS, but let's timebox our effort to a day or so. If it takes longer than that it's probably not worth it.

cryptofilegsa:
Consider jadudm proposal here: https://github.com/jadudm/replacing-jquery

mark-meyer:
That's an interesting proposal. I looks like jQuery is used pretty intensively throughout the front-end code, so the danger of creating a jQuery clone is real. As a small example, the constructor will need to be quite a bit more robust to handle the various kinds of arguments as seen in `populateHourTotals()` :

```
    $('.entries .entry').each(function (i, entry) {
      entry = $(entry);
      var project = $('.entry-project select', entry).val();
      var hours = $('.entry-amount input', entry).val();
      ...
```

I wonder how the work to create and maintain this would compare to introducing a front-end framework like Vue.js?

tbaxter-18f:
I remain unconvinced that we need any JS framework for lib for our modest JS needs. I think for the purposes of this issue I'd just like a good investigation of what it would take... maybe I'm wrong. The outcome of this could just be a raft of new issues to address it bit by bit, culminating in an eventual story that actually removes it.

jadudm:
I agree completely that the clone creation is real. And yes, I saw that example in the code... I assume that, if we went down this road, that we'd have a MiniQuery (pick a name... *NotActuallyJQuery*?) that did what we needed, and no more (e.g. it would cover the use cases that currently exist). I personally see a number of follow-ons that are necessary to trust it: possible tests for just the class, integration tests, etc.

My thinking was essentially "one way to remove JQuery from the dependencies would be to introduce a new, lighter abstraction but leave the front-end code intact." In other words, leave the syntax and semantics intact, which I assume would reduce the likelihood of introducing new errors into the codebase. But, I worry that it might encourage (tacitly/explicitly) the continued use of *NotActuallyJQuery*. That, long term, would probably be Bad<sup>TM</sup>.

So... it's a thought, but I don't have history with the project, and don't know whether the tradeoff of a (probably?) light lift now to eliminate a ding from Snyk is desirable vs., say, a framework shift.

Or, just rewriting all the $() statements in vanilla JS. But, there are a bunch of them.

tbaxter-18f:
[...]

PR #1156, opened 2020-08-25
Translate jQuery on timecard.js to vanilla Javascript

hbillings:
## Description
Biggest part of #1044. 

@jadudm did the majority of the heavy lifting for this one, and I finished up the jQuery/JS translation. I have some concerns and probable discussion topics, though, that I think are worth mentioning here.

- Do we need Cypress? @adunkman set up puppeteer tests, and I'm not clear on why we switched. I personally don't think Tock's interactivity is such that we really benefit from everything Cypress provides, but I'm happy to be convinced otherwise! If we keep it, we need to set it up so that it only builds on dev, and not on prod. 

- The way that we're traversing the DOM and listening for events is incredibly brittle. It's a little easier to reason about now that we're out of jQuery land, but we're doing things like dispatching synthetic events on page load to trip another event listener that trips a function, and traversing the entire parent DOM structure to populate hourly totals. It's pretty hairy to reason about. At the very least, I'd recommend taking the next step here, which would be refactoring the JS into something a bit more organized. But I'd be really, really interested in opinions about introducing something like React or Vue to handle at least the timecard entry page.

- Finally and perhaps most ironically, this doesn't mean we can get rid of jQuery. We're using jQuery's `datepicker` element on `reporting_period_form.html` and `user_detail.html`, we're still using Chosen (which is due to be swapped out), and Django's admin still relies on jQuery. I think this has been a valuable exercise because if we ever want to iterate on this code, it's now better documented and tested, but... we still got jQuery.


adunkman:
> Do we need Cypress? @adunkman set up puppeteer tests, and I'm not clear on why we switched. 

@jadudm will need to speak to Cypress — I recall him asking some questions about puppeteer in Slack a while back, but didn’t follow the solutioneering. 🚂

Sgtpluck:
Closing this in favor of #1232

PR #1157, opened 2020-08-26
Update python dependencies

Jkrzy:
## Description

Updating our python dependencies including the [latest Markdown](https://github.com/Python-Markdown/markdown/releases/tag/3.2.2) which resolves the import issue we're seeing with `importlib-metadata`

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -87,11 +87,11 @@
         },
         "djangorestframework": {
             "hashes": [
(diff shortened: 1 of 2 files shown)


Jkrzy:
Yep, `pipenv update`  per [our docs](https://github.com/18F/tock/blob/main/docs/dependency-management.md) and after reviewing I'm not realizing I didn't include `requirements.txt` here 🤦. Thanks for asking!

Issue #23, opened 2015-03-26
Autocomplete Projects


vzvenyach:

afeld:
Have implemented this a few times across 18F projects recently – happy to help. Also, might be useful to sort them by name, and/or doing grouped options (e.g. USCIS projects, general 18F stuff vs. client projects, etc.)

seanherron:
Merging into meta issue https://github.com/18F/tock/issues/97

NoahKunin:
woooo

and now @batemapf is going to hate me.

In order for autocomplete to work elegantly, we should never put strings together like `Ops/Acquisitions`. As a result, just starting to type `Acquisitions` won't bring up the proper category.

What's the process to change all of these to something like `Ops - Acquisitions`?

seanherron:
We could go a step further and divide them in to optgroups, so 18F Operations is a group under which we have multiple line items or something.

NoahKunin:
I just saw that functionality! Pretty slick. Makes sense. I just want to be sure it doesn't destroy all of 18F Finance's pivot tables...

PR #1161, opened 2020-08-31
Replace Chosen with accessible-autocomplete

### LINKED ISSUE #1109: Chosen.js and usability/accessibility
The chosen.js library was a significant stumbling point in front-end testing. It obscures/hides the original HTML `select` element, and replaces it with a javascript-y `div`. Doing some reading around the web, It is unclear from community discussions whether Chosen plays well with screenreaders (e.g. https://github.com/harvesthq/chosen/issues/264, https://stackoverflow.com/questions/45936179/accessibility-jaws-dropdown-list-with-jquery-chosen-library). 

Do we need it? Does it work against us (especially with regards to usability)? Should we consider pulling it from timecard.js/the time entry interface? 

It certainly added complexity for testing.

hbillings:
## Description

Closes #1109. Chosen.js is not accessible; replace it with [accessible-autocomplete](https://github.com/alphagov/accessible-autocomplete). 
Todo:
- [x] Tests.
- [x] Figure out dropdown styling.

Looks like this closed: 
Screen Shot 2020-08-31 at 2 22 36 PM

And this expanded:
Screen Shot 2020-08-31 at 2 22 21 PM

I'm wondering what we'd like to do about the dropdown. Currently, Tock's dropdown separates projects into categories, although I have to admit I'm a little unsure how it's doing that:
Screen Shot 2020-08-31 at 2 24 49 PM

I'm also not clear on whether this is a feature that people would miss if it disappeared. It _seems_ like it makes things more usable to me and like it's probably worth attempting to figure out with the new autocomplete, but I wanted to open this and get some opinions flowing.


neilmb:
I'd happily take the chosen replacement as-is and backlog a ticket to figure out categorization.

adunkman:
:wave: Resident Chosenista here. Chosen does that through [`<optgroup>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/optgroup) — not sure if that’s helpful, but figured I’d chime in! 😄

hbillings:
Okay, so it looks like the `optgroup`s get created by Django's form element on the back end, but `accessible-autocomplete` doesn't pick them up. I'm inclined to say let's ship it as-is and see if people complain about it. @Jkrzy @arouault what do you think about that?

arouault:
@hbillings Hmm - having no real information about why the groupings were created in the first place, it's hard to say My gut tells me categories exist because business units are important, ie, there was a situation where non-18F units were billing to 18F tock lines because they didn't know the distinction. I'd argue for keeping categories. Can we transfer over the drop down categorization in some fashion?

Jkrzy:
The groupings are derived from 'AccountingCode's and originated in PR #23, ~5 years ago. 

It also looks like accessible autocomplete doesn't have an existing path to categories. https://github.com/alphagov/accessible-autocomplete/issues/145

+1 to Neil's thought, we know we need to move away from Chosen. Let's sort out if/how we need to enhance that UI after it's in place.

hbillings:
@Jkrzy I like the option in that example of using the custom template option to add a hint to the text. That would probably be a pretty easy enhancement (whether it needs to go in this PR or not is another story). (I share your concern about business units tocking to the wrong lines, @arouault, and I'd like to try to mitigate that if possible! I think it might require a little creativity and trial-and-error, though.)

tbaxter-18f:
Tocking to the wrong line _should_ be difficult since we added the enhancement this year to only show tock lines for your org.

I know you can create Django choices with optgroups, but I don't know if the accessible autocomplete would work well with them or not.

I agree with the above. It's a feature of only limited benefit, and we can log a follow-up issue to bring it back if we want to.

Jkrzy:
@hbillings, let me know if there's a more convenient way the backend can be providing `things` for the new autocomplete!

hbillings:
I went back to write the test I thought this needed, and I don't actually think it needs one. We'd basically just be testing the Chosen library.

tock/tock/static/js/components/timecard.js
```diff
@@ -304,9 +309,18 @@ $(document).ready(function () {
...
+    accessibleAutocomplete.enhanceSelectElement({
+      showAllValues: true,
+      defaultValue: '',
+      selectElement: select
+    })
+  });
+
+  $('.entry-project input')
     .on('change', function (e) {
+      debugger;
```
adunkman, line 323: ✂️ Looks like a `debugger` snuck in! Tricky little bugger.

tock/tock/static/js/components/timecard.js
```diff
@@ -197,10 +197,10 @@ $("body").on("change", ".entry-project select", function () {
...
 
 $(document).ready(function () {
-  var chosenOptions = {
-    search_contains: true,
-    group_search: false
-  };
+  // var chosenOptions = {
+  // search_contains: true,
+  // group_search: false
+  // };
```
adunkman, line 203: ✂️  I think we can probably delete these commented lines including the ones above, if that works with you!

package.json
```diff
@@ -12,7 +12,7 @@
     "node": ">=8.10.0"
   },
   "dependencies": {
-    "chosen-js": "^1.8.7",
+    "accessible-autocomplete": "^2.0.3",
```
adunkman, line 15: Looks like we’re loading this from `vendor/` rather than `node_modules` — is this needed here?

adunkman (approved):
A few clean-up comments, but this looks good! Giving a preemptive 👍 assuming the comments are addressed.

package.json
tbaxter-18f: @hbillings thoughts?

tock/tock/static/js/components/timecard.js
```diff
@@ -164,7 +164,7 @@ function addTockLines(tockLines) {
     setTimeout(function () {
       // Set the project and trigger a GUI update
       $("div.entry:last .entry-project select").val(line.project);
-      $("div.entry:last .entry-project select").trigger("chosen:updated");
+      // $("div.entry:last .entry-project select").trigger("chosen:updated");
```
tbaxter-18f, line 167:
```suggestion
```

Jkrzy (approved):
Thanks @sgtpluck!

Issue #1027, opened 2020-04-09
Use preferred, friendly name


tbaxter-18f:
currently names are your legal name, in a john.doe format. It'd be nice if we accounted for preferred names.

It's possible we may just be able to modify the username for the preferred output, maybe splitting on the ., or it may require a new field.

tbaxter-18f:
I noticed today we already capture firstname and last name, so this is really just a display question, not a model question. 

I think we should start by scoping it simply to the utilization reports, since that's where names are most noticeable

Sgtpluck:
Closed with https://github.com/18F/tock/pull/1167

PR #1167, opened 2020-09-01
Use preferred names for templates in Tock

Jkrzy:
## Description

For https://github.com/18F/tock/issues/1027, we now render preferred names (First + Last) -- configurable via the admin panel -- for all pages in Tock.

This does not include changes to the API endpoints or data exports.

There's a handful of users (~9) who don't have a First/Last name value set -- We could handled with a bit more code here, rendering `username` as a fallback OR updating these users in the admin.

We should also update the Tock change form so folks can request a name update...OR create a form so folks can update it themselves


tock/utilization/unit.py
```diff
@@ -49,10 +49,10 @@ def _get_unit_billing_data(users, unit, recent_periods=None, fiscal_year=False):
         start, end, utilization = utilization_report(users, unit=unit)
 
     if utilization:
-        data = utilization.values('username', 'billable', 'target')
-
+        data = utilization.values('username', 'first_name', 'last_name', 'billable', 'target')
```
tbaxter-18f, line 52: Is there still a reason to put username in `data`?

Sgtpluck: should `username` be used as a fallback?

tock/employees/models.py
```diff
@@ -180,6 +179,15 @@ def is_late(self):
...
 
+    def display_name(self):
+        user = self.user
+        if user.first_name and user.last_name:
+            return user.first_name + " " + user.last_name
+        elif user.first_name:
+            return user.first_name
+        else:
+            return user.username   
+
```
tbaxter-18f, line 190: @Sgtpluck Django has an existing `get_full_name` method on the User object, so we could just sorta alias that: https://docs.djangoproject.com/en/3.1/ref/contrib/auth/#django.contrib.auth.models.User.get_full_name

Also, since it does entail a DB lookup for something that doesn't change often, Django's cached properties might make sense here: https://medium.com/@fdemmer/django-cached-property-on-models-f4673de33990

tock/tock/templates/employees/user_detail.html
```diff
@@ -4,7 +4,7 @@
 {% endblock %}
 
 {% block content %}
-  <h1>Let's Tock about {{ object.user.first_name }} {{ object.user.last_name }}!</h1>
+  <h1>Let's Tock about {{ object.user.user_data.display_name }}!</h1>
```
tbaxter-18f, line 7: Here, unless I'm mistaken, `object` is our `user_data` object, so if we add a convenience method to it, then we could just use that: `object.display_name` or whatever. On the other hand, if we're OK with following the django user object (which user_data extends) then we could use django's built-in `get_full_name` like this: `object.user. get_full_name`. 

Right now, if I'm reading it right, I think you're going from user_data up to user, then back down to user_data again.

tock/employees/models.py
```diff
@@ -180,6 +179,15 @@ def is_late(self):
         else:
             return False
 
+    def display_name(self):
```
Jkrzy, line 182: ❤️  handling this in our own 1:1 model -- We can get a little pythony here and use Django's [User.get_full_name](https://github.com/django/django/blob/2.2.17/django/contrib/auth/models.py#L341) to...

This will return any Truthy string returned by `get_full_name` otherwise we'll return the username.

```suggestion
    def display_name(self):
        return self.user.get_full_name() or self.user.username
```

tock/tock/templates/employees/user_form.html
```diff
@@ -4,7 +4,7 @@
 {% endblock %}
 
 {% block content %}
-<h1>Let's Tock about {{ user.get_full_name }}!</h1>
+<h1>Let's Tock about {{ display_name }}!</h1>
```
tbaxter-18f, line 7: I haven't looked at the view to confirm, but it looks like this page is passed a `user` object, so I don't think a change is needed and `user.get_full_name` should still be correct.

tock/utilization/unit.py
```diff
@@ -49,11 +49,10 @@ def _get_unit_billing_data(users, unit, recent_periods=None, fiscal_year=False):
         start, end, utilization = utilization_report(users, unit=unit)
 
     if utilization:
-        data = utilization.values('username', 'first_name', 'last_name', 'billable', 'target')
+        data = utilization.values('user_data', 'billable', 'target')
```
tbaxter-18f, line 52: the use of `values()` here was probably done to make this page quicker/more efficient, but it could complicate you getting the data you need on the page. I'm not sure about getting the whole `user_data` object in values() and if that will work like you want it to. This may help if there's a problem: https://docs.djangoproject.com/en/3.1/ref/models/querysets/#values

tock/employees/models.py
tbaxter-18f: @Sgtpluck in case it's confusing, @Jkrzy is suggesting pretty much the same thing I suggested up in the template :-)

Sgtpluck: oh dope! i was trying to figure out what `get_full_name` did; i was trying to figure out if django automagically made getters or something. good to know there are some built-in model methods!

tock/tock/templates/employees/user_detail.html
Sgtpluck: hah, awesome

tock/utilization/unit.py
Sgtpluck: oooh yes it grabs the ID of the object rather than the object itself, that documentation was extremely useful.

tock/employees/models.py
```diff
@@ -180,6 +179,11 @@ def is_late(self):
         else:
             return False
 
+    @cached_property
```
tbaxter-18f, line 182: I could be wrong, but I think you may have to import cached_property.

tock/employees/models.py
tbaxter-18f: Resolved by updates

tock/employees/models.py
Sgtpluck: it was already imported! `billable expectations` is also cached.

tock/projects/tests.py
```diff
@@ -494,7 +494,7 @@ def test_project_timeline(self):
         self.assertEqual(
             res['groups'],
             {
-                self.user: {
+                self.user.user_data.display_name: {
```
tbaxter-18f, line 497: in this case it may be better to use self.user.get_full_name() rather than traverse the DB to user_data

tock/projects/views.py
```diff
@@ -72,7 +71,7 @@ def project_timeline(project, period_limit=5):
                 break
             periods.append(report_date)
 
-        groups[tc.user][report_date] = float(t.hours_spent)
+        groups[tc.user.user_data.display_name][report_date] = float(t.hours_spent)
```
tbaxter-18f, line 74: as above, it *may* be better to use self.user.get_full_name() rather than traverse the DB to user_data

tock/tock/templates/employees/user_form.html
tbaxter-18f: Ah, we're passing display_name to the form?

Sgtpluck: i think passing the `display_name` through will be the most consistent experience -- though happy to change it if you disagree strongly!

tock/utilization/unit.py
```diff
@@ -49,11 +52,11 @@ def _get_unit_billing_data(users, unit, recent_periods=None, fiscal_year=False):
         start, end, utilization = utilization_report(users, unit=unit)
 
     if utilization:
-        data = utilization.values('username', 'billable', 'target')
+        data = utilization.values('user_data', 'billable', 'target')
 
         # Build context for each employee
         output_data = [{
-                'username': employee['username'],
+                'display_name': UserData.objects.get(id=employee['user_data']).display_name,
```
Sgtpluck, line 59: is adding this lookup here ok? i'm a little concerned about it if someone was using `values` to optimize the page as you mentioned @tbaxter-18f

tock/projects/views.py
Sgtpluck: hmmmm is putting this on `user_data` the wrong place? i guess i'm just concerned about those edge cases where there is no `full_name` and it'll just be blank.

tock/utilization/unit.py
tbaxter-18f: Hmmm. We may need to look at the Utilization object... is it storing username there? Is that OK? Will we have a mismatch? Not sure.

tbaxter-18f: if the ID itself is useful, I'd suggest getting it explicity, just for clarity... `values('id', 'billable', 'target)`

If you prefer, you could say primary key instead: `values('pk', billable', target')

It'll be the same, so whichever feels more right to you

tock/utilization/unit.py
tbaxter-18f: My concern is "Build context for each employee" I'm not sure how many employees we're talking about here, but that would be a DB lookup for each one.

This may be a thing we need to talk through more. @Jkrzy without diving in deep, it looks like we're keying the utilization table off of username. Thoughts?

tock/utilization/unit.py
Sgtpluck: Oh, sorry, I mean -- instead of returning all of `user_data` it returns the a key-value pair of `user_data: $USER_DATA_ID`

tock/utilization/unit.py
Jkrzy: In the aggregation we're already traversing UserData -> User and grabbing the components for `display_name`, these queries would do that work again. I'd rather use those pieces we already have, and duplicate a bit of code, to generate `display_name`. 


```suggestion
                'display_name': f"{employee['first_name']} {employee['last_name']}".strip() or employee['username'],
```

tock/projects/views.py
tbaxter-18f: I think we can handle that on the data entry side and simply say all employees must have first and last names populated (they do)

Jkrzy:
:amaze: Thanks @Sgtpluck! I added a one line tweak to avoid some extra queries AND I can't approve this since I'm the original author 😅. 

Take a look at https://github.com/18F/tock/pull/1167/commits/836cb80dce878549874aa9fea49be536f0c032b1 and 👍 👎  please. This change lets us render the `Out of Office` project with ~11 queries instead of ~350 (one for each employee)

tock/projects/views.py
tbaxter-18f: @Jkrzy is this where your select_related is helping us not have the extra query? I'm still feeling we should just use `user.get_full_name()` but if there's no DB hit it really doesn't matter.

Jkrzy: Yep! We stuck with handling cases where folks don't have a First/Last in the system

tbaxter-18f: Cool, cool. That shouldn't ever happen given how folks are onboarded, but if there's no DB hit, that's great.

tbaxter-18f (approved):
Big round of 👏  for @Sgtpluck. Great job, especially as you're still learning your way around Django.

Issue #1183, opened 2020-09-30
Update billable hour expectations for recent timecards


Jkrzy:
Employee (UserData) billable expectations were not updated as expected when #1119 was closed on 8/13/2020.

## Impact
* Upon creation, `Timecards` inherit their billable expectation value from `UserData.expected_billable_hours`
* Since `UserData.expected_billable_hours` was not updated in #1119, Timecards created after 8/9/2020 inherited the previous billable expectation value -- 32 hours in most cases.
* `Timecards` for August and September need to be updated to reflect billing targets in effect as of 7/1/2020
  * In most cases this means a updating each timecard to set the billable expectation to `36` where it is currently `32`
* Until corrected, Utilization reports in Tock will have an inflated `% billable` for timecards in August and September.

### Example
_|Current|Correct
-|-|-
Billable|36.00|36.0|
Non-Billable|4.00|4.00
Excluded from Utilization|0.00|0.00
Billable Expectation|0.80|0.90|
Target Hours|**32.00**|**36.00**
Utilization|**112%**|**100%**

## Origin story

`UserData.expected_billable_hours` was **not** updated as expected owing to a bug in the management command implemented in https://github.com/18F/tock/pull/1143. 

https://github.com/18F/tock/blob/cab64f1d16f1a84f18a4633b49b670897326d602/tock/tock/management/commands/update_billable_expectations.py#L34

The billable expectation field was changed in 442ba75fab2a77b5ae3472e548c7cc1d73c94a1a, so this _should_ be
```diff
- userdata.billable_expectation = target 
+ userdata.expected_billable_hours = target 
```

## How we'll fix it

- [x] Fix the `update_billable_expectation` management command
- [x] Review the previous used spreadsheet noting any changes since 8/13/2020, we'll need to address these cases separately
- [ ] Re-run the corrected `update_billable_expectation` with an updated input spreadsheet OR update via admin interface

Jkrzy:
We've completed an audit of billable expectations and assembled the input to re-run this command.

Jkrzy:
Executed successfully at ~12:00 EST today.

PR #1189, opened 2020-10-14
set billable hours, not billable expectation

timoballard:
## Description

Resolves a bug in the `update_billable_expectations` management command, as described in #1183.


Jkrzy (approved):
Looks good, thanks Tim!

PR #1210, opened 2021-01-21
Correct billable target updating management command and admin tweak

Jkrzy:
## Description

In https://github.com/18F/tock/pull/1189 we updated a management command to update `UserData.expected_billable_hours` but we did not _also_ update the value which was being saved there.

This PR changes the management command such that an integer is saved to `UserData.expected_billable_hours` as opposed to a decimal percentage of billable expectations. Allowing us to resolve https://github.com/18F/tock/issues/1183.

Also enables list-editing of billable targets in the UserAdmin list view.

### DIFF
--- a/tock/employees/admin.py
+++ b/tock/employees/admin.py
@@ -61,6 +61,7 @@ class UserDataAdmin(admin.ModelAdmin):
         'organization',
         'unit',
     )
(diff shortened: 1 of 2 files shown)


tbaxter-18f (approved):
nice!

PR #1212, opened 2021-01-25
Remove requirements.txt

Jkrzy:
## Description

Resolves #1211 by removing `requirements.txt` in favor of a single declaration of our python dependencies in `Pipfile.lock`

PR #1213, opened 2021-02-01
Revert #1212 and update python dependencies

Jkrzy:
## Description

In #1212 we optimistically removed requirements.txt but Snyk's github integration doesn't actually support Pipfile's just yet. https://support.snyk.io/hc/en-us/articles/360004699377-Snyk-for-Python

Reverting that change and again updating python dependencies, including `requirements.txt`, here.


Sgtpluck (approved):
boooo but 👍

Issue #1181, opened 2020-09-14
Session time-outs: need to allow for session extension (accessibility/508)


iamjolly:
When sessions time-outs occur, no warning or ability to extend a user's session is offered. This leads to usability and accessibility problems.

Per [WCAG 2.2.1 Timing Adjustable (Level A)](https://www.w3.org/TR/UNDERSTANDING-WCAG20/time-limits-required-behaviors.html):

For each time limit that is set by the content, at least one of the following is true: 
- **Turn off:** The user is allowed to turn off the time limit before encountering it; or
- **Adjust:** The user is allowed to adjust the time limit before encountering it over a wide range that is at least ten times the length of the default setting; or
- **Extend:** The user is warned before time expires and given at least 20 seconds to extend the time limit with a simple action (for example, "press the space bar"), and the user is allowed to extend the time limit at least ten times; or
- **Real-time Exception:** The time limit is a required part of a real-time event (for example, an auction), and no alternative to the time limit is possible; or
- **Essential Exception:** The time limit is essential and extending it would invalidate the activity; or
- **20 Hour Exception:** The time limit is longer than 20 hours.

This will also help users who do not require accessibility supports but may be frustrated by not realizing a session is "dead" and not being able to log new entries (losing any unsaved entries) due to needed to log back in. 

**Recommendation:** 
1. Prompt users with a modal dialog with a warning that lack of activity may require log-out and offer to allow them to extend their session. 
2. If a user is logged out for inactivity, redirect them to a logged-out page with a link to log back in.

Jkrzy:
Initial pass at this implemented in #1216!

Jkrzy:
As-is this uses the default alert dialog, we'll open an issue to iterate onward.

> Doing this with a USWDS confirmation modal will give us a bit more control and let us make this even more accessible. Replacing the alert dialog with something where we can better control the title and interpretation by a screenreader.
> 
> I'm hopeful there's some existing art we can borrow from another TTS site

PR #1216, opened 2021-02-03
Create timer so user can extend session if necessary

Sgtpluck:
## Description

Addresses #1181. Creates a little javascript timeout to give user a prompt to extend the session right before timeout.

## Additional information
I'm just using the built-in alert bc it was fastest/easiest, but it doesn't look very nice. Looks fairly straightforward to make a little warn template if that seems worth it.


tock/tock/urls.py
```diff
@@ -39,6 +39,9 @@ def check_if_staff(user):
 
     path('auth/', include('uaa_client.urls')),
 
+    path('session-warning', tock.views.session_warning, name="SessionWarning"),
```
Sgtpluck, line 42: this seemed like an okay place to put these endpoints but ... i dunno! is there a better one?

tock/tock/templates/base.html
```diff
@@ -37,6 +37,35 @@
...
+        var request = new XMLHttpRequest();
+        request.open("GET", "/session-extend", true);
+        request.onload = function () {
+          if (request.status === 401) {
+            alert("Your session has expired. Please log back in.");
+            logout();
+          }
+        };
+        request.send();
+        setTimeout(setAlert, timeoutCount - 120 * 1000);
```
Sgtpluck, line 51: if it succeeds, gotta set a new timer

tock/tock/templates/base.html
```diff
@@ -37,6 +37,35 @@
     ga('set', 'forceSSL', true);
     ga('send', 'pageview');
   </script>
+  <script>
```
Sgtpluck, line 40: i still don't like having this much logic in a template but trying to move it over to a JS file  led to some annoying race conditions.

tock/tock/templates/base.html
```diff
@@ -37,6 +37,35 @@
...
+      const timeoutCount = "{{x_tock_session_timeout_ms}}";
+
+      function setAlert() {
+        alert(
+          "Your session is about to expire! We will attempt to refresh your session."
+        );
+        extendSession();
+      }
+
+      setTimeout(setAlert, timeoutCount - 120 * 1000);
```
Jkrzy, line 67: Do we need to modify `timeoutCount` here? I think we're already rendering it on the template in milliseconds.

Jkrzy: Ah I see, we're firing the alert 2 minutes prior to the timeout.

tock/tock/templates/base.html
```diff
@@ -37,6 +37,35 @@
...
+
+      function logout() {
+        window.location.reload();
+      }
+
+      const timeoutCount = "{{x_tock_session_timeout_ms}}";
+
+      function setAlert() {
+        alert(
+          "Your session is about to expire! We will attempt to refresh your session."
```
Jkrzy, line 62: A small words tweak
```suggestion
          "Your session will expire in 2 minutes. Press OK or the enter key to extend your session."
```

Jkrzy (approved):
Looks good, thanks @Sgtpluck!!!!

Doing this with a USWDS confirmation  modal will give us a bit more control and let us make this even more accessible. Mind writing up a small issue for iterating on this by replacing the `alert` dialog with something where we can better control the title and interpretation by a screenreader.

I'm hopeful there's some existing art we can borrow from another TTS site

Issue #1217, opened 2021-02-03
Project level visualization of hours tocked v. time


Jkrzy:
Let's expand on the recent visualization work and create project level charts which show hours tocked per reporting period stratified by user. 

>I would LoveLoveLove :love-cry: if there were an analytics view that could show the hours over time per project or Tock line, possibly stacked in colors with the staff who billed to that Tock line.

### We'll know we're done when
- [x] We're rendering plotly area charts on a project specific page showing hours tocked per reporting period.
- [ ] We're rendering summary data tables below those charts
- [x] We've got an approach to handling projects for which the number of unique Tockers is large enough to reduce the utility of the chart.
-
An approximation of what this might end up looking like:

<img width="843" alt="Screen Shot 2021-02-03 at 11 04 52 AM" src="https://user-images.githubusercontent.com/3485564/106774314-fff94180-660f-11eb-8474-7ebd980d785a.png">

Jkrzy:
I started working on this a bit here https://github.com/18F/tock/pull/1221 but won't have time this week to wrap it up nicely and ready it for review/merging.  Free for anyone to pick-up if they're available/interested!

Jkrzy:
Resolved in #1221

PR #1221, opened 2021-02-04
Project level visualizations

Jkrzy:
## Description

For #1217, Expanding on the good works of NeilB to add project specific charting.

On the project details page, we now render area charts describing hours tocked by user and reporting period. This includes some tweaks to the details page to accommodate the new content.
A table containing the plotted data is also provided along with an export button. 

The table may extend horizontally quite a ways as the number of folks who've tocked to a specific project grows.


tock/tock/templates/employees/user_list.html
```diff
@@ -14,7 +14,6 @@ <h1>Tock Users</h1>
       <th>First Name</th>
       <th>Last Name</th>
       <th>Start Date</th>
-      <th>End Date</th>
```
Jkrzy, line 17: Removing an unused column

tock/utilization/views.py
```diff
@@ -58,29 +53,34 @@ def get_queryset(self):
...
+        context.update({'org_totals': org_billing_context()})
         return context
 
 
-class UtilizationAnalyticsView(LoginRequiredMixin, TemplateView):
-    template_name = "utilization/utilization_analytics.html"
-
-    def get_context_data(self, **kwargs):
-        context = super(UtilizationAnalyticsView, self).get_context_data(**kwargs)
+class FilteredAnalyticsView(LoginRequiredMixin):
```
neilmb, line 60: I don't see where this refactored class was used elsewhere. Was there a different version where this new super-class was useful? I don't have a problem with factoring this functionality out, but it doesn't seem related to the rest of this PR.

tock/tock/templates/projects/project_detail.html
```diff
@@ -1,99 +1,100 @@
...
-  <li> <b> Alerts: </b>
-{% if object.alerts.exists %}
-    {% for alert in object.alerts.all %}
-        {% if alert.destination_url %}
-          <a href="{{ alert.destination_url }}" target="_blank"><span class="{{ alert.full_style }}">{{ alert.full_alert_text }}</span></a>
-        {% else %}
+<div class="grid-container">
+  <div class="grid-row">
+    <div class="tablet:grid-col">
+      <ul>
```
neilmb, line 19: Semantically I wonder if this is a worthy use of a `<dl>`? Instead of styling field names manually with `<b>`, could use `<dt>` and `<dd>`. But maybe it's not worth the effort of styling that infrequently used element?

neilmb (approved):
Glad that the original analytics stuff is proving to be reusable. LGTM apart from quibbles.

tock/utilization/views.py
Jkrzy: Thanks @neilmb, that's definitely unrelated to the functionality and a remnant I didn't completely walk back.

My[ original approach](https://github.com/18F/tock/pull/1221/commits/ca2d2cfc8703789d7d3f8fa5fe8c4a847787b6da#diff-cdf622b15efe20a10cd3c1e46afc2d717720d093cbc6b11dda8028c24e0a3c24R165) had the project charts on another `utilization/` page and the associated view alongside this.  When I switched gears to handling it in the projects app I left `FilteredAnalyticsView`.

We'll leave it in but I should have called it out specifically or otherwise separated from the rest of these changes.

tock/tock/templates/projects/project_detail.html
Jkrzy: Likely and I'm definitely guilty here of not reworking the vast majority of the original html from `project_detail.html` and only made the changes necessary to get the charts rendering well-enough.

I'll file another issue to look at cleaning this up a bit, thanks @neilmb!

PR #1223, opened 2021-02-08
[WIP] Fire change event on note inputs

### LINKED ISSUE #1222: Accessible-Autocomplete round #2 -- Notes fields not rendering on timecard form
Notes fields, required inputs for certain projects, are not being revealed as expected after the release of  #1161 in which we moved to a new autocomplete implementation. We reverted these changes in #1224. 

We'll revisit and continue implementation of #1161 in this issue.

### We'll know we're done when
- [ ] Notes fields are revealed when a project, having notes required/displayed, is selected in the autocomplete.
- [ ] We've written a test to confirm the above behavior
- [ ] Notes fields are rendered as expected for project lines added by the user after initial page load 

**Notes from original triage**
On an initial look, we're not triggering a `change` event on the new `input` elements. https://github.com/18F/tock/blob/3081ab99b40dded0d7fde89742a51995bd6a0723/tock/tock/static/js/components/timecard.js#L323

It also looks like the behavior is a bit wonky when adding new project lines to a timecard. The enhanced select elements aren't displaying notes as we would expect and their `id` attributes are appending new `-select` strings on each, e.g. `id_timecardobjects-7-project-select-select-select`

Jkrzy:
## Description
Resolves #1222 by firing `change` events on `input` elements instead of `select`.

### DIFF
--- a/tock/tock/static/js/components/timecard.js
+++ b/tock/tock/static/js/components/timecard.js
@@ -320,7 +320,7 @@ $(document).ready(function () {
 
   // Force an update to each project selection menu in case a notes field
   // needs to be re-displayed.
-  $('.entry-project select').trigger('change');
+  $('.entry-project input').trigger('change');
 
   // Disable scrolling in numeric input form fields from the mouse
   // wheel or touchpad.


Sgtpluck (approved):

Jkrzy:
Thanks @Sgtpluck, I'm not totally convinced this is the right fix yet and I'd love to expand test coverage around this so we know it works.

Jkrzy:
Remedying this is a bit more complicated than the one liner proposed here. Going to close this and reassess.

PR #1225, opened 2021-02-11
Complete the transition from chosen to accessibleAutocomplete

Sgtpluck:
This addresses #1222 which was opened because of some bugs that were discovered in the port away from chosen -> accessibleAutocomplete

There are some complicated interactions between the original select element and the new accessibleAutocomplete that created a race condition that interfered with our ability to determine if notes needed to be displayed. This solution takes advantage of a provided onConfirm event to find the correct value and select it before calculating the notes visibility.

I also added a little bit of test documentation, and added some test for the behavior. (Two of them are currently being skipped because I'm unsure how to seed data. Help for that would be great!)


jest-puppeteer.config.js
```diff
@@ -1,6 +1,10 @@
...
-  // server: {
-  //   command: `./node_modules/.bin/serve --listen ${process.env.JEST_PORT} --no-clipboard ./dist`,
+  port: parseInt(process.env.JEST_PORT, 10),
+  // the launch options are helpful for debugging purposes
+
+  // launch: {
+  //   headless: false,
+  //   devtools: true,
   // },
+  browserContext: "incognito"
```
Sgtpluck, line 9: i think it's good for tests not to share context with other tests! it would be nice if we could give it its own server maybe to not share context with local dev but ... that's a change for another day

tock/test_data/data-update-deduped.json
```diff
@@ -222360,5 +222360,24 @@
...
+    "pk": 124,
+    "fields": {
+      "name": "notes Project",
+      "mbnumber": "",
+      "accounting_code": 16,
+      "description": "notes Project description",
+      "start_date": null,
+      "end_date": null,
+      "active": true,
+      "notes_required": true,
```
Sgtpluck, line 222374: added for testing purposes!

docs/local-development.md
```diff
@@ -96,6 +96,16 @@ docker-compose run app bash
...
 
+You can currently run the javascript tests locally by installing local dependencies and then
+running the tests. (We will dockerize these at some point!)
+
+```
+npm install
+npm run test
+```
+
+If you need help debugging the jest-puppeteer tests, turning off the `headless` mode in [the puppeteer configuration file](jest-puppeteer.config.js)
```
tbaxter-18f, line 107:
```suggestion
If you need help debugging the jest-puppeteer tests, try turning off the `headless` mode in [the puppeteer configuration file](jest-puppeteer.config.js)
```

Jkrzy (approved):
Looks great @Sgtpluck, thanks for digging into this and making Tock's garden of javascript a bit prettier 🌳 🌷 🌹 🌷 🌳

PR #1226, opened 2021-02-12
Vendor uswds icons and use relative paths for clock icons

Jkrzy:
## Description

We're missing a few icons on page render, this brings USWDS's `usa-icons` into the assets Tock makes available and updates paths to our friendly clocks.


lauraGgit:
closing no longer needed with #1408

PR #1232, opened 2021-02-18
Remove all jquery from timecard.js

Sgtpluck:
## Description

This builds on the work of #1156, which was close but not quite there in terms of rooting out all the jquery on timecard.js. I looked briefly into just updating that PR, but with the accessibleAutocomplete change, it had drifted enough away from main that that would have been more difficult. Thanks for the work @hbillings and @jadudm 

This change:
1. Removes all code related to localStorage. Very happy to discuss this if anyone has feelings about it; Conversations with @Jkrzy have led me to believe that the cost in reasoning and brittleness is far greater than the benefits we're currently getting.
2. Removes all jquery in timecard.js
3. Reorganizes the code somewhat.
4. Adds JSDOC documentation to make the existing functions a bit more understandable.
5. Moves the test file into an integration directory, leaving open the option for future unit tests. (I did not export any tests, though, so I didn't make any other test changes.

Open Considerations:
1. I didn't add more tests. I could do that! It will take a bit longer, as puppeteer is very fragile.

### DIFF
--- a/tock/tock/static/js/components/timecard.js
+++ b/tock/tock/static/js/components/timecard.js
@@ -15,7 +15,7 @@
  * */
 function getFormData() {
   return Array.from(document.querySelectorAll('.entry')).map((entry, i) => {
-    const markedForDeletion = $('.entry-delete input', entry).prop('checked');
+    const markedForDeletion = entry.querySelector('.entry-delete input').checked
 
     if (markedForDeletion) {
       return;
@@ -279,8 +279,9 @@ function addEntry() {
   });
 
   // Increment the TOTAL_FORMS
-  // FIXME
-  // Why is this needed? the post breaks if it's removed.
+  // This field is necessary for the ManagementForm template we are using.
+  // For more information, see:
+  // https://docs.djangoproject.com/en/2.2/topics/forms/formsets/#understanding-the-managementform
   const tf = parseInt(
     document.querySelector('#id_timecardobjects-TOTAL_FORMS').value
   );
--- a/tock/tock/static/js/components/timecard.js
+++ b/tock/tock/static/js/components/timecard.js
@@ -1,49 +1,93 @@
-function clearLocalStorage() {
-  // Clear anything saved locally.
-  if (window.localStorage) {
-    window.localStorage.removeItem(`tock-entered-hours-${objectId}`);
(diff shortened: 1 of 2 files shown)


tock/tock/static/js/components/timecard.js
```diff
@@ -1,336 +1,391 @@
...
+  accessibleAutocomplete.enhanceSelectElement({
+    showAllValues: true,
+    defaultValue: '',
+    selectElement: newEntry.querySelector('select'),
+    onConfirm: handleConfirm,
+  });
+
+  // Increment the TOTAL_FORMS
+  // FIXME
+  // Why is this needed? the post breaks if it's removed.
```
Jkrzy, line 283: Good question! It's a component of Django's Formsets which we're using to process the inbound N "Django form for a timecard entry". Specifically the management form. https://docs.djangoproject.com/en/2.2/topics/forms/formsets/#understanding-the-managementform

If we _don't_ update this when adding adding project lines, the response behavior get's a bit wonky when there are validation failures. 

I.e. `id_timecardobjects-TOTAL_FORMS = 2`. We submit a timecard with 3 lines that fails validation.  Only the first two project lines will be rendered in the response, the third will be lost.

Sgtpluck: ah, thank you for the explanation! i will update this documenting that so no one else has to scratch their head confused.

tock/tock/static/js/components/timecard.js
```diff
@@ -1,336 +1,391 @@
...
+/** @function
+ * Pulls data from the form
+ * @name getFormData
+ * @returns {FormData}
+ *
+ * */
 function getFormData() {
-  return $('.entry').map((i, entry) => {
+  return Array.from(document.querySelectorAll('.entry')).map((entry, i) => {
     const markedForDeletion = $('.entry-delete input', entry).prop('checked');
```
Jkrzy, line 18: Oooo, should this be a selector?

Sgtpluck: whoopsie, thank you for catching.

Jkrzy (approved):
:amaze: This is amazing, thanks @Sgtpluck!

PR #1234, opened 2021-02-19
Add username to utilization templates

### LINKED ISSUE #1233: Employee detail links broken on utilization reports
From slack:
>I noticed today that clicking on anybody's name from tock.18f.gov/utilization/ gives a server error:

### We'll know we're done when:
- [ ] valid EmployeeDetail links are rendering on the report page.

Jkrzy:
## Description

Resolves #1233 by adding `username` back to the report template context so it can be used to render the EmployeeDetail link. 

We're passing raw python structures to the template so we're building the urls in the template as opposed to using Django convenience methods like `.get_absolute_url`

### DIFF
--- a/tock/tock/templates/utilization/includes/unit_utilization_table.html
+++ b/tock/tock/templates/utilization/includes/unit_utilization_table.html
@@ -21,7 +21,7 @@
         {% for employee in unit.utilization.last_week_data %}
         <tr>
             <td>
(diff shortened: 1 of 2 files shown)


neilmb (approved):

Sgtpluck (approved):
oops!

Sgtpluck (approved):
oops!

PR #1239, opened 2021-03-01
fix JS error but there is a deletion ordering problem

Sgtpluck:
## Description

Fixes [immediate JS problem with delete](https://gsa-tts.slack.com/archives/C1JFYCX3P/p1614376281000900), but there seems to be an ordering issue where the assumption is that the deleted line is the bottom-most one.

### DIFF
--- a/tock/tock/static/js/components/timecard.js
+++ b/tock/tock/static/js/components/timecard.js
@@ -15,7 +15,7 @@
  * */
 function getFormData() {
   let data = []
-  
+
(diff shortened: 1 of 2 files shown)


Jkrzy:
Thanks for the quick fix @Sgtpluck! To resolve the delete ordering behavior we saw this morning I reverted back to the only-increment approach for the django management form so the backend will handle all inbound form data.

Jkrzy (approved):

PR #1242, opened 2021-03-02
Clear delete inputs for newly added tock lines

Jkrzy:
## Description

When a new project line is added to a timecard the delete input must be unchecked.

### Additional details
My test here feels not great, would appreciate any recommendations to simplify where possible!

### DIFF
--- a/tock/tock/static/js/components/timecard.js
+++ b/tock/tock/static/js/components/timecard.js
@@ -232,16 +232,13 @@ function addEntry() {
   // remove display: none style
   newEntry.querySelector('select').style.display = '';
 
-  let fields = newEntry.querySelectorAll('input, select, textarea');
-  // remove values
-  for (f of fields) {
-    if (f.matches('[type="checkbox"]')) {
-      f.checked = false;
-    }
-    else {
+  let non_checkbox_fields = newEntry.querySelectorAll('input:not([type="checkbox"]), select, textarea');
+  // remove values for fields we've cloned
+  for (f of non_checkbox_fields) {
       f.value = '';
-    }
   }
+  // Uncheck the cloned Delete input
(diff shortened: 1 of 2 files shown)


tock/tock/static/js/components/timecard.js
```diff
@@ -235,7 +235,12 @@ function addEntry() {
   let fields = newEntry.querySelectorAll('input, select, textarea');
   // remove values
   for (f of fields) {
-    f.value = '';
+    if (f.matches('[type="checkbox"]')) {
```
Sgtpluck, line 238: it might be nicer to just actively select the delete input and set it to false?

```
fields.querySelector('.entry-delete input').checked = false
```

Sgtpluck (approved):
had a lil suggestion for simplifying the code, but i think the test actually (unfortunately) looks good -- it's def a bit of a smell how much wiggling we have to do to test, but i think for our current setup it seems reasonable

tock/tock/static/js/components/timecard.js
Jkrzy: My thought was that in either case we needed excluded the checkbox input from `fields` when setting `value=''` and this was easier in the moment than excluding checkboxes from the querySelectorAll above 😅 

I'm not too attached to either implementation but this works as well --What do you think 👍  👎 ❓

```javascript
  let non_checkbox_fields = newEntry.querySelectorAll('input:not([type="checkbox"]), select, textarea');
  // remove values
  for (f of non_checkbox_fields) {
      f.value = '';
  }
  // Delete input must be unchecked
  newEntry.querySelector('.entry-delete input').checked = false
```

Jkrzy: Went with the above approach to handle checkboxes independently of the other fields we're cloning. Thanks @Sgtpluck!

Issue #1243, opened 2021-03-05
View of active projects sorted by engagement management hours billed


neilmb:
There is a hypothesis that hours being billed to the EM line for a project could be useful as an indicator of potential problems on projects. Tock could provide a page that lists the active projects and sorts them in descending order of the number of engagement management hours over some past unit of time.

Because projects with more staff MIGHT have more EM hours as a matter of course, we should check the trend there and consider sorting instead by the number of EM hours divided by the number of people billing against the project.

There is an open question of how to use the text descriptions of the individual tock lines to make a connection between a project's tock line and its corresponding EM line. In a perfect world, we could add a new field to link between the two lines, but text comparison might get us close enough for now.

@kfoley-18F

neilmb:
<img width="1151" alt="Screen Shot 2021-03-10 at 09 46 12" src="https://user-images.githubusercontent.com/443389/110661183-ece70d80-8189-11eb-872c-d2e9daf2ecb8.png">

WIP in branch https://github.com/18F/tock/tree/nmb/1243-engagement

PR #1250, opened 2021-03-10
Project view by Engagement Management lines

neilmb:
## Description

This addresses #1243 by adding a new view at `projects/engagement`. It lists only active projects that have "matching" engagement management lines where matching means the engagement management line ends with "Engagement Management" and it starts with the full name of another project line. (This does miss some pairs where they don't use exactly the same names.) 

The total number of hours billed for the project and EM line are computed and the projects are listed in descending order of the ratio of those hours (more relative engagement management time is listed first).

## Additional information

Screen Shot 2021-03-10 at 10 47 24


Jkrzy (changes requested):
Thanks @neilmb this is a super interesting hypothesis to test!

We're missing a navigable path to these reports from the interface, let's add a link from the `View Reports` drop down.

Jkrzy:
After this rolls out, let's set a check-in point a month or so into the future to evaluate if this is an informative way to look at project tocking. At that time we'll chat with the COO and make a decision on a next iteration or retiring this.

kfoley-18F:
I would like to get this merged soon. @neilmb - do you have time to add a nav link? Joe had suggested to the View Reports dropdown, I can also see making View Projects a dropdown instead of a static menu item, and having "All Projects" go to the current view and "Project reports" go to this one. (or something similar...) If you are buried, let me know and I'll get some other dev help for picking up this item and getting it over the line. 

Many thanks for this!!!

neilmb:
Added a second-level menu for "View Projects" and used the names "All Projects" for `/projects/` and "Engagement Levels" for `/projects/engagement/`. Happy to change the names if they are not clear.

Screen Shot 2021-07-28 at 09 33 10

neilmb:
@kfoley-18F Changed as requested, tests finally pass, ready for re-review.

kfoley-18F (approved):
W00t! thank you!!
LGTM

kfoley-18F:
@Jkrzy - does this satisfy your change request??

Jkrzy (approved):

Issue #824, opened 2018-09-20
Develop :grateful-tock: bot that recognizes on-time submissions


tbaxter-18f:
We already have angry-tock reporting late submitters, but @pburkholder suggested (and the motion quickly carried) that it would be nice to also have a happy tock that recognizes and thanks people who have consistently submitted to Tock on time over X time period.

We'll need to figure out X. Peter suggested 26 weeks. I think that may be a bit ambitious and we might want to try a shorter period. We could also consider having award levels or something for larger periods.

jduss4:
I took a first pass at this.  Rather than hardcode "X" to be a set number of weeks, instead the API user requests the number of previous reporting periods back that they would like, so this should be flexible.

Jkrzy:
Thanks @jduss4! https://github.com/18F/tock/pull/1255 expands Tock to track and provide the data necessary to recognize on-time submissions. Would you recommend we handle acting on that data in this issue or should we create another?

I can imagine us rendering some-sort of recognition in the Tock UI itself AND/OR  proposing a change to https://github.com/18F/charlie/, piping that recognition into Slack.

jduss4:
Good question.  I think if I were a supervisor or admin, I would prefer to have a report in Tock that I could look at for quarterly / before annual reviews.  However, I have never been a supervisor / admin, so I am not really the right person to ask about their preferences!

In general my impression is that Charlie seems like a good vehicle for constant checkins, like the weekly angry tocks. Since grateful tock would be seldom used, or at least not something that would be getting sent as a weekly notification, I don't know how helpful the integration would be.  Unless if we could set Charlie to fire off gratitude once a quarter?  The reminder might be nice for those who wouldn't otherwise remember to check Tock for a report.

If we decide not to do anything with Charlie and want to just develop some periodic reports based on employee tock-ins, then I vote we keep this issue open while we finish that off.  If we are doing something with Charlie then this issue will have accomplished creating the API endpoint for the Tock side of things and we can open an issue for Charlie instead.

Jkrzy:
@tbaxter-18f, it's been awhile since this one was opened by you but, want to weigh in? 😄

PR #1255, opened 2021-03-23
Grateful tock

jduss4:
## Description

A first stab at creating an API endpoint for "grateful tock," discussed in https://github.com/18F/tock/issues/824

- adds a field for `timecard.submitted_date` in order to determine if a timecard was submitted on time or late
- creates `api/submissions/NUM.json` endpoint to count punctual timecards by user
  - all counts are reported back for those timecards which were on time, not just those users with a 100% score

## Additional information

Several questions should be addressed with this change:

- should we automatically populate the `submitted_time` for timecards prior to this change?
- what happens if someone is on vacation / out sick and does not submit before the reporting period ends?

Furthermore, we should consider what behavior we want here vs in Charlie. For example, right now if you requested "5" reporting periods ago, you might get some users with "4" punctual timecards and others with "3" or fewer -- do we want to display the "4" or only display those with 100% timeliness?  Would we prefer that this was filtered on Charlie's side for reporting?


jduss4:
NOTE:  I am not a Python nor a Django developer and likely there are syntax / style / query construction faux pas that should be altered in these commits, so please feel free to bring the critiques

Jkrzy:
Thanks @jduss4! are you still working on this or is it ready for 😻 ? I see it's still marked as a draft.

jduss4:
Hi @Jkrzy I suppose I left it in draft because I wasn't clear how well what I have done so far fit the requirements.  Specifically I don't know what to do about these two things:

- should we automatically populate the submitted_time for timecards prior to this change?
- what happens if someone is on vacation / out sick and does not submit before the reporting period ends?

Depending on the answers to those things, I may need to make some alterations.  Otherwise the code is probably reviewable now and should also be functional if you want to test anything out!  If testing manually, you will need to go into the admin interface for timecards and set some of the `submitted_date` fields to be during the respective reporting periods.

Jkrzy:
Let's make an issue for our friend [Charlie](https://github.com/18F/charlie/issues) to consume this new endpoint and work some magic in Slack.

Jkrzy:
> Should we automatically populate the submitted_time for timecards prior to this change?

In practice timecards are not commonly modified after submission so setting the value to `Timecard.modified` would could approximate submission date time but I don't think it's necessary. Let's only track this going forward to have the data be the most reliable.

 > What happens if someone is on vacation / out sick and does not submit before the reporting period ends?

We don't have any plans to track `why` a timecard was submitted before/after the end of a reporting period. Unless they Tocked in advance or let the folks in #tock know to tock on their behalf. That's OK.

tock/hours/models.py
```diff
@@ -218,14 +218,21 @@ def save(self, *args, **kwargs):
...
         A reporting period is complete if it's end date
         has is today or in the past.
         """
         today = datetime.date.today()
-        return ReportingPeriod.objects.filter(end_date__lte=today).order_by('-start_date')[:number_of_periods]
+        periods = ReportingPeriod.objects.filter(end_date__lte=today)
+
+        if number_of_results:
+            range_end = number_of_periods + number_of_results
+            return periods.order_by('-start_date')[number_of_periods:range_end]
```
jduss4, line 232: I just realized that I think I need to subtract from the `number_of_periods` because of how the Django limit / offset logic works.

`[:5]` => 5th result
`[5:10]` => *6th* through 10th result

So if the user puts in "5 weeks" I think this will need to be `[4:5]` to catch the 5th result.

jduss4:
Now accepting feedback about if you want me to squash some commits :)

tock/api/urls.py
```diff
@@ -14,6 +14,11 @@
         ReportingPeriodAudit.as_view(),
         name='ReportingPeriodAudit'
     ),
+    re_path(
+        r'^submissions/(?P<num_past_reporting_periods>\d+).json$',
```
Jkrzy, line 18: Opportunity here to borrow the approach being used just above this for the ProjectInstanceView and avoid needing regular expressions. We'll also be able to skip casting the inbound value to an integer ourselves in the view.
```suggestion
    path(
        'submissions/<int:num_past_reporting_periods>.json',
```

tock/api/views.py
```diff
@@ -171,6 +180,33 @@ def get_queryset(self):
...
+
+    serializer_class = SubmissionSerializer
+
+    def get_queryset(self):
+        rp_num = int(self.kwargs['num_past_reporting_periods'])
+
+        reporting_period = ReportingPeriod.get_most_recent_periods(
+            number_of_periods=rp_num,
+            number_of_results=1
+        ).get()
```
Jkrzy, line 197: Grabbing the earliest reporting period may be a bit simpler here if we used negative indexing to grab the `last` object in the list instead of modifying `get_most_recent_periods`. This works by first evaluating the QuerySet to a plain old `list` so we can use negative indexing. 


```suggestion
        earliest_reporting_period = list(ReportingPeriod.get_most_recent_periods(
            number_of_periods=rp_num,
        ))[-1]
```

tock/api/views.py
```diff
@@ -284,6 +320,19 @@ def get_timecardobjects(queryset, params={}):
 
     return queryset
 
+def get_user_timecard_count(queryset):
+    """
+    Get a list of users and the number of the timecards
+    from a queryset of timecards passed in
+    """
```
Jkrzy, line 327: Love this approach!!

tock/api/views.py
```diff
@@ -171,6 +180,33 @@ def get_queryset(self):
...
+class Submissions(generics.ListAPIView):
+    """
+    Returns a list of users and the number of timecards they have
+    submitted on time since the requested reporting period
+    """
+
+    serializer_class = SubmissionSerializer
+
+    def get_queryset(self):
+        rp_num = int(self.kwargs['num_past_reporting_periods'])
```
Jkrzy, line 192: A bit of boundary handling would be a great next step here, as-is this throws an exception if someone attempts to ask "How many timecards, since X timecards ago have been submitted on time" for values of X greater than we have reporting periods.

tock/hours/models.py
```diff
@@ -218,14 +218,22 @@ def save(self, *args, **kwargs):
...
         has is today or in the past.
         """
         today = datetime.date.today()
-        return ReportingPeriod.objects.filter(end_date__lte=today).order_by('-start_date')[:number_of_periods]
+        periods = ReportingPeriod.objects.filter(end_date__lte=today)
+
+        if number_of_periods and number_of_results:
+            starting_index = number_of_periods - 1
+            range_end = starting_index + number_of_results
+            return periods.order_by('-start_date')[starting_index:range_end]
```
Jkrzy, line 233: Noted above but this will break for any input where the `starting_index` is larger than the number of reporting periods in the system.

Jkrzy (changes requested):
:amaze: @jduss4 Thanks so much for putting this together, really appreciate the docs and test coverage.

I left a few comments throughout where I saw potential opportunities to simplify things a bit. Would love your feedback on those. Overall this looks great!

api-docs/README.md
```diff
@@ -31,6 +31,7 @@ To access similar data in CSV format within Tock, please visit the
 - [Projects](projects.md): Fetches information about various projects.
 - [Reporting period (audit specific time period)](reporting-period-audit-specific.md): Fetches a list of all users who have not submitted a timecard for the specified period.
 - [Reporting periods](reporting-period-audit.md): Fetches a list of all available reporting periods and basic information about them.
+- [Submissions by range](submissions.md): Fetches a list of users and a count of punctual timecard submission by past periods
```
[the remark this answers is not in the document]
jduss4: "and a count of timecards submitted on time by past reporting period?"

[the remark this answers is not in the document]
Jkrzy: I was thinking it would be helpful for folks to know what Tock itself is considering as `on-time` or `punctual` here. Even just a plain-language description of this implementation would be helpful as I don't think there's an existing and consistently shared understanding of a date/time by which a timecard must be submitted.

tock/api/urls.py
jduss4: This is way better, thanks!

api-docs/README.md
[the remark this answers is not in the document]
jduss4: I added a bit to clarify when a timecard is considered "on time" as added in this PR (if my assumptions are wrong then we should change some code!)

tock/api/views.py
jduss4: This is definitely simpler. I went with this suggestion and removed the logic I added to the reporting period model for `number_of_results`

tock/api/views.py
jduss4: Since switching to the `list()[-1]` suggestion, I believe there should no longer be any problems with bigger-than-reality requests

tock/hours/models.py
jduss4: I believe this is resolved as well

jduss4:
I have one question before this gets merged:  what is the difference between a user and user_data?  I wasn't clear on whether I should be using `user.id` or `user_data` for the JSON response.

https://github.com/18F/tock/pull/1255/files#diff-a8868cbdadbfa18685d8ea42750df6024f25dc7ef80b5421320cd208d0759361R73

Jkrzy:
Great question, and the answer is :django-pony: related! 

Within Tock, `user` refers to [django's standard `User` model](https://docs.djangoproject.com/en/2.2/ref/contrib/auth/#user-model)

`user_data` refers to `tock.employees.models.UserData`, a model defined within Tock that holds "extra" attributes we want to associate with the `User` model. It's a `OneToOne` relationship in Tock and most commonly accessed with syntax like  -- To get an employee's start date -- `user.user_data.start_date`.

In this specific issue, `user.id` is almost certainly fine until we recognize otherwise, though renaming the field to `user_id` may make it a bit more clear.

jduss4:
I see!  Thanks, that's good to know.  I can change the field if you think that would be best.  I believe I was imitating `user` from the `TimecardSerializer` but have no particular attachment to it

Jkrzy (approved):
Thanks @jduss4!

PR #1258, opened 2021-03-23
Update timecard prefilling

### LINKED ISSUE #1257: Prefill data in excess of number of previous observed projects are not rendered
While investigating #1256 we found that not all prefill data is added to the initial timecard form when a user's number of active `TimecardPrefillData` instances exceeds the number of projects which were present on the previous timecard.

This originates [here](https://github.com/18F/tock/blob/a6c27e54bb61652a8b5186168c9475643fc9688b/tock/hours/views.py#L550) where we're setting the number of "empty" forms on the timecard page based on the previous timecard. We should be accounting for the total set of prefill data to include `holiday_prefills` and `TimecardPrefillData` instances.

Jkrzy:
## Description

Resolves #1257

For new timecards, we'll now include project lines as follows:

- All previously encountered projects
- All active prefill data configured in the admin
- An empty project line

Previously, if there were active prefills in excess of the number of projects tocked to, they would not be rendered as expected.


tock/hours/views.py
```diff
@@ -583,8 +578,12 @@ def prefilled_formset(self):
             for project_id, hours in timecard_prefills.items()
         ]
 
-        formset = timecard_formset_factory(extra=extra)
-        return formset(initial=init)
+        # Render all of our initial forms and an extra blank one
+        extra = len(init) + 1
+
+        FormSet = timecard_formset_factory(extra=extra)
```
Sgtpluck, line 584: python ignorance question: what's the repercussion (if any?) of setting this is a class rather than just a variable?

Sgtpluck (approved):
i think it looks good, but i have a question (for my own edification)

tock/hours/views.py
Jkrzy: Good question!

Functionally, thanks python, there's no difference between us calling this `FormSet` or `formset`. 

I changed to `FormSet` here to be a bit more idiomatic knowing a few Tock/Django details:
* we know that `timecard_formset_factory` is a passthru to `inlineformset_factory`
* We know that `inlineformset_factory` and [formset factories in general, return classes](https://docs.djangoproject.com/en/2.2/ref/forms/formsets/#django.forms.formsets.formset_factory)

tbaxter-18f: This _might_ help https://docs.djangoproject.com/en/3.1/topics/forms/formsets/ Or not. Cause django class-based-views can get hairy.

tbaxter-18f (approved):

tock/hours/views.py
Sgtpluck: thank you both!!!

Issue #1259, opened 2021-03-24
Create timecard api endpoint 


Jkrzy:
As a tock user, I want to be able to retrieve a list of timecards, along with their attributes, via the API so that I can generate adhoc reports in sheets.

### We'll know we're done when
- [x] We've added an API endpoint with django restframework that serializes and serves the `Timecard` model

nick-jones-gov:
Took a quick look into this, is this something that the existing `/api/timecards.json` endpoint (defined [here](https://github.com/18F/tock/blob/main/tock/api/urls.py#L22)) could be used for? Or maybe there's additional/different data you're looking for here?

Jkrzy:
The existing `timecards.json` endpoint returns a `TimeCardObject` queryset and instances of `TimecardObject`.

The naming here is super confusing but this framing has helped me navigate it:

An instance of `TimecardObject` is one project-entry on an associated `TimeCard`.

The existing `timecards.json` exports "timecard data" by returning all entries  -- `TimecardObjects -- on all timecards for a given query.

What we're looking for here is an endpoint which returns only the data from `hours.models.Timecard`.

nick-jones-gov:
Ah ok, that's helpful - thanks @Jkrzy!

nick-jones-gov:
are there specific fields you're looking to filter on here? I'm assuming that you don't want all `Timecard`s that exist in the DB. I'm imagining that filtering by `submitted` status, user, some concept of time (i.e. "all timecards whose reporting period ended after X date"), maybe business unit? Or if this is a more open question, I can start with what makes sense to me and we can refine in code review.

PR #1265, opened 2021-04-20
Add API endpoint for fetching timecard data

nick-jones-gov:
## Description

This addresses #1259 by adding a new endpoint to fetch Timecard data. It's ready for review, but I'm expecting we'll want to discuss some of the details of what exact data and query parameters to include.

## Additional information
This is based on some initial conversation in #1259. I didn't attempt to refactor any of the "timecard" vs "timecard object" verbiage in various API endpoints, serializers, or API views, though I'm definitely open to that idea. I did add some documentation around that in hopes of making that distinction clearer to future code readers and writers. As-is, I used "full timecard" to mean `Timecard` to disambiguate from the existing use of "timecard" referring to `TimecardObject`s (e.g. the existing `/timecards.json` endpoint returns `TimecardObject`s).

### DIFF
--- a/tock/api/urls.py
+++ b/tock/api/urls.py
@@ -3,7 +3,7 @@
 from django.urls import path, re_path
 
 from api.views import ProjectList, ProjectInstanceView, UserList, ReportingPeriodList, ReportingPeriodAudit, \
-    Submissions, TimecardList, hours_by_quarter, hours_by_quarter_by_user, UserDataView
+    Submissions, TimecardList, hours_by_quarter, hours_by_quarter_by_user, UserDataView, FullTimecardList
 
 urlpatterns = [
     path('projects.json', ProjectList.as_view(), name='ProjectList'),
@@ -19,7 +19,11 @@ urlpatterns = [
         Submissions.as_view(),
         name='Submissions'
     ),
+    # Note that this endpoint returns TimecardObjects
     path('timecards.json', TimecardList.as_view(), name='TimecardList'),
+    # ...and this one returns Timecards (referred to as "full timecards" to disambiguate
+    # from existing endpoints / naming conventions)
+    path('full_timecards.json', FullTimecardList.as_view(), name='FullTimecardList'),
     path('hours/by_quarter.json', hours_by_quarter, name='HoursByQuarter'),
     path('hours/by_quarter_by_user.json', hours_by_quarter_by_user, name='HoursByQuarterByUser'),
     path('users.json', UserList.as_view(), name='UserList'),
--- a/tock/api/views.py
+++ b/tock/api/views.py
@@ -3,9 +3,10 @@ import datetime
 
 from django.contrib.auth import get_user_model
 from django.db import connection
-from django.db.models import Count, F
+from django.db.models import Count, F, Sum
 
 from rest_framework import serializers, generics
+from rest_framework.exceptions import ParseError
 
 from hours.models import TimecardObject, Timecard, ReportingPeriod
 from projects.models import Project
@@ -129,6 +130,41 @@ class TimecardSerializer(serializers.Serializer):
         allow_null=True
     )
 
+class FullTimecardSerializer(serializers.ModelSerializer):
+    # Fields that require accessing other models
+    user_name = serializers.CharField(source='user.username')
+    reporting_period_start_date = serializers.DateField(source='reporting_period.start_date')
+    reporting_period_end_date = serializers.DateField(source='reporting_period.end_date')
+    hours_spent = serializers.SerializerMethodField()
+
+    class Meta:
+        model = Timecard
+        fields = [
+            # straight pass-through fields
+            'id',
+            'submitted',
+            'submitted_date',
+            'billable_expectation',
+            'target_hours',
+            'billable_hours',
+            'non_billable_hours',
+            'excluded_hours',
+            'utilization',
+            # fields from other models
+            'user_name',
+            'reporting_period_start_date',
+            'reporting_period_end_date',
+            # fields that require something more complicated (e.g. an aggregation)
+            'hours_spent'
+        ]
+
+    def get_hours_spent(self, obj):
+        # Each Timecard has many TimecardObjects, sum up the hours spent in each
+        # of those TimecardObject objects
+        total_hours_spent_agg = obj.timecardobjects.aggregate(total_hours=Sum('hours_spent'))
+
+        return total_hours_spent_agg['total_hours']
+
 # API Views
 
 class UserDataView(generics.ListAPIView):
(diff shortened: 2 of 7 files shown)


tock/api/views.py
```diff
@@ -129,6 +130,41 @@ class TimecardSerializer(serializers.Serializer):
...
+            'utilization',
+            # fields from other models
+            'user_name',
+            'reporting_period_start_date',
+            'reporting_period_end_date',
+            # fields that require something more complicated (e.g. an aggregation)
+            'hours_spent'
+        ]
+
+    def get_hours_spent(self, obj):
```
nick-jones-gov, line 161: I added this before realizing that `Timecard` has a `calculate_hours` method that updates various `_hours` fields when a `Timecard` object is updated. Are those `_hours` fields sufficient, or do we want to include this `total_hours` field too?

tock/api/views.py
```diff
@@ -206,6 +242,29 @@ def get_queryset(self):
...
+        # Lookup the associated user and reporting_period in the original
+        # query since we'll be accessing them later. See https://docs.djangoproject.com/en/3.2/ref/models/querysets/#django.db.models.query.QuerySet.select_related
+        queryset = Timecard.objects.select_related(
+            'user',
+            'reporting_period',
+        )
+        if self.request.query_params.get('only_submitted') is not None:
+            queryset = queryset.filter(submitted=True)
+
+        earliest_reporting_period_start = self.request.query_params.get('earliest_reporting_period_start')
```
nick-jones-gov, line 258: Wasn't sure if this was the type of parameter that people would find useful - open to changing this and/or adding other query params!

tock/api/urls.py
```diff
@@ -19,7 +19,11 @@
         Submissions.as_view(),
         name='Submissions'
     ),
+    # Note that this endpoint returns TimecardObjects
     path('timecards.json', TimecardList.as_view(), name='TimecardList'),
+    # ...and this one returns Timecards (referred to as "full timecards" to disambiguate
+    # from existing endpoints / naming conventions)
+    path('full_timecards.json', FullTimecardList.as_view(), name='FullTimecardList'),
```
nick-jones-gov, line 26: I'm very open to naming conversations here, wanted to put this CR up first and start the conversation before going on any refactoring voyage to rename "timecard"-related things that are in fact `TimecardObject`s

tock/hours/models.py
```diff
@@ -455,6 +461,12 @@ def save(self, *args, **kwargs):
         super(TimecardNote, self).save(*args, **kwargs)
 
 class TimecardObject(models.Model):
+    """
+    Not to be confused with a Timecard, a TimecardObject can be thought of as one "entry" in a Timecard,
+    i.e. for a given billing period, one person generally has one Timecard, but many TimecardObjects ("entries"). Each
+    represents a project that person worked on, and the amount of time (and some other data defined
+    below).
+    """
```
Jkrzy, line 469: ❤️

tock/hours/models.py
```diff
@@ -229,6 +229,12 @@ def get_most_recent_periods(cls, number_of_periods=1):
 
 
 class Timecard(models.Model):
+    """
+    A Timecard is a top-level representation of what a user worked on during a reporting period.
+
+    For details on what projects they worked on and for how long, you'll want to access the
+    associated `TimecardObject` objects, which store that level of detail.
+    """
```
Jkrzy, line 237: ❤️

tock/api/views.py
Jkrzy: Great question! The fields on `Timecard` are sufficient, any extra calculations for these instances on the way out are unexpected and _shouldn't_ be needed

tock/api/views.py
Jkrzy: Ooo, this is a great spot to re-use some of @neilmb's work from a few months ago -- specifically https://github.com/18F/tock/blob/623554bf1eb490a5e1ef48552063184661c0c339/tock/api/views.py#L228.

Using this will let us filter on a handful of timecard fields, enough for this initial version.

tock/api/urls.py
Jkrzy: I'm drawn to embarking on a renaming voyage as well but this will suffice for our immediate. The additional comments definitely help here.

tock/api/views.py
nick-jones-gov: Sounds good!

tock/api/views.py
nick-jones-gov: Ah good call, refactored to re-use that!

tock/api/views.py
```diff
@@ -260,14 +307,14 @@ def filter_timecards(queryset, params={}):
...
-        after_date = params.get('after')
+        after_date = date_from_iso_format(params.get('after'))
         queryset = queryset.filter(
             reporting_period__end_date__gte=after_date
         )
 
     if 'before' in params:
         # get everything before a specified date
-        before_date = params.get('before')
+        before_date = date_from_iso_format(params.get('before'))
```
nick-jones-gov, line 317: @Jkrzy and @neilmb not sure if this came up before, but I'm also wondering how we think about reporting periods in the context of the "before" and "after" arguments here - say we had a reporting period of April 1 - April 15. If someone queries for "before April 7", are they expecting to get timecards whose end date is _after_ April 7? Or should the entirety of the reporting period fall before April 7?

As-is now, if someone asks for timecards "before April 7", we would include ones from the time period April 1 - April 15 (because its start date is April 1, which falls before April 7). I could see it either way, but think it might be confusing the way it is now since you'd be including data from work that could've been done after April 7 (e.g. if someone tocked 10 hours on April 9, that would be counted).

neilmb: It looks to me like `after` gets every period whose end_date is after the specified date https://github.com/18F/tock/blob/main/tock/api/views.py#L265 and `before` gets every period whose start_date is before the specified date https://github.com/18F/tock/blob/main/tock/api/views.py#L272.

I think that I chose that way on purpose, giving the larger returned set of data so that you don't have to move a date forward (or backward) in time until you get past the end (start) date to get a time period to be returned.  (Hmm, was April 15, 2020 a Saturday?)

But yes, I remember thinking that there were two different ways to interpret this.

nick-jones-gov: Yes, agree with how the `before` and `after` params are working now! Just wanted to clarify/double check that that's how users of the API were using / understanding it. Thanks @neilmb!

nick-jones-gov:
@Jkrzy this is ready for another round of review when you get a chance!

Jkrzy (approved):
Thanks @nick-jones-gov, looks good.

PR #1281, opened 2021-06-04
Add “https (443)” to logging connections as per SSP review request.

tadhg-ohiggins:
Without showing ports
How will we know what's secure?
Add the port numbers.

During the SSP review we were asked to add `https (443)` to the logging connection lines in the network diagram. This commit adds that information in as minimal way as I could find.


kfoley-18F (approved):
LGTM! thanks for updating this 👍

PR #1297, opened 2021-06-22
Replace node-sass with sass

### LINKED ISSUE #1293: swap out node-sass for dart-sass
node-sass is deprecated - https://www.npmjs.com/package/node-sass has details. 
Recommended replacement is dart-sass: https://sass-lang.com/dart-sass

node-sass is also the source of a number of vulnerabilites - https://snyk.io/test/github/18f/tock?targetFile=package.json

Time to update!

matthinz:
## Description

`node-sass` has been deprecated in favor of `sass` (formerly `dart-sass`). This PR replaces the `node-sass` dependency with a `sass` and adjusts the `build-css` script to use it.

## Additional information

The generated css is quite similar, but the diff between the `node-sass` and `sass` output is somewhat noisy. The differences in output are limited to things that either do not affect CSS rendering (ordering of selectors in multi-selector rules) or should not in day-to-day use (increased precision in floating point math resulting in e.g. `1.3128205128ex` instead of `1.31282ex`).

There are also warnings due to use of the deprecated `/` operator, but it looks like those are upstream in [uswds](https://github.com/uswds/uswds/issues/4204).

Closes #1293

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,6 +6,7 @@
       "version": "7.14.5",
       "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.14.5.tgz",
       "integrity": "sha512-9pzDqyc6OLDaqe+zbACgFkb6fKMNG6CObKpnYXChRsvYGyEdc7CA2BaqeOM+vOtCS5ndmJicPJhKAwYRI6UfFw==",
+      "dev": true,
       "requires": {
         "@babel/highlight": "^7.14.5"
       }
@@ -204,7 +205,8 @@
     "@babel/helper-validator-identifier": {
       "version": "7.14.8",
       "resolved": "https://registry.npmjs.org/@babel/helper-validator-identifier/-/helper-validator-identifier-7.14.8.tgz",
-      "integrity": "sha512-ZGy6/XQjllhYQrNw/3zfWRwZCTVSiBLZ9DHVZxn9n2gip/7ab8mv2TWlKPIBk26RwedCBoWdjLmn+t9na2Gcow=="
+      "integrity": "sha512-ZGy6/XQjllhYQrNw/3zfWRwZCTVSiBLZ9DHVZxn9n2gip/7ab8mv2TWlKPIBk26RwedCBoWdjLmn+t9na2Gcow==",
+      "dev": true
     },
     "@babel/helper-validator-option": {
       "version": "7.14.5",
@@ -227,6 +229,7 @@
       "version": "7.14.5",
       "resolved": "https://registry.npmjs.org/@babel/highlight/-/highlight-7.14.5.tgz",
       "integrity": "sha512-qf9u2WFWVV0MppaL877j2dBtQIDgmidgjGk5VIMw3OadXvYaXn66U1BFlH2t4+t3i+8PhedppRv+i40ABzd+gg==",
+      "dev": true,
       "requires": {
         "@babel/helper-validator-identifier": "^7.14.5",
         "chalk": "^2.0.0",
@@ -237,6 +240,7 @@
           "version": "3.2.1",
           "resolved": "https://registry.npmjs.org/ansi-styles/-/ansi-styles-3.2.1.tgz",
           "integrity": "sha512-VT0ZI6kZRdTh8YyJw3SMbYm/u+NqfsAxEpWO0Pf9sq8/e94WxxOpPKx9FR1FlyCtOVDNOQ+8ntlqFxiRc+r5qA==",
+          "dev": true,
           "requires": {
             "color-convert": "^1.9.0"
           }
@@ -245,6 +249,7 @@
           "version": "2.4.2",
           "resolved": "https://registry.npmjs.org/chalk/-/chalk-2.4.2.tgz",
           "integrity": "sha512-Mti+f9lpJNcwF4tWV8/OrTTtF1gZi+f8FqlyAdouralcFWFQWF2+NgCHShjkCb+IFBLq9buZwE1xckQU4peSuQ==",
+          "dev": true,
           "requires": {
             "ansi-styles": "^3.2.1",
             "escape-string-regexp": "^1.0.5",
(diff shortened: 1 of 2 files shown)


kfoley-18F:
@matthinz  - looks like sass uses earlier versions on some dependencies (lookin' at you, lodash) where there are some known vulnerabilities. :/ 
is the-package-formerly-known-as-dart-sass _actually_ the best choice (I was going on light googling and recommendations for this ticket) or is there a different alternative at this point that might have more recent dependency versions?

matthinz:
@kfoley-18F I'm not seeing `lodash` in the dependency tree for `sass`. It looks like the only dependency it is bringing in is [chokidar](https://www.npmjs.com/package/chokidar). Here is what `npm ls` gives me:

```
├─┬ sass@1.35.1
│ └─┬ chokidar@3.5.2
│   ├─┬ anymatch@3.1.2
│   │ ├── normalize-path@3.0.0 deduped
│   │ └── picomatch@2.2.2
│   ├─┬ braces@3.0.2
│   │ └─┬ fill-range@7.0.1
│   │   └─┬ to-regex-range@5.0.1
│   │     └── is-number@7.0.0
│   ├── fsevents@2.3.2
│   ├─┬ glob-parent@5.1.2
│   │ └── is-glob@4.0.1 deduped
│   ├─┬ is-binary-path@2.1.0
│   │ └── binary-extensions@2.2.0
│   ├─┬ is-glob@4.0.1
│   │ └── is-extglob@2.1.1
│   ├── normalize-path@3.0.0
│   └─┬ readdirp@3.6.0
│     └── picomatch@2.2.2 deduped
```

The `sass` package is built from the [sass/dart-sass project](https://github.com/sass/dart-sass), which is replacing the Ruby and libsass implementations as the canonical version. More info: https://github.com/sass/dart-sass#why-dart

package-lock.json
```diff
@@ -6,6 +6,7 @@
       "version": "7.14.5",
       "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.14.5.tgz",
       "integrity": "sha512-9pzDqyc6OLDaqe+zbACgFkb6fKMNG6CObKpnYXChRsvYGyEdc7CA2BaqeOM+vOtCS5ndmJicPJhKAwYRI6UfFw==",
+      "dev": true,
```
kfoley-18F, line 9: What's the stability and maintenance on this dev=true package?

package-lock.json
```diff
@@ -204,7 +205,8 @@
     "@babel/helper-validator-identifier": {
       "version": "7.14.8",
       "resolved": "https://registry.npmjs.org/@babel/helper-validator-identifier/-/helper-validator-identifier-7.14.8.tgz",
-      "integrity": "sha512-ZGy6/XQjllhYQrNw/3zfWRwZCTVSiBLZ9DHVZxn9n2gip/7ab8mv2TWlKPIBk26RwedCBoWdjLmn+t9na2Gcow=="
+      "integrity": "sha512-ZGy6/XQjllhYQrNw/3zfWRwZCTVSiBLZ9DHVZxn9n2gip/7ab8mv2TWlKPIBk26RwedCBoWdjLmn+t9na2Gcow==",
+      "dev": true
```
kfoley-18F, line 209: same concern as other dev package

package-lock.json
```diff
@@ -1069,17 +1006,8 @@
...
-      "version": "0.7.0",
-      "resolved": "https://registry.npmjs.org/aws-sign2/-/aws-sign2-0.7.0.tgz",
-      "integrity": "sha1-tG6JCTSpWR8tL2+G1+ap8bP+dqg="
-    },
-    "aws4": {
-      "version": "1.11.0",
-      "resolved": "https://registry.npmjs.org/aws4/-/aws4-1.11.0.tgz",
-      "integrity": "sha512-xh1Rl34h6Fi1DC2WWKfxUTVqRsNnr6LsKz2+hfwDxQJWmrx8+c7ylaqBMcHfl1U1r2dsifOvKX3LQuLNZ+XSvA=="
+      "integrity": "sha1-x57Zf380y48robyXkLzDZkdLS3k=",
+      "dev": true
```
kfoley-18F, line 1010: why is this dev= true but otherwise unchanged from before?

package-lock.json
```diff
@@ -1231,29 +1156,15 @@
...
-    "camelcase-keys": {
-      "version": "6.2.2",
-      "resolved": "https://registry.npmjs.org/camelcase-keys/-/camelcase-keys-6.2.2.tgz",
-      "integrity": "sha512-YrwaA0vEKazPBkn0ipTiMpSajYDSe+KjQfrjhcBMxJt/znbvlHd8Pw/Vamaz5EB4Wfhs3SUR3Z9mwRu/P3s3Yg==",
-      "requires": {
-        "camelcase": "^5.3.1",
-        "map-obj": "^4.0.0",
-        "quick-lru": "^4.0.1"
-      }
+      "integrity": "sha512-L28STB170nwWS63UjtlEOE3dldQApaJXZkOI1uMFfzf3rRuPegHaHesyee+YxQ+W6SvRDQV6UrdOdRiR153wJg==",
```
kfoley-18F, line 1159: +++ unchanged from previous version but now dev=true?

kfoley-18F: actually looks like a lot of these are now dev=true but the package (version, sha) is unchanged from before.... 
🤔

matthinz: npm has had a history of different versions of the CLI producing slightly different `package-lock.json` files, so my initial instinct here was "🤷 , npm being npm?". However, then I actually went and did some research and learned something:

In this context, `dev: true` means "this package is either one of our `devDependencies` or is only depended on by our `devDependencies` ([source](https://docs.npmjs.com/cli/v6/configuring-npm/package-lock-json#dev)). This is a hint to npm itself--if you were to run `npm ci --production` it will not install anything in `package-lock.json` flagged as `dev`. It is _not_ saying "hey npm, it's fine for you to install a dev version of this package." That's why the integrity hash is unchanged--the same code is being installed as before. For our purposes, this doesn't mean much, since we install dev + non-dev dependencies in CI for our build.

`node-sass` had several dependencies in common with packages listed in our `devDependencies`. We removed `node-sass` from our `dependencies`, and npm recognized that those previously shared dependencies are now only required by our `devDependencies` and marked them as `dev: true` in `package-lock.json`.

kfoley-18F (approved):
Let's get this on staging and take a quick test pass there!

PR #1308, opened 2021-07-22
Npm audit / Node version update

matthinz:
This PR:

- Updates the version of Node.js used to run `sass` in development to 14.17.3 (latest LTS)
- Applies `npm audit fix` to update dependencies with vulnerabilities in a semver-compatible way
- Applies `npm audit fix --force` to update dependencies with vulnerabilities in a potentially non-semver compatible way
- Adds `node-notifier` as a dev dependency (this was previously directly depended on by Jest, but newer Jest versions only require it if you  are using `notify: true` in your jest.config.js [which we are]).

This is what `npm audit` looks like now:

```
$ npm audit
                                                                                
                       === npm audit security report ===                        
                                                                                
found 0 vulnerabilities
 in 658 scanned packages
```

This should close #1298 

I'd like to rebase #1297 on `main` after this lands and proceed with eliminating `node-sass` entirely.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -23,7 +23,7 @@ db:
     - POSTGRES_DB=tock
     - POSTGRES_USER=tock_user
 sass:
-  image: node:8.11.1
+  image: node:14.17.3
   volumes:
     - .:/tock
     - /tock/node_modules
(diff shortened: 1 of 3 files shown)


kfoley-18F (approved):

kfoley-18F:
Thanks @matthinz !! 
That's a LOT of dependency updates!

Issue #1310, opened 2021-07-26
staticfiles is deprecated, need to update it


kewlguy781:
staticfiles codes bonked the server after updating Django. This may be considered as blocker.

kewlguy781:
<h1 style="padding: 0px; margin: 0px; font-weight: normal; color: rgb(0, 0, 0); font-family: sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">TemplateSyntaxError at /</h1><pre class="exception_value" style="padding: 0px; margin: 10px 0px; font-size: 1.5em; white-space: pre-wrap; font-family: sans-serif; color: rgb(87, 87, 87); font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">'staticfiles' is not a registered tag library. Must be one of:
admin_list
admin_modify
admin_urls
analytics
cache
has_submitted_timesheet
i18n
index
l10n
log
project_tags
static
tz</pre>

Request Method: | GET
-- | --
http://localhost:8000/
3.1.13
TemplateSyntaxError
'staticfiles' is not a registered tag library. Must be one of: admin_list admin_modify admin_urls analytics cache has_submitted_timesheet i18n index l10n log project_tags static tz
/usr/local/lib/python3.9/site-packages/django/template/defaulttags.py, line 1036, in find_library
/usr/local/bin/python
3.9.5
['/tock',  '/usr/local/lib/python39.zip',  '/usr/local/lib/python3.9',  '/usr/local/lib/python3.9/lib-dynload',  '/usr/local/lib/python3.9/site-packages']
Mon, 26 Jul 2021 17:56:41 -0400

kewlguy781:
the error is showing from `/tock/tock/templates/base.html`, error at line 10 But will need to do a folder wide search.

PR #1311, opened 2021-07-26
Jon

kewlguy781:
## Description

This pull request will update the Django to 3.1.13, as well as fixing some codes that has been depreciated (`staticfiles` to `static` as per issue #1310 which is then resolved.)


kewlguy781:
Note: This will be a bit on hold till we get the previous release into the staging. We would like to make sure the last release is being validated and keep the release independent with the previous release and this request.

kewlguy781:
Because there were a release that require me to redo the branch. I have created branch `jon2` and moved all the changes I made over here to there.

PR #1313, opened 2021-07-28
Dockerize jest / puppeteer integration tests

### LINKED ISSUE #1309: Dockerize Jest / Puppeteer tests
The docs include [a note](https://github.com/18F/tock/blob/main/docs/local-development.md#testing-locally) about how the Jest / Puppeteer tests need to be Dockerized. As it stands, it's a little awkward to run them locally, since the local `node_modules` folder is volumed into the running `sass` container, which means you need to do a separate `npm install` to make them usable for running on a Mac.

matthinz:
## Description

This PR makes it possible to run Puppeteer-based integration tests locally inside a Docker container.

## Additional information

I added an additional `integration-tests` service to `docker-compose.yml` that creates a Docker container based on the image we use in CI. By default, `docker-compose up` won't run the tests, but you can run them once the environment  is up and available:

```sh
docker-compose run integration-tests npm test
```

The best route for actually authoring _new_ tests is likely still running outside of Docker, since that will allow you to debug tests more easily. I experimented with getting my Chrome dev tools to connect to the browser running inside the Docker container, but did not have much luck.

Closes #1309


kfoley-18F (approved):
🚢  !!

Thanks @matthinz !

PR #1319, opened 2021-08-05
CVE for <=3.2.1 django-debug-toolbar

kfoley-18F:
IonChannel shows a critical-level CVE for django-debug-toolbar. 
The pipfile.lock shows Tock is already using 3.2.1, so this should not change deployed resources at all. 
Recognizing that this is not deployed to production, it's still triggering IonChannel reports, and critical issues have a short fix window.

Updating the pipfile to require >3.2.1 should clear it.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -25,7 +25,7 @@ pyjwt = "~=1.7.1"
 [dev-packages]
 bandit = "*"
 coverage = "*"
-django-debug-toolbar = "*"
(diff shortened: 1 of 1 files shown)


kfoley-18F:
Ran into gh key issues updating. Tadhg submitted this fix in [#1320](https://github.com/18F/tock/pull/1320) instead.

PR #1323, opened 2021-08-13
Update Django to 3.2

kewlguy781:
## Description

This PR will update the Django web application to 3.2 (which is the next LTS release)

## Additional information

Changes done to work with Django 3.2:

- We no longer can mix the numbers' data type. For example:
`billable = Coalesce(Sum('timecardobjects__hours_spent', filter=billable_filter), 0)` -> 
`billable = Coalesce(Sum('timecardobjects__hours_spent', filter=billable_filter), Decimal('0'))` The former code set the variable as integer 0, while the rest is in Decimal. We changed it to Decimal to play nice with other variables that are already in Decimal. See [Value() Changes under Model](https://docs.djangoproject.com/en/3.2/releases/3.2/#models)
- Change of Autofield. They have changed the autofield to 64-bit integer. [See the documentation](https://docs.djangoproject.com/en/3.2/releases/3.2/#customizing-type-of-auto-created-primary-keys). For now we have set it as `AutoField` instead of new default `BigAutoField` however, we will need to revisit this and migrate into database.
- We also updated [`cg-django-uaa`](https://github.com/cloud-gov/cg-django-uaa) to support Django 3.2 (Special shout out to @bengerman13 for working with me)


kfoley-18F:
Woohooooo!! Thank you @kewlguy781 !!!

Requesting some review of the Decimal and AutoField changes from @Jkrzy pretty please. 

I'm super-excited to get this merged... ❤️

requirements.txt
```diff
@@ -10,11 +10,11 @@ beautifulsoup4==4.9.3
 bleach==3.3.1
 certifi==2021.5.30
 cfenv==0.5.3
-cg-django-uaa==2.1.1
+cg-django-uaa==2.1.3
 charset-normalizer==2.0.4; python_version >= '3'
 dj-database-url==0.5.0
 django-webtest==1.9.7
-django==3.1.13
+django==3.2.6
```
Jkrzy, line 17: Thanks @kewlguy781! It looks like we may have updated `requirements.txt` directly. We should use pipenv to generate this from our lock file so we've got consistent definitions across both files.

```bash
pipenv lock --requirements > requirements.txt
```

Jkrzy (changes requested):
This looks great! There's one small opportunity I noted to align `requirements.txt` with the lock file

Jkrzy (approved):

PR #1350, opened 2021-09-22
Fix bug with not ignoring empty forms

neilmb:
## Description

The weekly billing code introduced a new (usually hidden) field on the timecard form and the Javascript was setting a value for that field which was confusing Django's built-in mechanism for detecting if form fields have changed. The observed bug was that on submitting the second time card, the `extra` field that Django added was giving a "This field is required" error.

This fixes the bug by changing `timecard.js` to leave that field unselected and it adds a test for submitting two time cards to make sure that the second one works.

### DIFF
--- a/tock/hours/forms.py
+++ b/tock/hours/forms.py
@@ -130,24 +130,17 @@ def projects_as_choices(queryset=None):
     return accounting_codes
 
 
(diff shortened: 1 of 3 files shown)


kewlguy781 (approved):
look good. (and thanks for cleaning out my print codes)

PR #1351, opened 2021-09-22
Fix docker syntax for JS test execution

caleywoods:
## Description

`docker-compose run` is required instead of just `docker run` when attempting to run the JS integration test suite. I encountered a docker failure message when trying to run the previous command. @matthinz pointed me in the right direction.


matthinz (approved):
Thank you @caleywoods!

PR #1356, opened 2021-09-27
Remove hourly summation elements if weekly billable project exists on timecard

caleywoods:
## Description

Remove hourly summation elements if weekly billable project exists on timecard

## Additional information

The hourly summation elements in the footer should hide upon selection of a weekly billable project as well as being hidden when coming into a saved timecard that has a weekly billable project on it. Removing any weekly billable projects from the timecard either by deleting it or updating it should restore the hourly summation elements.

JS tests were added and were passing on my local machine but seem to fail when ran inside the docker instance, that probably needs its own issue.

### DIFF
--- a/tock/tock/static/js/components/timecard.js
+++ b/tock/tock/static/js/components/timecard.js
@@ -208,6 +208,7 @@ function toggleNotesField(selectBoxId) {
   const options = document.querySelector('#' + selectBoxId + '-select')
     .selectedOptions[0].dataset;
   if (options.is_weekly_bill === 'true') {
+    setHourSummationVisibility('hide');
     project_allocation.classList.remove('entry-hidden');
     hours_spent.classList.add('entry-hidden');
   } else {
--- a/tock/tock/static/js/components/timecard.js
+++ b/tock/tock/static/js/components/timecard.js
@@ -370,7 +370,8 @@ function handleConfirm(val) {
  * */
 function setHourSummationVisibility(visibility) {
   const showBillingHourSummationElements = visibility === 'show';
-  const [totalReportedElement, totalBillableElement] = Array.from(document.querySelectorAll('.entries-total .grid-col-2'));
+  const totalReportedElement = document.querySelector('#total-reported-div');
+  const totalBillableElement = document.querySelector('#total-billable-div');
 
   if (showBillingHourSummationElements) {
     totalReportedElement.classList.remove('entry-hidden');
--- a/tock/tock/static/js/components/timecard.js
+++ b/tock/tock/static/js/components/timecard.js
@@ -362,6 +362,24 @@ function handleConfirm(val) {
   }
 }
 
+/** @function
+ * Show or hide the billing hour summation elements in the footer
+ * @name setHourSummationVisibility
+ * @param {string} visibility
+ * */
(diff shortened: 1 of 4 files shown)


kewlguy781:
CircleCi show as fail. https://app.circleci.com/pipelines/github/18F/tock/1458/workflows/5e7d1820-5fa8-4ace-8d78-16c3dbe22d5b/jobs/7259

tock/tock/static/js/components/timecard.js
```diff
@@ -363,6 +368,56 @@ function handleConfirm(val) {
...
+ * @name handleBillingElementState
+ * @param {string} visibility
+ * */
+function handleBillingElementState(visibility) {
+  const showBillingHourSummationElements = visibility === 'show';
+  const totalReportedElement = document.querySelector('#total-reported-div');
+  const totalBillableElement = document.querySelector('#total-billable-div');
+
+  if (showBillingHourSummationElements) {
+    const weeklyBillingAlertElement = document.querySelector('#weekly-billing-alert');
```
neilmb, line 382: **suggestion (non-blocking):** these two calls are repeated inside the conditional, but the ones for `totalReportedElement` and `totalBillableElement` are not. I think it's clearer to be consistent and pull all four of the `querySelector` calls outside the if statements and only make the changes in the if and else blocks.

tock/tock/static/js/components/timecard.js
```diff
@@ -363,6 +368,56 @@ function handleConfirm(val) {
...
+  const selects = document.querySelectorAll('.entry-project select');
+  let weeklyBilledProjectSeen = false;
+  for (let i = 0; i < selects.length; i++) {
+    const currentSelect = selects[i];
+    const project = currentSelect.selectedOptions[0];
+    const projectIsWeeklyBilled = (project.dataset || {}).is_weekly_bill === 'true';
+    // Flip our 'state' so we know we've encountered at least one weekly billed project
+    // Get out of the loop as soon as we've found one weekly billed project
+    if (projectIsWeeklyBilled && !weeklyBilledProjectSeen) {
+      weeklyBilledProjectSeen = true;
```
neilmb, line 414: **suggestion (non-blocking):** I think that we could `return true;` directly from here and then change L418 to `return false;`. Then the `weeklyBilledProjectSeen` variable isn't needed at all.

tock/tock/tests/js/integration/timecard.test.js
```diff
@@ -81,7 +81,86 @@ describe("Timecard", () => {
...
+  describe("handles weekly billing", () => {
+    beforeEach(() => page.goto(`${baseUrl}/reporting_period/2015-03-30/`));
+
+    test("hides hours input and shows project allocation element", async () => {
+      // https://github.com/18F/tock/issues/848
+      await page.evaluate(() => {
+        document.querySelector("#id_timecardobjects-0-hours_spent").value = ''
+      });
+
+      // @TODO: Seed database with a weekly billing project and know the ID for this test
```
neilmb, line 95: **question:** It looks like this is already done in the test data, do we still need this TODO comment?

neilmb (approved):
Great to have this, just a few comments and minor suggestions.

PR #1362, opened 2021-09-29
More decimal places for small allocations

neilmb:
## Description

The new smallest weekly allocation of 0.125 requires us to allow 3 decimal places for `project_allocation` and the employee's `expected_project_allocation`. Includes a new forms test to make sure that we are allowed to set this value.

@colinmurphy01


kfoley-18F (approved):
🚢  LGTM!

:thankyou: for the prompt fix @neilmb !

Issue #1365, opened 2021-10-01
Weekly billing TimecardObject server error and hourly validation issue


caleywoods:
This issue is capturing two problems that are related within the Django admin interface when editing a users previously saved Timecard with a weekly billing project or attempting to add one.

The first issue is that when you first enter a saved timecard and this timecard either already has a weekly billing project that you need to edit or you're adding a new weekly billing project, the "hours spent" column is completely empty and if you don't type a zero into that field before hitting save you will encounter a server error. The reproduction steps for local tock are as follows:

1. Visit the Django admin area and click on "Timecards" under the "Hours" section.
2. Click "Add Timecard" in the upper right and then pick a user and reporting period at the top of the form. You may need to create a new reporting period if you don't have one open.
3. In the Timecard Objects area of the form, select a weekly billed project from the list. If you've loaded your test database recently you can use project "125 - Weekly Billing Test", if you don't have a weekly billed project, please create one and come back to this process. 
4. Leave the "Hours Spent" field completely blank (this is the root cause of the server error)
5. Select any allocation percentage from the "Project allocation" dropdown and then attempt to save the form.
6. Observe that you should have encountered a server error telling you that it encountered a `TypeError`.

The second issue is that we're still attempting to validate that a user is working 40 hours on their timecard when a weekly billed project exists. To reproduce this issue:

1. Visit the Django admin area and click on "Timecards" under the "Hours" section.
2. Click "Add Timecard" in the upper right and then pick a user and reporting period at the top of the form. You may need to create a new reporting period if you don't have one open.
3.  In the Timecard Objects area of the form, select a weekly billed project from the list. If you've loaded your test database recently you can use project "125 - Weekly Billing Test", if you don't have a weekly billed project, please create one and come back to this process. 
4. Enter a zero (0) in the "Hours Spent" field
5. Select any allocation percentage form the "Project allocation" dropdown and then attempt to save the form.
6. Observe that you should be instructed to correct an error and that warns you that you've entered fewer than 40 hours just below the Timecard Objects section of the form.

We should create tests to cover these scenarios and implement the fixes.

caleywoods:
Resolved by #1366

PR #1366, opened 2021-10-01
Admin weekly billing fixes

caleywoods:
## Description

This PR addresses #1365 

Hourly validation is skipped for TimecardObjects if a weekly billing project is detected on the Timecard and a server `TypeError` is resolved that originated when a user would leave the "Hours Spent" field blank when entering a weekly billed project. An integration test was added to catch this issue.


kfoley-18F (approved):
🚢  
Thanks @caleywoods !!

Issue #1363, opened 2021-09-29
Allocation percentages should be displayed in admin console


neilmb:
The project allocation field is present in the admin consoles TimecardObject inline formset, but the actual value of the allocation for that timecard object is not displayed. It's not obvious why this model field isn't being bound correctly to the "selected" choices in the form widget.

We may have to override the `initial` property of the field in some init method somewhere to get the instance's value to be displayed correctly in the widget.

PR #1371, opened 2021-10-07
Project allocation percentages to show in timecard admin

neilmb:
## Description

#1363 reports that the select widget in the timecard admin page doesn't display the allocation percentage correctly, even when it is set in the database. The issue ended up being a problem with a DecimalField that specifies a restricted set of choices and so gets rendered on the page with Django's `Select` widget.

The `Select` widget determines if a value is selected by string comparing the list of values against the current value of the widget. This doesn't work for numeric values because the string comparison gives different things depending on formatting. The fix here is to use a custom widget where the selection logic is overridden to use numeric comparison instead of string comparison.

## Additional information
Screen Shot 2021-10-07 at 12 55 58

### DIFF
--- a/tock/hours/admin.py
+++ b/tock/hours/admin.py
@@ -1,9 +1,11 @@
 from decimal import Decimal
+from math import isclose
 
 from django.conf import settings
 from django.contrib import admin
 from django.core.exceptions import ValidationError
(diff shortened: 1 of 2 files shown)


colinmurphy01:
<3

Issue #1373, opened 2021-10-08
Weekly Billing - Allocation % doesn't display in dropdown when user "saves" (instead of submit)


colinmurphy01:
Reproduce:
- Open a new timecard
- Select weekly billing project and select allocation %
- Select "Save your time card for later"
- Timecard saves and page reloads
- Allocation % saved doesn't render in dropdown, similar to hours

Expected:
- Saved allocation % is rendered in dropdown

I don't know how many people save their timecards continuously, so I can't assess this impact this bug has on people.

caleywoods:
Thanks @colinmurphy01. I think @neilmb has already been looking at this but I'm not sure what its most recent status is. We were thinking the fix could be similar to the change made in the Admin area to get the allocation percentage to show up correctly.

PR #1375, opened 2021-10-12
Use our custom decimal choice widget on timecard entry page

neilmb:
## Description

#1373 describes a problem with allocation percentages not showing up when people save their time cards before submitting. The underlying issue is the same as in #1363. The new Decimal-type choice widget that we introduced for the admin page in #1371 should also be used on the timecard page to fix the problem. 

This PR does that and adds a test with a saved timecard to be sure that things are displayed correctly after saving.


caleywoods (approved):
LGTM 👍

PR #1385, opened 2021-12-06
mitigate against browser phishing

lauraGgit:
## Description

Include a high-level description of the issue your pull request fixes; include only one issue per pull request. Link to the GitHub issue it resolves.

## Additional information

* Screenshot images
* Notes


neilmb (approved):
I'm assuming that this is the result of grepping for `target=_blank`, but all of these look good. The Dockerfile and plotly.js changes seem unrelated to the intent of the PR. Do those need to stay?

lauraGgit:
@neilmb the plotly changes were also the result of the grep to add the norel attribut. We might consider trying to improve that upstream. 

The Dockerfile change was because the dockerfile update was breaking local development since the `Pipfile.lock` wasn't happy with conflicting python versions and the deployed version I believe is at 3.9.

PR #1388, opened 2021-12-08
add django-csp based on doj-crt policies

lauraGgit:
## Description

Add a CSP policy for our site. Not entirely sure how to validate this.

## Additional information

using mozilla's django-csp to set a production middleware that sets the apps CSP, heaviliy leveraging the csp used on [doj-crt](https://github.com/usdoj-crt/crt-portal/blob/master/crt_portal/crt_portal/settings.py#L330)


neilmb (approved):
Love it, thanks to packages that do just what is needed and @lauraGgit knowing how to use them.

PR #1401, opened 2022-01-06
add nonce to inline script for csp - fix css

lauraGgit:
## Description

add nonces to inline script tags so that they will be considered safe against xss attacks. These fixes will be necessary for closing out security poams.

Fix css svg so we can finally get the cute clocks that @hbillings put in forever ago.

## Additional information

* Screenshot images
* Notes


neilmb (approved):
LGTM. 

If we weren't limping this along, I'd wonder about an end-to-end test environment where we could debug our CSP setup, but I'm studiously not wondering that.

PR #1406, opened 2022-01-10
remove static from csp because no cdn or s3

lauraGgit:
## Description

remove the local static directory from the `csp` since would be the same as `self`

## Additional information

* Screenshot images
* Notes


cantsin (approved):
LGTM

Issue #1403, opened 2022-01-07
Bunch of squares


beechnut:
![Screen Shot 2022-01-07 at 4 31 19 PM](https://user-images.githubusercontent.com/1887182/148610082-94432e84-edb1-4018-8247-a8f3d6de8978.png)
----
![Screen Shot 2022-01-07 at 4 31 17 PM](https://user-images.githubusercontent.com/1887182/148610086-99b601ad-5b54-4bc0-8c44-9b1fd26418fb.png)
----
![Screen Shot 2022-01-07 at 4 31 23 PM](https://user-images.githubusercontent.com/1887182/148610080-57362cfd-4129-4818-a29e-28d82687d4b9.png)

I'm seeing lots of square blocks that I assume are supposed to be icons that aren't rendering properly.

Using Chrome version 97.0.4692.71

lauraGgit:
Thank you for reporting! Fixed by #1408 .

PR #1408, opened 2022-01-11
Update uswds using `uswds-gulp`

lauraGgit:
## Description

work on migrating tock to uswds 2.13.0 via converting the assets from vendored to pipeline-driven - by leveraging the gulpfile configuration that uswds provides.

This approach includes the autoprefixer.

## This pr includes

- [x] configuration for uswds
- [x] removing uswds v.2.3 (most of the file changes)
- [x] some additional css tweak
- [x] one massive package-lock.json
- [x] update to base.html loading of `uswds.min.js` so that it will avoid a race condition.
- [x] update docs to include uswds updates
- [x] change `sass` docker container name to `gulp` to reflect new container function
- [x] this should also address the squares in #1403


neilmb (approved):
This seems like it was probably hard-won to succeed with upgrading USWDS. Do we want to add any notes in `docs/` about how to accomplish the USWDS version upgrade?

lauraGgit:
@neilmb if you have a change to give another review prior to us merging that would be great.

docs/dependency-management.md
```diff
@@ -43,14 +43,16 @@ The Python version used in production is specified in `runtime.txt`. When updati
 3. Update the `python_version` line in `Pipfile` to ensure dependencies for the correct Python version are used
 4. Update `.circleci/config.yml` to ensure the correct Python version is used in CI
 
+## USWDS
+The uswds is initially installed via `npm` and then is compiled via gulp using the [uswds-gulp](https://github.com/uswds/uswds) configuration. To update USWDS, update the 2.x version in the `package.json` and review the release notes to identify any breaking changes.
```
neilmb, line 47: Thanks for including this here.

tock/tock/static/sass/core.scss
```diff
@@ -17,7 +15,7 @@ $theme-font-path:          "../../vendor/uswds_v2.3.1/fonts";
   position: absolute; // for some reason, USWDS 2.x overrides this to set position as relative, which breaks nav
 }
 
-
+//test comment --lg 11
```
neilmb, line 18: **question:** Is this actually used in a test case or should it be taken out to keep things clean?

neilmb (approved):

PR #1415, opened 2022-01-24
1411 datepicker analytics

### LINKED ISSUE #1412: fix datepicker for analytics
On [the analytics page](https://tock.18f.gov/utilization/analytics)
image this date picker says that it expects a date in `yyyy-mm-dd` format which is inconsistent with the [uswds datepicker](https://designsystem.digital.gov/components/date-picker/) validation.

Acceptance criteria:
As a user looking to query tock usage data, I can put in a date and query the data.
- without recieving this validation error
image
- without recieving a server error (if use the calendar dropdown data picker)

neilmb:
## Description

This closes #1412 by fixing the analytics view to work with the mm/dd/yyyy format that is supported by the USWDS date picker component. It does this by converting the two date query parameters to ISO format where possible. 

## Additional information

The format tip is changed to match:
Screen Shot 2022-01-24 at 14 25 14


lauraGgit (approved):
pushed up plotly fix, and looks good!

PR #1419, opened 2022-01-28
Fix plotly CSP situation with nonces and basic bundle

### LINKED ISSUE #1417: Plotly violates strict CSP
Plotly.js is used on a few pages to make graphs. It violates our CSP policy because it uses inline `<script>` elements that are put on the page by our Django app. Because of this, the plots don't work now.

The solution is to add `nonce=...` to the script from within Django because we have the CSP nonce and we should be able to put it in the script to indicate that the inline script is safe.

neilmb:
## Description

Fixes #1417 by adding a nonce to the inline scripts created by plotly and using the smallest Plotly bundle to remove `data:` images. Also updated to using Plotly's recommended `fig.to_html` method.


lauraGgit (commented):
Bandit which runs as part of the test sweet ironically is flagging the `add_nonce` as potentially introducing the possibility of xss attacks. I don't quite know enough about how django renders html to know whether this effectively ends up as a false positive, but definitely taking a look to make sure.

neilmb:
The Bandit error is a reasonable warning about using Django's `mark_safe` to turn off HTML escaping. But, we are using the plotly library to generate the HTML and so we need to not escape it. I marked these as `# nosec` and included a comment about why we need that.

lauraGgit (approved):
reviewed locally and looks good.
yay to csp friendly loading

PR #1450, opened 2022-05-04
[WIP] Updates

mgwalker:
## Description

This pull requests makes the following platform and necessary dependency updates. It also makes some tweaks for maintainability and sustainability.

## Additional information

- Python 3.9.9 to Python 3.10.1
  - Had to stop at 3.10.1 as that's the latest version currently supported by cloud.gov
  - Let `pipenv` resolve the latest matching dependency versions from `pipfile.lock` and updated all of that (details below)

- Node.js 14.17.3 to Node.js 16.15.0
  - Version is not restricted by cloud.gov because Node.js is only used for building and testing, not runtime
  - Used `npm outdated` to identify dependencies that should be updated (details below)

- Removes gulp tasks related to packaging USWDS sprites
  - The gulp tasks in question were not being executed anywhere, so they were cruft. However, they depended on a library called `gulp-svg-sprite` which has downstream dependencies on PhantomJS. PhantomJS has been deprecated for four years, so removing these tasks also let me remove that ancient dependency. Yay!
- Replaces dependency on `uswds-gulp` with a path
  - This dependency just returns a path to USWDS's `dist` directory, which is fine I guess, except that it also pulls in `gulp-svg-sprite`, which was dragging along PhantomJS. That seems extremely heavy just to get a path, so I nixed it. It might be better in the near future to use `require.resolve()` to get at the right path, but this should work for now, and I think it's clear enough what it does that it should be sustainable.
- Adds `--no-save` flag to `npm install` command for the test Docker container
  - npm somewhat frequently adds metadata to the `package-lock.json` file format. Whenever you run `npm install`, it will check if your `package-lock.json` should be updated with additional metadata. That seems like it should be fine, but it can result in dependencies being updated. Again, that seems fine, but updates should probably be the result of human action, not starting a Docker container. (I also ran into a problem where the container didn't have write permission to its own copy of `package-lock.json` which was my original motivation for this change. In the future, it might be preferable update these two places with `npm ci` runs instead of `npm install`.)

#### Updated Python dependencies

| dependency | old | new |
| ------------ | --- | --- |
| asgiref | 3.5.0 | 3.5.1 |
| beautifulsoup4 | 4.10.0 | 4.11.1 |
| bleach | 4.1.0 | 5.0.0 |
| charset-normalizer | 2.0.11 | 2.0.12 |
| django-webtest | 1.9.9 | 1.9.10 |
| importlib-metadata | 4.10.1 | removed |
| newrelic | 7.2.4.171 | 7.10.0.175 |
| numpy | 1.22.2 | 1.22.3 |
| packaging | 21.3 | removed |
| pandas | 1.3.5 | 1.4.2 |
| plotly | 5.5.0 | 5.7.0 |
| pyparsing | 3.0.7 | removed |
| pytz | 2021.3 | 2022.1 |
| setuptools | 60.8.2 | 62.1.0 |
| soupsieve | 2.3.1 | 2.3.2.post1 |
| urllib3 | 1.26.8 | 1.28.9 |
 | waitress | 2.0.0 | 2.1.1 |
 | whitenoise | 5.3.0 | 6.0.0 |
| zipp | 3.7.0 | removed |

#### Updated Node.js dependencies

Only first-level dependencies are listed here because Node library dependency graphs are... uh... deep.

| dependency | old | new |
| ------------ | --- | --- |
| gulp-svg-sprite | 1.5.0 | removed |
| jest | 27.0.6 | 28.0.3 |
| jest-pupeteer | 3.9.1 | 6.1.0 |
| puppeteer | 1.20.0 | 13.7.0 |
| uswds-gulp | github | removed |

- closes #1446 
- closes #1441
- closes #1440
- closes #1422
- closes #1420
- closes #1410
-


cantsin (approved):
looks fantastic 🤩 let's merge this in and 👀 CI/staging.

PR #1452, opened 2022-05-05
Update PostgreSQL images used in local development

### LINKED ISSUE #1327: PostgreSQL version should be updated to latest minor version
Currently on `9.6.8`
https://github.com/18F/tock/blob/main/docker-compose.yml#L23

Currently supported minor version: `9.6.23`
https://www.postgresql.org/support/versioning/

### LINKED ISSUE #1329: PostgreSQL version should be updated to v10
The final release of current version 9.6 is November 11, 2021
https://www.postgresql.org/support/versioning/

The final release of version 10 is November 10, 2022

mgwalker:
## Description

On cloud.gov, Tock is using postgres 12.8. This PR finishes the work from #1328 and #1330 by getting the local development environment synced up to the same version as the cloud.gov version.

## Additional information

Somewhere between postgres 9 and 12, they disabled password-less accounts by default. You can either re-enable them explicitly or you can give the default user a password as well. I opted to do the latter thing. It only shows up in the `docker-compose.yml` so it was really easy to do.

- closes #1328
- closes #1330
- closes #1329
- closes #1327

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,10 +24,11 @@ jobs:
       # Specify service dependencies here if necessary
       # CircleCI maintains a library of pre-built images
       # documented at https://circleci.com/docs/2.0/circleci-images/
-      - image: circleci/postgres:9.6.8
+      - image: cimg/postgres:12.8
         environment:
           - POSTGRES_USER=circleci
           - POSTGRES_DB=tock-test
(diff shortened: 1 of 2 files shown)


cantsin (approved):
Since this is for local dev/testing only LGTM

PR #1453, opened 2022-05-05
Refactor the CSV export for CSP reasons

### LINKED ISSUE #1418: Export CSV functionality violates CSP policy
The "Export CSV" button that we have with some of our data tables uses an inline `onClick` handler which isn't allowed by our CSP policy. Because of this, that button currently does nothing.

The solution is to restructure the handler so that it comes from a Javascript file that we serve (https://stackoverflow.com/questions/36324333/refused-to-execute-inline-event-handler-because-it-violates-csp-sandbox).

The current click handler in https://github.com/18F/tock/blob/main/tock/tock/static/js/components/save_csv.js uses a random ID from https://github.com/18F/tock/blob/main/tock/utilization/templatetags/analytics.py to find the correct table element on the page. In the new structure, we will need to use DOM-walking starting from the button in the click event to get the nearby table element.

(proposed solution from a discussion with @lauraGgit)

mgwalker:
## Description

Tock's CSP policy disallows inline Javascript, so `onclick=""` attributes are forbidden. Those attributes are how CSV exports are triggered on the analytics page. The handler relies on a random pseudo ID that is generated at render time to identify the table being exported.

This PR removes the `onclick` attribute from the button, and adds `data-csv`, `data-csv-id`, and `data-csv-name` attributes. The `data-csv` attribute is just a flag that the button should act as a CSV exporter. `data-csv-id` and `data-csv-name` are populated with the pseudo ID and filename hint, respectively.

Then, after the DOM is loaded, this PR adds code that finds all button elements with a `data-csv` attribute. It iterates over them, pulling out their IDs and names, and adding a click event listener. It uses the ID and name to finally call the `download_table_as_csv` function after the button is cilcked.

## Additional information

- closes #1418

### DIFF
--- a/tock/tock/static/js/components/save_csv.js
+++ b/tock/tock/static/js/components/save_csv.js
@@ -1,8 +1,34 @@
+// Wait until the DOM content is loaded. In testing, it seems this script
+// can be loaded before the DOM is populated, so... wait! In our case, the
+// buttons remain in the DOM forever once they're loaded. This approach does
+// not work if the buttons are removed from the DOM when hidden, but the
+// accordions we have just hide their content when collapsed rather than
+// remove it.
+window.addEventListener('DOMContentLoaded', function () {
+    var downloadButtons = document.querySelectorAll('button[data-csv]');
+
+    for (var i = 0; i < downloadButtons.length; i += 1) {
+        var button = downloadButtons.item(i);
+
+        // Use an IIFE to capture variables. Otherwise all of the button events
+        // will use the values from the last button, which would be wrong. A
+        // classic Javascript quirk...
+        (function () {
+            var table_id = button.getAttribute('data-csv-id');
+            var filename_prefix = button.getAttribute('data-csv-name');
+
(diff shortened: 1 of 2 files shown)


tock/tock/static/js/components/save_csv.js
```diff
@@ -1,8 +1,34 @@
...
+                download_table_as_csv(table_id + "-table", filename_prefix);
+            });
+        })();
+    }
+});
+
 // Quick and simple export target #table_id into a csv
 // inspired by https://stackoverflow.com/a/56370447
 function download_table_as_csv(table_id, filename_prefix) {
-	var csv_string = table_to_csv(table_id);
```
mgwalker, line 4: The Javascript changes from here down are where I ran `prettier` on the code. I tried to configure prettier to match the existing style instead of introducing a bunch of whitespace changes. It seems like most of the changes are whitespace alignment or semicolons.

neilmb (approved):
Looks like a very nice solution to the CSP issue.

PR #1457, opened 2022-05-17
Deploy to tock-staging in staging

mgwalker:
## Description

Currently, staging deployments use the app name `tock` instead of `tock-staging`. The documentation says to use `tock-staging`, and for safety's sake, that's what we'll do. This PR updates the manifest to use `tock-staging`.


cantsin (approved):
LGTM!

PR #1467, opened 2022-08-04
Nmb/snyk rfd

neilmb:
## Description

Snyk identified https://security.snyk.io/vuln/SNYK-PYTHON-DJANGO-2968205 and we need to upgrade Django to at least version 3.2.15. This makes just that change to the Pipfile and locks the new dependencies. The new dependencies are also written out to the vestigial requirements.txt.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -4,7 +4,7 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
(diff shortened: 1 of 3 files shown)


kfoley-18F (approved):
LGTM! 👍

PR #1491, opened 2022-08-31
Change Pipfile to support pyjwt upgrade

JeanMarie-TTS:
## Description

pyjwt is used by cg-django-uaa. Due to s ecurity vulnerability in pyjwt 1,7, we need to upgrade these two packages.

## Additional information

* django test passed
* integration test failed, but it failed even before these changes (tested on an M1 laptop)
*


neilmb:
The Circle CI build is failing because the `Pipfile.lock` is not updated with the changes to the `Pipfile` https://app.circleci.com/pipelines/github/18F/tock/2025/workflows/c92f69bd-093e-49f9-8351-f7481079b665/jobs/7942?invite=true#step-103-29

> Your Pipfile.lock (e6e3a8) is out of date. Expected: (bbe7e5).

Can you run `pipenv lock` and commit the new `Pipfile.lock` as part of this PR?

JeanMarie-TTS:
Committed my local Pipfile.lock to the branch

cantsin:
🤔 can we narrow down the `Pipfile.lock` changes to only these libraries that we are upgrading here (`pyjwt` and `cf-django-ua`)? We're pulling in a lot of other updates because of `*` and I think it's best to try and reduce the update churn here.

JeanMarie-TTS:
Need to discuss our approach to Pipfile. I did the following locally:
`pipenv --rm`
`git checkout main Pipfile`
`git checkout main Pipfile.lock`
`pipenv install --ignore-pipfile`
`pipenv install pyjwt~=2.4`
`pipenv install cg-django-uaa~=2.1.4`
Have not yet committed the changes. But there were other dependencies that got updated in Pipfile.lock as a result of the above commands. So, I'm not sure this is the way to proceed.

cantsin (approved):
LGTM!

doit.sh
```diff
@@ -0,0 +1,12 @@
+docker-compose down
```
cantsin, line 1: Do we want to add this file? Not sure where this came from.

cantsin (changes requested):

neilmb (approved):

cantsin (approved):
LGTM

PR #1501, opened 2022-11-22
upgrade django to 3.2.16

carjug:
## Description

Upgrades Django from 3.2.15 to 3.2.16 to resolve this dependabot security issue: https://github.com/18F/tock/security/dependabot/28


cantsin (approved):
LGTM; looks like `requirements.txt` already has these changes ...

PR #1506, opened 2022-12-30
Cah/1499 update utilization analytics

carjug:
## Description

Adds Weekly Allocation to Utilization Analytics. This includes the addition of two new model fields to the `Timecard` model: `total_weekly_allocation` and `total_allocation_hours` which represent the summed total project allocation % across projects per timecard and the corresponding hours to that percentage respectively. There has also been added calculation methods for each in the `save()` method for the `Timecard` model.

I also modified the displayed Utilization Rate on the Utilization Analytics page to include weekly allocation numbers.

## Additional information
Here is how the new Utilization Analytics will be represented with this PR
Screen Shot 2022-12-30 at 3 36 44 PM


tock/hours/models.py
```diff
@@ -314,6 +319,17 @@ def calculate_target_hours(self):
...
+    def calculate_total_weekly_allocation(self):
+        """
+        Loops through related time card objects and sums total
+        weekly allocation percentage
+        """
+        timecardobjs = self.timecardobjects.all()
+        total_weekly_allocation = 0.00
+        for tco in timecardobjs:
+            total_weekly_allocation += float(tco.project_allocation)
+        return total_weekly_allocation
```
cantsin, line 331: Do we need a test here? Are there any edge cases with DecimalField that we need to account for when summing up? I don't think so but wanted to make sure I wasn't missing something here.

tock/hours/migrations/0066_auto_20221230_1357.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def calculate_total_weekly_allocation(apps, schema_editor):
+    Timecard = apps.get_model('hours', 'Timecard')
+    timecards = Timecard.objects.all()
+
+    for tc in timecards:
+        timecardobjs = tc.timecardobjects.all()
+        if len(timecardobjs) == 0:
+            continue
```
cantsin, line 14: If the timecard doesn't have any objects then it looks like the default for the new fields is null? I'm not sure what happens in this scenario -- did we want the default to be 0 instead?

cantsin (commented):
Looks great! 🎉 I had some questions, mostly about initial/null values and tests, but I think it's almost ready as-is.

tock/hours/models.py
carjug: I don't believe so, because the default for `project_allocation` is `0.00`

tock/hours/migrations/0066_auto_20221230_1357.py
carjug: Oh yes! let's make the default `0`, good call.

tock/hours/models.py
cantsin: Right you are! Sounds good then, no concerns now.

tock/hours/migrations/0066_auto_20221230_1357.py
cantsin: Great, thanks!

cantsin (approved):
LGTM!

PR #1510, opened 2023-01-05
Remove data migration and add null=true to new timecard model fields …

carjug:
This is a fix to get Tock back into a working state as a stop-gap until we can run the data migration on the Timecard data. (currently working on this)

- Adds `null=True` to new weekly allocation fields on the `Timecard` model.
- Removes prior data migration file


mgwalker (approved):
I don't know the context but the code looks good and I trust your judgement on the rest!

carjug:
Thanks @mgwalker 🎉

PR #1523, opened 2023-01-10
increase memory usage and reduce instances to  deploy

carjug:
## Description
increase memory usage and reduce instances to  deploy data migration


cantsin (approved):
LGTM

PR #1525, opened 2023-01-18
Add weekly allocation hours to csv exports

carjug:
## Description

Include Weekly Allocation Hours in csv exports 

## Additional information

New column called "Weekly Allocation Hours" which corresponds to timecard `total_allocation_hours`


cantsin (approved):
LGTM! Do we have any tests for CSV export?

carjug:
@cantsin yes we do :) they need not be modified for this change because the change gets inherited (as far as I can tell).

PR #1533, opened 2023-01-31
Et/added screenshot to logging documentation

edwintorres:
## Description

Added screenshot of the cloud.gov logging system to the logging documentation, so it's more clear where the query is run.


cantsin (approved):
LGTM!

PR #1564, opened 2023-02-28
Update maintenance page for accessibility

### LINKED ISSUE #1535: Style Tock Maintenance Page for Accessibility
When we need to make production changes to the Tock Database, there is a maintenance app that gets deployed in place of Tock. Currently the app displays a page which is just plain (unstyled) HTML. We should fix this for accessibility reasons.

mgwalker:
## Description

Fixes invalid HTML, adds a `lang="en-US"` attribute to the `html` element, adds a heading element, and adds a `main` element to wrap the contents. There are no more errors or warnings in WAVE.

## Additional information

Closes #1535

### DIFF
--- a/tock/maintenance_page/index.html
+++ b/tock/maintenance_page/index.html
@@ -4,7 +4,10 @@
     <title>Tock - Down for Maintenance</title>
   </head>
   <body>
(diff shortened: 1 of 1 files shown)


cantsin (approved):
❤️

Issue #1575, opened 2023-03-31
Timecards can have both `hours spent` and `project allocation` values


cantsin:
To reproduce this bug: 

- Select a hourly project
- Type in a `hours spent` value
- Using the same project selector, select a weekly project instead
- Select a weekly allocation value
- Submit

The resulting timecard will have both values entered. 

To fix this, we want to clear allocation values on the front-end (when switching projects) and add validation for the back-end.

cantsin:
Backend work is complete but still need to do frontend validation.

cantsin:
Closing as we have both frontend/backend validation in 🎉

PR #1576, opened 2023-04-03
Validate billing based on project weekly/hourly status

cantsin:
## Description

Partially fixes https://github.com/18F/tock/issues/1575 by adjusting the backend (and tests) to fail form submission if hourly data is submitted for a weekly project, or vice versa. 

## Additional information

Ideally the user should never see this message, as the front end will clean up project selection values, but _if_ the javascript fails and the user happens to submit mixed hourly/weekly data, here are the error(s) they will see:

### DIFF
--- a/docs/local-development.md
+++ b/docs/local-development.md
@@ -88,7 +88,7 @@ that case, restart the `tock_app` container with a `docker-compose up` command.
 
 ### Testing locally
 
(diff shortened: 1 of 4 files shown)


edwintorres (approved):
I reviewed and tested locally, it worked well.

PR #1579, opened 2023-04-05
Fix hours field to clear the number when a weekly project is selected

edwintorres:
## Description

Clear hours field on the front-end when weekly billing project is selected. It was reported in issue #1575


cantsin (approved):
LGTM

PR #1581, opened 2023-04-06
Added test that the value was reset to blank after weekly billing 

edwintorres:
## Description

Added test that the value was reset to blank after weekly billing was selected


cantsin (approved):
LGTM

Issue #1589, opened 2023-04-19
[utilization] Add `timecard.utilization` current value to tests


jduss4:
[sub-task of #1500]

Before we update utilization in Tock, we need to update tests to document current behavior.

The model `Timecard` has a field, `utilization`, which is populated using [`calculate_utilization`](https://github.com/18F/tock/blob/1ffb93a83a76d642a579fa0d6a036db3297cd679/tock/hours/models.py#L333) on save.

Add to existing tests to capture the current expected values from `timecard.utilization`.

NOTE: there are multiple places where utilization is being calculated throughout the Tock repo, but `timecard.utilization` does not appear to be leveraged for those purposes. Therefore, existing utilization tests likely do not fulfill this issue (though I might be wrong)!

**Done when**

- [ ] `timecard.utilization`'s calculated value is being documented in at least one existing test

PR #1596, opened 2023-04-24
expands tests of timecard model

jduss4:
## Description

Adds tests to the timecard model to test `calculate_utilization` functionality for values such as `timecard.billable_hours` and `timecard.utilization`.

closes https://github.com/18F/tock/issues/1589

## Question

It appears to me that when `timecardObjects` are updated, the associated timecard itself is not re-saved, therefore fields such as `timecard.billable_hours` and `timecard.target_hours` will be out of sync after the timecard's initial creation. Is this an accurate observation?


cantsin:
> It appears to me that when timecardObjects are updated, the associated timecard itself is not re-saved, therefore fields such as timecard.billable_hours and timecard.target_hours will be out of sync after the timecard's initial creation. Is this an accurate observation?

Which situation are we talking about? Users do not have the ability to update individual details after submission (which is what we want; we don't want an in-progress submission to impact timecard calculations). Or, are we talking about an alternative case where an admin edits a `TimecardObject` directly?

cantsin (approved):
LGTM

jduss4:
@cantsin I believe it would only come up when an admin updates a timecard object, although I haven't explored that workflow to see how it looks from the admin side.  I will make a note to look more into that to see if it's controlled by frontend / interface logic and open an issue if it looks like timecard fields dependent on timecard objects may be falling out of sync.

PR #1597, opened 2023-04-25
pin codecov to 2.1.13

cantsin:
## Description

Pin codecov to `2.1.13` due to `2.1.12` being pulled from pypi. See https://about.codecov.io/blog/message-regarding-the-pypi-package/ for the details.


edwintorres (approved):
LGTM

Issue #1590, opened 2023-04-19
[utilization] Add tests for current utilization views with combination of weekly and hourly billing


jduss4:
[sub-issue of #1500 ]

Before we update utilization in Tock, we should update tests to document the current behavior.

Tests covering utilization values in the views exist already, but do not incorporate weekly allocation billing. Add to tests or create new tests to check what information is presented to the user in the following places, and be sure to consider that some of them may span time ranges larger than a single reporting period:

- [ ] /utilization
- [ ] /employees/[name]
- [ ] /utilization/analytics

Use your judgment here about how and what to test. The main purpose is to make sure that we understand how weekly billing is currently captured as utilization (likely 0%) and document that in a test so the infrastructure for test-driven development is in place when we want to update the utilization behavior.

jduss4:
Started work on this in https://github.com/18F/tock/tree/utilization-test-util-view

Issue #1500, opened 2022-11-16
[Epic] Update https://tock.18f.gov/utilization tables to account for weekly billing


kfoley-18F:
As a person monitoring 18F staff utilization
I want to view utilization calculations that accommodates both hourly billing and weekly allocation
so that I have an accurate picture of how the target hours of staff are being used.

## Sub-issues

**Discovery**
- [x] [Understanding the impact to the billing process](https://github.com/18F/tock/issues/1588)

**Reinforce current tests to prep for development**
- [x] [Add `timecard.utilization` current value to tests](https://github.com/18F/tock/issues/1589)
- [x] [Add tests for current utilization views with combination of weekly and hourly billing](https://github.com/18F/tock/issues/1590)

**Develop!**
- [x] [Update the `timecard.utilization` field value's calculation to include weekly billing](https://github.com/18F/tock/issues/1591)
- [x] [Alter the utils to use the `timecard.utilization` field](https://github.com/18F/tock/issues/1592) (relies on #1591) (https://github.com/18F/tock/pull/1625)
- [x] Double check all documentation about utilization is updated (https://github.com/18F/tock/pull/1619)

## Original issue

https://tock.18f.gov/utilization is set up to roll up hours by chapter/department. It doesn't handle weekly billing, but it may need revisiting on how to show allocation + hours over a whole year. 
Let's revisit how this data is displayed to capture both weekly and hourly billed work.

neilmb:
Need to decide whether to convert allocation percentages into "hours" to get a Utilization rate for each person, or keep them separate so you would have two utilizations?

neilmb:
This page is `GroupUtilizationView` in `tock/utilization/views.py` and the hours data comes from `unit_billing_context()` in `tock/utilization/unit.py`.

carjug:
Possibly convert all of the numbers in the chart to an allocation (%) and have it roll up to an average % allocation.
Where this gets tricky is to convert hourly folks into % allocations....
And also allocation rate (%) vs non-billable or OOO time (hours)

mgwalker:
It sounds like this needs design and engineering work?

jduss4:
I chatted with Kelsey a bit about the requirements for this ticket and did a little [rough information gathering in a document.](https://docs.google.com/document/d/1EeUHQO86HX1dN29POiEeOeGapN9jbTf6ZaSKDLPGMyQ/edit) 🔒 

[...]

PR #1598, opened 2023-05-01
Utilization test util view

jduss4:
## Description

Adds tests to `utilization/tests/test_views.py` that incorporate weekly billing. Currently, the calculation of an individual / organization's utilization does not take into account billable weekly allocation. These tests are intended to confirm the expected behavior of utilization ignoring weekly billing.

- adds more setup by way of projects, users, and timecards
- amends unsubmitted test to include newly added timecards
- tests that a billable weekly project with no additional hours will consider utilization as 'No hours submitted.'
- tests that weekly + hourly calculates utilization only with hours

Closes https://github.com/18F/tock/issues/1590 
This is part of the [utilization epic](https://github.com/18F/tock/issues/1500)

(Tagging @alexbielen and @kingcomma from the TLC crew for their awareness)

### DIFF
--- a/tock/utilization/tests/test_views.py
+++ b/tock/utilization/tests/test_views.py
@@ -50,15 +50,16 @@ class TestGroupUtilizationView(WebTest):
         test_org = Organization.objects.get_or_create(id=1)[0]
         test_unit = Unit.objects.get_or_create(org=test_org)[0]
 
-        self.user = User.objects.create(
-            username='regular.user',
-        )
-
-        # When we create the user, we have to assign them a unit from test data
-        # or else we can't find them in test data.
+        # create a set of users and assign them a unit from the test data
+        self.user = User.objects.create(username='regular.user')
         self.user_data = UserData.objects.get_or_create(user=self.user)[0]
         self.user_data.unit = test_unit
         self.user_data.save()
+
+        self.user_with_no_hours = User.objects.create(username='user.no.hours')
+        self.user_data_no_hours = UserData.objects.get_or_create(user=self.user_with_no_hours)[0]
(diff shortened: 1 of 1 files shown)


tock/utilization/tests/test_views.py
```diff
@@ -219,6 +280,58 @@ def test_includes_user_no_longer_with_unit(self):
...
+        has not submitted hours
+        """
+        response = self.app.get(
+            url=reverse('utilization:GroupUtilizationView'),
+            user=self.user_weekly_only
+        )
+
+        last_week_data = response.context['object_list'][0]['utilization']['last_week_data']
+
+        data = next(item for item in last_week_data if item['username'] == 'user.weekly.only')
```
cantsin, line 299: Did you have a reason for using `next`? The use of `next` here is slightly un-idiomatic i think? it's boring but probably clearer to write `data = [item for item in last_week_data if item['username'] == 'user.weekly.only'][0]`? I suppose to avoid the `[0]` you could also de-structure: `[x,_] = ...` but I think it's better to be boring/plain here.

cantsin (approved):
LGTM with one suggested change that I don't feel too strongly about

tock/utilization/tests/test_views.py
jduss4: My only reason was that it was the top answer on a [stack overflow question](https://stackoverflow.com/questions/8653516/search-a-list-of-dictionaries-in-python) and I don't know python, so I wasn't sure what was idiomatic or not!  I was hoping for something more like `.find()` if we're being honest :)  I will flip it to your suggestion, as it makes more sense to me when reading it.

Issue #1591, opened 2023-04-19
[utilization] Update timecard utilization calculation to include weekly billing


jduss4:
[sub-issue of #1500]

The model Timecard has a field, utilization, which is populated using [calculate_utilization](https://github.com/18F/tock/blob/1ffb93a83a76d642a579fa0d6a036db3297cd679/tock/hours/models.py#L333) on save.  This calculation only takes into consideration hourly billing, and does not incorporate weekly allocation.

- [ ] Determine how to add weekly allocation into the equation ([see notes previously collected on this topic](https://docs.google.com/document/d/1EeUHQO86HX1dN29POiEeOeGapN9jbTf6ZaSKDLPGMyQ/edit#) 🔒 ) and confirm with Kelsey
- [x] Alter tests to expect a different utilization value when timecard has weekly allocation
- [x] Change the behavior of `calculate_utilization`

Issue #1592, opened 2023-04-20
[utilization] Alter the utils to use the timecard.utilization field


jduss4:
[sub-issue of #1500]

We need to update employee, unit, and organization utilization calculations to include weekly allocation alongside hourly billing. Once https://github.com/18F/tock/issues/1591 is completed, timecards will bear an accurate utilization amount. Determine if it is possible to replace the current utilization values calculated in the utilization app with the timecard field values and update the utilization display in the following places:

- `/utilization` by person / chapter / org across multiple time ranges
- `/employees/[name]` across multiple time ranges and specific to projects
- `/utilization/analytics` by unit and org across custom time ranges

I am not sure yet if it is possible to use the timecard utilization value to replace the existing calculations, but I really hope so!  If not, use your time working on this issue to explore how it might be updated to include weekly allocation alongside hourly billing using the existing `utils`.

**Done when**
- [ ] tests modified to include the desired utilization values are passing
- [ ] utilization for [person / unit / org] with weekly allocations across [time ranges] displays correctly
- [ ] any `utils` no longer being used are removed

PR #1599, opened 2023-05-02
calculates timecard utilization including weekly allocation

jduss4:
## Description

Incorporates weekly allocation into the `Timecard.utilization` field calculation. Closes https://github.com/18F/tock/issues/1591. Follow-on work is planned for utilizing this new calculation in reports and views (#1592).

Changes the following logic in the `Timecard` model:
- `total_weekly_allocation` considers billable weekly projects only
- `target_hours` is now based on the `HOURS_IN_A_REGULAR_WORK_WEEK` setting and no longer on the sum of billable and non-billable hours (which could be zero for users on weekly billing projects)
- `utilization` calculation sums `total_weekly_allocation` and hourly utilization (`billable_hours`/`target_hours`)

Updates existing tests to accomodate utilization changes and adds:
- test fixture for non-billable weekly project
- test for no billable hours nor weekly allocation
- test for no billable hours and non-billable weekly allocation
- test for billable hours and non-billable weekly allocation
- tests for combinations of excluded hours (i.e., OOO) and weekly/hourly projects

## Additional information

There are a couple changes/questions we especially want to highlight for review:
- After these changes, users who tock to weekly billing projects will have timecards with different (we believe _correct_) target hours numbers. E.g., a user with no OOO and tocking only to weekly billing projects would have 0 target hours before these changes and 40 * .8 = 32 target hours after these changes. All tests pass after this change, but we're unsure if there are critical areas where the target hours number is used that we aren't aware of.
- `utilization.utils.calculate_utilization` [checks for a case where `target_hours` is `None`](https://github.com/18F/tock/blob/9817837b3b80bb9ec4b3057e5e009f0eddd4b85e/tock/utilization/utils.py#L11-L19), but this is no longer possible after these changes. It doesn't appear this was possible _before_ these changes either, but we want to make sure we aren't missing a critical scenario where `target_hours` really should be `None`.

Tagging @alexbielen for awareness.

(Post updated by @kingcomma on 5/8)

### DIFF
--- a/tock/hours/models.py
+++ b/tock/hours/models.py
@@ -289,9 +289,6 @@ class Timecard(models.Model):
 
         super().save(*args, **kwargs)
 
-    def calculate_hourly_utilization_denominator(self):
-        return self.billable_hours + self.non_billable_hours
-
     def calculate_submitted_date(self):
         """
         If the submitted status is false, make sure submitted_date is unset
@@ -309,15 +306,21 @@ class Timecard(models.Model):
 
     def calculate_target_hours(self):
         """
-        If a user works over 40 hours in a week, their target hours
-        does not increase.
+        A user's target hours for a given timecard is the number of hours
+        in a regular work week (i.e., 40), less any excluded hours (e.g.,
+        OOO time), multipled by their billable expectation. Target hours
+        are rounded to the nearest whole hour.
 
-        Therefore, set `target_hours` to the lesser of:
-            1. Target hours in a regular 40 hour work week
-            2. Billable expectation * number of hours worked
+        Billable hours, non-billable hours, and weekly allocation are not
+        factored into target hours. This supports calculating utilization
+        rates across both hourly and weekly projects.
+
+        Capping the number of hours worked in a week at 40 prevents target
+        hours from increasing for users who work > 40 hours in a given week.
         """
-        this_weeks_target = round(self.billable_expectation * self.calculate_hourly_utilization_denominator(), 0)
-        return min(this_weeks_target, self.max_target_hours())
+
+        total_hours_worked = settings.HOURS_IN_A_REGULAR_WORK_WEEK - self.excluded_hours
+        return round(self.billable_expectation * total_hours_worked, 0)
 
     def calculate_total_weekly_allocation(self):
         """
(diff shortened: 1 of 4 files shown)


tock/hours/models.py
```diff
@@ -316,24 +316,35 @@ def calculate_target_hours(self):
...
+        """
         if self.target_hours == 0:
-            return None
-        return self.billable_hours / self.target_hours
+            if self.total_weekly_allocation == 0.00:
+                return None
+            else:
+                return self.total_weekly_allocation
+        else:
+            hourly_utilization = float(self.billable_hours / self.target_hours)
```
kingcomma, line 346: Thought: For a user not on a weekly billing project, `self.target_hours` can be relied upon to either be a [user's billable expectation multiplied by the max hours in a work week setting](https://github.com/18F/tock/blob/utilization-timecard/tock/hours/models.py#L349-L350) or 0 (if the user [has not yet submitted their timecard](https://github.com/18F/tock/commit/1b6c1475c5ed63ca7050cb7971b0f3c1e98203b4)). 

For users with a weekly billing project, this falls apart. `self.target_hours` can now be any value from 0 to expectation * max hours in a work week (generally 32?). This is the case where a user has a weekly billing project on their time card and tocks a few hours to a billable hourly project (e.g., picks up a help wanted request). For example:

* 1 weekly project at 100%
* 1 billable project at 2 hours
* `self.target_hours` = 2 hours (2 billable hours + 0 non-billable hours)

In this case, you'd end up with a utilization of 200%: 100% weekly + 100% hourly (2 billable hours / 2 total hours).

The appropriate change to account for this *might* be to adjust `calculate_target_hours` to return either 0 if the user has not submitted a timecard or `max_target_hours` if they have ([it's not totally clear to me why this is not already the approach](https://github.com/18F/tock/pull/1038)). In that case, the example above would result in a combined utilization of 106.25% (2 / 32 = .0625 hourly utilization + 1.0 weekly utilization).

jduss4: Thaaat's a real good point.  I think billable hours divided by (max_target_hours - OOO time) could work?

kingcomma: Aha, OOO time is why `calculate_target_hours` doesn't just return 0 or `max_target_hours`! My explanation in my first paragraph was wrong then, `self.target_hours` can be other values in the case where a user tocked OOO or other time marked as `exclude_from_billability`.

cantsin (approved):
I think these new changes are sensible and I appreciate the extensive tests as well 😄 Do we want to do more hands-on tests in staging (`https://tock.app.cloud.gov/`)?

kingcomma:
@cantsin Hands-on testing in staging seems worthwhile to me. I'm specifically thinking of spot checking the values in the "Recent Tocks" table for a few users. Are there other bits you can think of that would benefit from manual testing / review?

cantsin:
Spot checking values SGTM -- let's do it!

PR #1601, opened 2023-05-08
egress documentation and scripts

cantsin:
## Description

Document the egress control process.

### DIFF
--- a/bin/run.sh
+++ b/bin/run.sh
@@ -44,5 +44,10 @@ echo "${DEPLOYMENT_DESCRIPTION}"
 # Record deployment using the New Relic Python Admin CLI
 newrelic-admin record-deploy "${NEW_RELIC_CONFIG_FILE}" "${DEPLOYMENT_DESCRIPTION}"
 
+# make sure python accepts the cloud.gov certificate authority file for proxy
+# (egress traffic control) purposes
+echo "Updating cloud.gov certificate authority (if needed)"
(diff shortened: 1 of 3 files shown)


bin/run.sh
```diff
@@ -44,5 +44,10 @@ echo "${DEPLOYMENT_DESCRIPTION}"
 # Record deployment using the New Relic Python Admin CLI
 newrelic-admin record-deploy "${NEW_RELIC_CONFIG_FILE}" "${DEPLOYMENT_DESCRIPTION}"
 
+# make sure python accepts the cloud.gov certificate authority file for proxy
```
cantsin, line 47: alternatively we could implement this as a cloud.gov [app task](https://docs.cloudfoundry.org/devguide/using-tasks.html) but I think it's a bit cleaner to embed in `run.sh` -- but open to discussion on this.

kfoley-18F: I agree with this choice. I suspect it also makes app backup+recovery clearer if there's an outage and we need to reconstruct Tock.

kfoley-18F (approved):
👏 👏 👏 
Thank you!!

edwintorres:
This part of tock is new to me, but I read the documentation and it's clear. 🚀

cantsin:
I'm closing this for now since we need to re-evaluate how we approach egress vis a vis staging Tock (not the staging environment, but how to actually get Tock started).

PR #1609, opened 2023-05-18
Utilization detail view

jduss4:
## Description

adds or updates a timecard's weekly allocation amount to the UI
updates view test to include allocation total

## Additional information

On the recent timecards view, I added weekly allocation which ONLY includes billable weekly allocation (comes from the timecard model).

image

Note: there was already an existing allocation field on the `reporting_period_user_detail` page which includes both billable and non-billable allocation (the info comes from the timecardobject model).  However, at the bottom where information is provided to explain the utilization calculation, only billable allocation is included. To distinguish, I left the top "project allocation" and made the bottom "billable allocation"

image

### DIFF
--- a/tock/hours/views.py
+++ b/tock/hours/views.py
@@ -802,6 +802,7 @@ class ReportingPeriodUserDetailView(PermissionMixin, DetailView):
         context = super(
             ReportingPeriodUserDetailView, self).get_context_data(**kwargs)
         context['user_billable_hours'] = timecard.billable_hours
+        context['user_weekly_allocation'] = timecard.total_weekly_allocation
         context['user_non_billable_hours'] = timecard.non_billable_hours
         context['user_excluded_hours'] = timecard.excluded_hours
         context['user_target_hours'] = timecard.target_hours
--- a/tock/hours/tests/test_views.py
+++ b/tock/hours/tests/test_views.py
@@ -3,7 +3,6 @@ import datetime
 from decimal import Decimal
 
 import hours.models
-import projects.models
 from api.tests import client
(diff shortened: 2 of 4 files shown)


tock/hours/views.py
```diff
@@ -802,6 +802,7 @@ def get_context_data(self, **kwargs):
         context = super(
             ReportingPeriodUserDetailView, self).get_context_data(**kwargs)
         context['user_billable_hours'] = timecard.billable_hours
+        context['user_weekly_allocation'] = timecard.total_weekly_allocation
```
kingcomma, line 805: Definitely out of scope for this work, but noting that the `total_weekly_allocation` name is becoming increasingly imprecise as we continue the utilization work. Something like `billable_allocation` is probably more descriptive (as hinted by the new "Billable Allocation" table header) and consistent with the `billable_hours`/`non_billable_hours` nomenclature that predates weekly billing.

kingcomma:
Looks great to me!

tock/hours/views.py
jduss4: I agree, the problem began when I added a filter so that field would only be billable, but the name didn't change.  Let's make a ticket!

cantsin (approved):
LGTM!

cantsin:
@jduss4 are we comfortable having these utilization view changes go out with a release next week (assuming #1610 goes in?) if so, i'll merge.

jduss4:
@cantsin we need to talk to one person to make sure that changing the target_hours doesn't impact the billing spreadsheet in unexpected ways, but otherwise this whole gaggle of utilization work should be good to go!

PR #1610, opened 2023-05-18
Add additional utilization tests for common business cases

### LINKED ISSUE #1607: [utilization] Add tests for common timecard scenarios
After #1591 a timecard's utilization calculation now accounts for weekly projects. Tests were added to account for this change, but discussion with @cantsin and @kfoley-18F revealed additional use cases that are not explicitly tested.

Add tests to cover the following scenarios:

- [ ] Cases where a timecard has 4 or 5 weekly billing projects allocated at 12.5% or 25% (e.g., 4 projects with 12.5%, one project with 25%); common amongst account managers
- [ ] Case where a user tocks only OOO time for a full week (should result in 0% utilization)
- [ ] Cases where a user's billable expectation is different than the default .8 (i.e., .4 for a supervisor), including a case where it is 0

We also considered adding tests for users with an alternative work schedule (AWS), but more research is needed to understand how these users tock week-to-week.

Blocks #1592; sub-issue of #1500

kingcomma:
## Description

Adds additional utilization tests for common business cases:

- Timecard with many weekly billing projects (e.g., an AM)
- Timecards with 40 hours of out of office time
- Timecards with billable expectations < .8

Closes #1607

### DIFF
--- a/tock/hours/tests/test_models.py
+++ b/tock/hours/tests/test_models.py
@@ -293,7 +293,7 @@ class TimecardTests(TestCase):
         self.assertEqual(self.timecard.utilization, Decimal('1.75'))
 
     def test_time_card_utilization_only_weekly_billing(self):
-        """Check the utilization is 100% with a 100% weekly billing project
+        """Check that utilization is 100% with a 100% weekly billing project
         and no hourly billing associated with the timecard"""
 
         weekly_project = Project.objects.get(name="Weekly Billing")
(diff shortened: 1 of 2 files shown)


jduss4:
This looks great!  I am so glad that the OOO + weekly allocation test works out without needing any more changes to the timecard model!

cantsin (approved):
yay!! 🎉 LGTM even as a draft PR.

kingcomma:
@cantsin I marked this as ready for review and haven't changed anything since your approval. I was going to try and squeeze in some alternative work schedule tests to this PR, but I think that will take a little longer than I expected and just hold this work up.  I'm 👍 if you're 👍 .

Issue #1612, opened 2023-05-18
[utilization] Adjust total_allocation_hours calculation


jduss4:
Related to epic #1500 

Currently, [`timecard.total_allocation_hours`](https://github.com/18F/tock/blob/main/tock/hours/models.py#L369) uses a [hardcoded value from `base.py`](https://github.com/18F/tock/blob/main/tock/tock/settings/base.py#L154) of 32 hours to calculate the equivalent amount of hours the _billable_ weekly allocation represents.

We may want to adjust this calculation to begin with the [`timecard.target_hours`](https://github.com/18F/tock/blob/main/tock/hours/models.py#L307) value instead of the hardcoded base.py value.

Done when:
- [x] total_allocation_hours has been altered to correctly estimate a weekly to hourly conversion
- [x] tests have been added to or updated to reflect the new field values

cantsin:
I agree -- I think hard-coding 32h/week was a small mistake, in hindsight (the relevant PR is from December 2022). Besides, we've had general 36h/week billable requirements in the past (and to be clear, not expecting that to happen again!). Anyway doing it on the timecard level makes sense. CC @kfoley-18F

Issue #1611, opened 2023-05-18
Update utilization.md to document how weekly projects affect utilization


kingcomma:
As the work towards #1500 proceeds, the documentation in `utilization.md` is becoming increasingly out of date. Update the doc to include:

- Description of how allocation to weekly billing projects is included in the utilization calculation
- Scenarios/examples for users tocking to weekly projects
- Definition of `total_weekly_allocation` value
- Updated definitions for [target hours](https://github.com/18F/tock/blob/main/docs/utilization.md#target-billable-hours) and [utilization](https://github.com/18F/tock/blob/main/docs/utilization.md#utilization)
- Description of how utilization works for users an alternative work schedule (AWS)
- Description of behavior for users with a 0% billable expectation (i.e., 18F Director)

PR #1618, opened 2023-05-23
updates timecard allocation hours calculation

jduss4:
## Description

Changes a timecard's `total_allocation_hours` calculation. Previously, this used a hardcoded value from `base.py` (currently 32 hrs). Instead, we start with the timecard's `target_hours` before multiplying that by the (billable) weekly allocation percentage to create an estimate of the allocation hours.

## Additional information

Closes https://github.com/18F/tock/issues/1612

## Concerns

`base.py` still contains `FULLTIME_ALLOCATION_HOURS` as a setting, as it is used in a migration. If not for the migration, it could be removed.

### DIFF
--- a/tock/hours/models.py
+++ b/tock/hours/models.py
@@ -322,6 +322,14 @@ class Timecard(models.Model):
         total_hours_worked = settings.HOURS_IN_A_REGULAR_WORK_WEEK - self.excluded_hours
         return round(self.billable_expectation * total_hours_worked, 0)
 
+    def calculate_total_allocation_hours(self):
+        """
+        Calculates an estimate of the hours spent on projects with a weekly billable
+        allocation by multiplying that allocation against this timecard's target hours.
+        """
+
+        return round(self.target_hours * self.total_weekly_allocation, 0)
+
     def calculate_total_weekly_allocation(self):
         """
         Loops through related time card objects and sums total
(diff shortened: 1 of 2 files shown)


tock/hours/models.py
```diff
@@ -322,6 +322,14 @@ def calculate_target_hours(self):
         total_hours_worked = settings.HOURS_IN_A_REGULAR_WORK_WEEK - self.excluded_hours
         return round(self.billable_expectation * total_hours_worked, 0)
 
+    def calculate_total_allocation_hours(self):
+        """
+        Calculates an estimate of the hours spent on projects with a weekly billable
```
kingcomma, line 327: Non-blocking question: Is this just an "estimate" because the allocation percentage is itself not a precise measurement? I.e., allocation represents how a user has been staffed and not necessarily an exact measurement of billable hours?

jduss4: My understanding is that a person with 100% allocation on a project could be working anywhere between 32 and 40 hours, but it's not possible for us to know exactly what.  When we talked to Matt S the other day, I think he said he's using 34 (or was it 36?) hours when utilization is 100%.

kingcomma: Got it, thanks for clarifying!

I don't have any suggestions for changes to this PR, but I think _I_ should include a callout in my documentation updates (#1611). Something under "billable allocation hours" that notes the intent and limitations of the derived value. I.e., "We need a number so we can perform utilization rate calculations across dimensions that include varying combinations of hourly, weekly, and excluded hours projects. The powers that be have decided x, y, z for translating allocation to hours for this purpose."

Oh, and yeah, my notes say Matt S treats 100% as 36.

cantsin (approved):
LGTM

PR #1619, opened 2023-05-24
Updates to utilization documentation with definitions and examples for weekly billing

### LINKED ISSUE #1611: Update utilization.md to document how weekly projects affect utilization

kingcomma:
## Description

Aligns `docs/utilization.md` with recently updated behavior to include weekly billing projects in utilization rate calculations (see #1591 and #1618). Also includes some clean up to consistently handle capitalization, pseudo-code, and variable names.

Closes #1611 

## Additional information

Would appreciate a gut check on whether or not the "Weeks 1-4" example is unwieldy now. It's the only example that uses anything other than simple arithmetic.

### DIFF
--- a/docs/utilization.md
+++ b/docs/utilization.md
@@ -75,7 +75,7 @@ Hourly Non-Billable | 8.0 | 2.0 | 8.0 | 8.0
 ### Week 1
 Stacey's `utilization` for Week 1 is 100%.
 
-Stacey tocked 24 billable hours to an hourly project and 8 billable hours to a weekly project, for a combined 32 hours of an expected 32 target billable hours.
+Stacey tocked 24 billable hours to an hourly project and 8 billable hours (25% of 32) to a weekly project, for a combined 32 hours of an expected 32 target billable hours.
 
 ```python
 # Week 1
--- a/docs/utilization.md
+++ b/docs/utilization.md
@@ -6,23 +6,23 @@ Tock utilization reports display progress towards billing targets for each user,
 
 Each user has a billing target which is the percentage of worked hours that are expected to be tocked to billable projects.
 
-We calculate utilization, or progress, towards these targets by totaling the hours worked on billable projects (both hourly and weekly) and dividing by the expected number of billable hours worked.
+We calculate utilization, or progress, towards these targets by totaling the hours worked on billable projects and dividing by the expected number of billable hours worked.
 
 A few more detailed definitions and example calculations are included below.
 
 ## Definitions
 
-### Utilization
+### Utilization for a single timecard
 
 The sum of two percentages:
-1. For hourly projects, the percentage of [billable hours](#billable-hours) to [target billable hours](#target-billable-hours) tocked over a period of time
-2. For weekly projects, the percentage of [billable allocation hours](#billable-allocation-hours) to target billable hours tocked over a period of time
+1. For hourly projects, the percentage of [billable hours](#billable-hours) to [target billable hours](#target-billable-hours) tocked to the timecard
+2. For weekly projects, the total [billable allocation](#billable-allocation) tocked to the timecard
 
-Utilization | Description
-------------|-------------
-< 100% | a user has not reached their billable goal
-100% | a user has reached their billable goal
-\> 100% | a user has exceeded their billable goal
+### Utilization in aggregate
+
+The sum of [billable hours](#billable-hours) and [billable allocation hours](#billable-allocation-hours) divided by the sum of [target billable hours](#target-billable-hours).
+
+This formula is used when calculating utilization across durations (e.g., the past month) and groups (e.g., all account managers).
 
 ### Billable hours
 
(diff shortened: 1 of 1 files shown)


docs/utilization.md
```diff
@@ -2,157 +2,201 @@
...
-Non-Billable | 16.0 | 4.0 | 8.0 | 8.0
+Hourly Billable | 24.0 | 30.0 | 0.0 | 32.0
+Weekly Billable | 25% | 0% | 100% | 0%
+Hourly Non-Billable | 8.0 | 2.0 | 8.0 | 8.0
 
 ### Week 1
-Stacey's `utilization` for Week 1 is 75%.
-Stacey tocked 24 of an expected 32 billable hours.
+Stacey's `utilization` for Week 1 is 100%.
+
```
edwintorres, line 77: > 8 billable hours to a weekly project

Justo be super clear, is 8 the result of 32 hors per 25%, can we add that calculation as a parenthesis here?

docs/utilization.md
```diff
@@ -2,157 +2,201 @@
...
+billable_allocation_hours = billable_allocation * target_billable_hours
+# .25 * 32 = 8
 
-utilization = billable_hours / target_billable_hours
-# 24 / 32 = 0.75
+utilization = (billable_hours + billable_allocation_hours) / target_billable_hours
+# (24 + 8) / 32 = 1
 ```
 
 ### Week 2
```
edwintorres, line 98: Suggestion: Add the table with the data before  each example, as it difficult to verify the math scrolling up and down constantly to check the numbers.

docs/utilization.md
```diff
@@ -2,157 +2,201 @@
...
 
-### NON-BILLABLE
-An employee is considered `Non billable` if their `TARGET BILLABLE HOURS` over a time period is 0. Utilization is not calculated for these individuals because
-we're currently unable to <a href="https://en.wikipedia.org/wiki/Division_by_zero">divide by zero</a>.
+**Note:** Allocation percentages technically refer to an employee's availability on a project rather than a specific number of hours worked. But we convert this percentage to an _expected_ hourly value to enable utilization rate calculations across durations (e.g., the past month) and groups (e.g., all of `Project ABC`) that may include varying combinations of hourly, weekly, and excluded hours projects.
 
-All `BILLABLE HOURS` tocked by `Non billable` employees are included in aggregations for their respective business unit and organization.
+### Target billable hours
+A user's expected number of hours tocked to billable projects over a period of time. Determined by their [billing expectation](#billing-expectation) over that period and the total number of hours in a week less any hours which were not worked like `Out of office`.
```
jduss4, line 46: [question] Are you using `TARGET_BILLABLE_HOURS` and `BILLING_EXPECTATION` instead of the timecard field names (`target_hours` and `billable_expectation`) to match what was already in the documentation?

docs/utilization.md
```diff
@@ -2,157 +2,201 @@
...
+utilization = (billable_hours + billable_allocation_hours) / target_billable_hours
+# (0 + 32) / 32 = 1
 ```
 
 ### Weeks 1-4
-Stacey's `utilization` over the last 4 weeks is 95%. Stacey tocked 116 of an expected ~122 billable hours over the 4 week period.
+Stacey's `utilization` over the last 4 weeks is ~104%.
 
-Stacey had eight hours of `Out of Office` time in week 2, which is excluded from all totals. Notice the `total_hours` is 152 for the 4 weeks.
+Notice how values for each week are calculated independently and then summed to produce a utilization number over the 4-week time frame. Because the target billable hours denominator may fluctuate week-to-week, we can not simply take the average of each week's utilization percentages.
```
jduss4, line 149: I think that we can just take the average!  I worked this out on paper a few weeks ago to make sure that I wasn't misremembering how math works!  I think maybe what we should mention is that we can't (???) because if someone with no billable expectation did bill hours, their utilization of 0% would not roll up as accurately as if we were summing in those hours with weeks when they had a billable expectation?  Does that make sense?

docs/utilization.md
jduss4: That makes sense to me!

docs/utilization.md
kingcomma: @jduss4 and I discussed offline. We decided that an average of utilization rates will not be accurate if we are calculating utilization as number of billable hours divided by number of target hours across multiple timecards. See the example below.

An average _would_ be accurate if we were trying to answer a question more like "How has a user been allocated to a project over time?" But that's not the case here. It's been decided that utilization rate should be calculated as explained above.

Example:

Value | Week 1 | Week 2 | Aggregate | Average
------|--------|---------|-------|---
Hours worked | 40 | 20 | 60 | 30
Billable hours | 32 | 8 | 40 | 20
Target hours | 32 | 16 | 48 | 24
Utilization | 100% | 50% | 83% | 75%

docs/utilization.md
kingcomma: @edwintorres Yeah, I see what you're saying. Before I make that change, do you have any concerns about this creating a need for the table to be updated in 4 places if the examples change in the future?

docs/utilization.md
kingcomma: Added the parenthetical to the line below

cantsin:
Anything left here to do? I think the PR is good as-is, but just checking!

docs/utilization.md
edwintorres: Yes I see that as a problem, but how much do you think this example table is going to change?

kingcomma:
I'm going to mark this PR as a draft while we figure out some open questions around how aggregate utilization should be calculated (#1626). Once we have an answer there, I'll give these docs a pass and update any descriptions and examples to match the agreed upon approach. I'll also add in the other requested improvements.

PR #1623, opened 2023-05-26
document how we deactivate users

cantsin:
## Description

Every month, we are required to deactivate Tock users who have left the organization. Adding some documentation and a provided script towards this process.


edwintorres (approved):
Thank you for this, very useful. I was no aware of the details of the proces.

PR #1636, opened 2023-06-28
setup egress environment vars

cantsin:
## Description

Trying to simplify the current egress workflow by adding two environment variables:

- `https_proxy` will be set to `egress_proxy` when the app launches
- `REQUESTS_CA_BUNDLE` will force the python libraries to load system ca certificates


edwintorres (approved):
LGTM

Issue #1566, opened 2023-03-02
Accessibility: Complex tables has empty table header cells


sarahmhgov:
Steps to Reproduce:

Issue exists on tables on the following page: 
- https://tock.18f.gov/utilization/analytics

Expected: 
- "Complex" tables (more than two levels of headers). Each header (th) should be given a unique id, and each data cell (td) should have a headers attribute with each related header cell’s id listed.

Actual:
- many of the table header cells are empty. 

Remediation guidance:

I recommend simplifying this table into a simple table, as the complex table functionality is not required to support this use case. 

<img width="786" alt="image" src="https://user-images.githubusercontent.com/124302726/222294948-82c19e32-6c15-411c-8597-87d81b0ab299.png">

### Tasks
- [x] Create a PR to replace the complex table with simple table
- [x] Tag @alexbielen as a reviewer

MelissaBraxton:
@jasnakai  - I added a task list here to make it clear how to get this to done!

jasnakai:
Met with Neil Martinsen-Burrell on the issue and to review a solution. We agreed that using the pandas dataFrame.toHTML functionality is not producing what we want. So, we will abandone that approach and instead use Django templates to build out the table.

jasnakai:
PR: #1640

MelissaBraxton:
Giving this a bump: @edwintorres or @cantsin  - Could you please review this PR when you have a chance?

jasnakai:
[Merged](https://github.com/18F/tock/pull/1640/files/ce8f1303845a03c19a06079d088ba73aa6630201..a25db5af57515513d9280d4e184edf3b8902fbea)

PR #1640, opened 2023-07-05
Fix empty header cells in utilization table

jasnakai:
For ticket: #1566 
## Description
Copied and updated the `frame_table` tag as `df_table` to use a custom table instead of the pandas dataframe.to_html output.

## Additional information
This PR does not address the `Headcount` table issues. The headcount table needs to be reviewed by design to devise a better solution for this information.


alexbielen:
@cantsin or @edwintorres can one of you take a look at this?

edwintorres (approved):
I tested it locally it LGTM.

image

PR #1650, opened 2023-08-03
api: escape query params when emitting error

cantsin:
## Description

when the API emits an error, escape all user input (query params).

## Additional information

this behavior was flagged in a netsparker scan, but it's a false warning. nonetheless, let's fix anyway in the name of sanitizing output.


edwintorres (approved):
I tested locally it is working. This change LGTM.

PR #1655, opened 2023-08-16
syncing up dependency updates

cantsin:
## Description

Pipfile and requirements.txt are slightly out of sync. 

- Updated Pipfile
- `pipenv requirements | sort > requirements.txt`

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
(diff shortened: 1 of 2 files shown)


cantsin:
Note that with zscaler, building with a custom `[[source]]` is blocked. I can't figure out how to pass on the certificates to make this work -- `SSL_CERT_DIR`, `CURL_CA_BUNDLE`, `REQUESTS_CA_BUNDLE`, etc, don't seem to let the certificate pass-through. I suspect it's because of `verify_ssl = true`.

PR #1658, opened 2023-08-17
Syncing up dependency updates - fic

edwintorres:
## Description

Added zope dependencies


cantsin:
For the record, forcing a pass for the snyk check since the vulnerability is in setuptools, which is already at the latest version -- indeed, the explicit fix is to `Pin setuptools to version 65.5.1` ; iow, zope.event here is a false positive. 

```
Introduced through zope.event@5.0
Fixed in setuptools@65.5.1
```

cantsin (approved):

PR #1660, opened 2023-08-18
Workaround to override zope sub-dependency

edwintorres:
## Description

As mentioned in https://github.com/18F/tock/pull/1658#issuecomment-1682936443 a sub-dependency is raising a flag on Snyk review. 

> For the record, forcing a pass for the snyk check since the vulnerability is in setuptools, which is already at the latest version -- indeed, the explicit fix is to `Pin setuptools to version 65.5.1` ; iow, zope.event here is a false positive.
> 
> ```
> Introduced through zope.event@5.0
> Fixed in setuptools@65.5.1
> ```

Tock use `pipenv` which do not provide a way to override that sub-dependency. But the maintainer offered a [workaround](https://github.com/pypa/pipenv/issues/1921#issuecomment-379131833). 

This PR uses that workaround. A long-term fix to this issue could be the use of pyproject.toml, a new [unified Python project settings file](https://www.python.org/dev/peps/pep-0518/). But that option requires more exploratory work.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -22,6 +22,7 @@ plotly = "*"
 pandas = "*"
 pyjwt = "~=2.4"
 django-csp = "*"
+setuptools = "==68.1.0"
 
 [dev-packages]
 bandit = "*"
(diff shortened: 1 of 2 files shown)


cantsin (approved):
amazing work @edwintorres !!

PR #1666, opened 2023-08-23
Add Total Hours Counter even with Weekly Billing Projects in Reporting Period Page #1638

### LINKED ISSUE #1638: Add Total Hours Counter even with Weekly Billing Projects in Reporting Period Page
**Problem**
As a user of the application, I noticed that whenever I add a project with weekly billing on the Reporting Period Page, the total hours counter at the bottom of the page is removed. This feature is extremely useful in tracking the total amount of hours spent across all projects, irrespective of their billing type.

Example when a user has a weekly billing project.

image

Example when a user do not have a weekly billing project

image

**Describe the solution you'd like**
Ideally, the total hours counter should persist on the Reporting Period Page regardless of the billing method used for the project. This feature would improve usability by providing a consistent and holistic view of hours spent across all types of projects at the time of weekly timesheet completion. 

Calculating and displaying total hours is useful for most users. However, AWS schedule users may find the chart's color-coding (which turns red after 40 hours a week) less relevant. Thus, maintaining the chart's visibility, despite color discrepancies with AWS schedules, is more beneficial than the current setup, providing a visual representation of logged hours for all users.

**Describe alternatives you've considered**
An alternative could be to provide separate counters for different types of billing (e.g., one for hourly and another for weekly billing), but it may cause fragmentation and could make the overall view less clear. A unified counter would make tracking and analysis much simpler.

edwintorres:
## Description

To close #1638, the Total hours counter is always visible in this PR. The color was changed to grey while billing hours remained the same. 

image

### DIFF
--- a/tock/tock/static/js/components/timecard.js
+++ b/tock/tock/static/js/components/timecard.js
@@ -150,16 +150,6 @@ function populateHourTotals() {
   billableElement.querySelector('.number-label').innerHTML =
     totals.billableHours;
 
-  if (totals.totalHours === 0) {
-    totalElement.classList.remove('valid', 'invalid');
-  } else if (totals.totalHours === totalHoursTarget) {
-    totalElement.classList.add('valid');
-    totalElement.classList.remove('invalid');
-  } else {
-    totalElement.classList.add('invalid');
-    totalElement.classList.remove('valid');
-  }
-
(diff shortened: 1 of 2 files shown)


cantsin (approved):
LGTM! Let's test in staging 🎉

PR #1703, opened 2023-12-14
adds documentation for importing data into docker

jduss4:
Adds documentation for importing data into docker and clarifies a few steps of the export process.

Please confirm that none of the information that I included in the documentation is sensitive. I believe that it is largely docker-compose defined.

Most of this documentation is taken straight from notes that @kingcomma put together when working on Tock.


cantsin (approved):
LGTM!

PR #1706, opened 2023-12-21
updates egress documentation

jduss4:
Reworks the existing documentation to include steps that I found helpful along the way. Caution: since the egress server is not behaving as expected, it's quite possible some of these steps are wrong!


cantsin (approved):
😍

PR #1730, opened 2024-03-13
Allow zero for project allocation on hourly projects

### LINKED ISSUE #1728: bug: You cannot submit weekly billing for a project under hourly billing
During the last few weeks, we have repeatedly gotten user feedback in `#tock` that the following error sometimes comes up when submitting time cards: 

image

> There was an error processing your time card
> * You cannot submit weekly billing for a project under hourly billing.

We thought we fixed this over at https://github.com/18F/tock/issues/1575 but I think this problem has come up in a different form. In particular, this is now happening (but not exclusively so) for prefilled time cards -- that is, time cards that carry over from the previous week. 

Not sure how to reproduce, yet, but we can start by looking at these affected users' time cards.

neilmb:
## Description

#1728 was tracked down to a situation where the timecard form might submit "0" for project allocation even for hourly billing projects. This changes the form cleaning code to allow that possibility without raising the error message from the bug.

We add a test for the situation that fails in our current code and passes with the fix.

## Additional information

There are a couple unrelated changes where `or None` was removed where it wasn't doing anything.

Fixes #1728

### DIFF
--- a/tock/hours/forms.py
+++ b/tock/hours/forms.py
@@ -170,6 +170,17 @@ class TimecardObjectForm(forms.ModelForm):
     def clean_hours_spent(self):
         return self.cleaned_data.get('hours_spent') or 0
 
+    @staticmethod
+    def _is_empty_or_zero(field):
+        """Check that a field is either the empty string or the number 0."""
+        if field == "":
+            return True
+        else:
+            # not an empty string
+            if Decimal(field) == 0:
+                return True
+        return False
+
     def clean(self):
         if 'project' in self.cleaned_data:
             if self.cleaned_data['project'].is_weekly_bill:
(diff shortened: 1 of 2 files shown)


tock/hours/forms.py
```diff
@@ -183,8 +194,12 @@ def clean(self):
                         ),
                     )
             else:
-                if ('project_allocation' in self.cleaned_data and \
-                   self.cleaned_data['project_allocation'] != ''):
+                if (
+                    "project_allocation" in self.cleaned_data
+                    and not self._is_empty_or_zero(
```
cantsin, line 199: do we need the `staticmethod` decorator here? Shouldn't we call it as `TimecardObjectForm._is_empty_or_zero` to be clearer?

cantsin (approved):
one question but LGTM, we can test in staging

tock/hours/forms.py
neilmb: We don't strictly need to use `@staticmethod`, but I think it's good practice when we don't need to use any properties of the instance or the class. As far as calling the method, `self` works and is much less verbose than having to repeat the class name.

Issue #1615, opened 2023-05-19
server error when viewing analytics of certain projects


cantsin:
As an example, if we visit https://tock.18f.gov/projects/1668/ then we get a server error. A (partial) trace is below:

```
  context = self.get_context_data(object=self.object)
File "/home/vcap/app/tock/projects/views.py", line 48, in get_context_data
  context.update(self.get_analytics(timecard_entries))
File "/home/vcap/app/tock/projects/views.py", line 31, in get_analytics
  plot, table = project_chart_and_table(timecard_entries)
File "/home/vcap/app/tock/utilization/analytics.py", line 241, in project_chart_and_table
...
  TypeError: unsupported format string passed to NoneType.__format__
```

I'm unsure why this bug is triggered; https://tock.18f.gov/projects/1667/ for instance, works fine.

kingcomma:
Happened to run into this one today while testing related functionality. It looks like this is happening for projects that:
1. Have at least one associated `TimecardObject` with an `hours_spent` value of `None` (i.e., the user tocked a `project_allocation` percentage)
1. Have at least one associated `TimecardObject` an `hours_spent` value greater than 0

In those cases, `None` values get past [this conditional](https://github.com/18F/tock/blob/1a5aa2a6e30349931076b263abb81467795a576c/tock/projects/views.py#L47-L48) and make their way to `utilization.analytics.project_chart_and_table` where they cause the TypeError.

I'm not sure how `None` values are getting recorded in `hours_spent`. As far as I know, this scenario shouldn't be happening under normal use: either a project is weekly and users tock 0 hours (the default) or a project is hourly and users tock one or more hours. That said, this is not a constraint that is enforced on the `TimecardObject` model itself. I.e., both hours and allocation can be added to a `TimecardObject` using the admin portal and `hours_spent` can be null.

You can get the list of the projects that meet the criteria above via the Django shell. All of these should show the same error:
```python
projects_with_decimal = TimecardObject.objects.filter(hours_spent__gt=0).values('project')
projects_with_none = TimecardObject.objects.filter(hours_spent=None).values('project')
print(projects_with_decimal.intersection(projects_with_none))
```

alexbielen:
Discussed with Jess and we feel this is low priority but still a good idea to fix at some point.

We're going to move this into the TLC Icebox for now. In the meantime, if you find another way to get this work done feel free to choose that path. If you do find a different path to getting the work done and no longer want TLC crew help, please remove the issue from our project board.

[...]

PR #1731, opened 2024-03-14
Fix projects view when timecards have None hours_spent

neilmb:
## Description

#1615 tracked down a situation where the individual project view can fail when a timecard object has `None` for the `hours_spent` field. This is uncommon, but does happen in our production data.

This fixes that situation by filling with `None`s in the data frame with zeroes which is the more common case for our data.
There is a failing test case that fails without the fix and passes with this change.


cantsin (approved):
LGTM!

Issue #1695, opened 2023-11-13
Finalize egress work


cantsin:
As part of Tock's compliance process, the capability for egress filtering is set up for cloud.gov deployments of Tock. We have set up egress spaces for both staging (which should always be on) and production (not on, but available), but more work needs to be done before we finalize egress for the entire stack:

- [x] Egress functionality does not seem to be stable. Production/staging sometimes breaks (for the former, maybe because of `recycle-prod`) and we'd like to do more testing as to why
- [x] Someone should run through the documentation and make sure it looks sane (especially the cloud.gov documentation), including re-creating egress from scratch in the space and adding more documentation around debugging (e.g., making sure caddy works locally)
- [x] We need more documentation on how to update and check the status of egress filtering
- [x] Because of the nature of egress, which involves sensitive variables, we don't have a good way to actually document our egress setup other than via documentation -- we could look at other projects and see how they address this

(added by @jduss4 )

- [x] Investigate removing the [public-egress](https://cloud.gov/docs/management/space-egress/#public-egress) policy from Tock staging. Currently the app fails to restage when this policy is removed due to connection errors related to New Relic (possibly related to https://github.com/GSA-TTS/cg-egress-proxy/pull/15) (a possible complication: https://github.com/18F/tock/pull/1661)

juliaklindpaintner:
Consider turning this into an epic for TLC crew purposes — will follow up!

alexbielen:
@cantsin and @edwintorres :

Are you still looking for TLC Crew to help on this next increment (Dec 11 - 25)?

edwintorres:
@alexbielen yes that's correct

jduss4:
I put in a little work running through the steps in the existing documentation and adding more to it here:  https://github.com/18F/tock/pull/1706

jduss4:
Unfortunately I was not able to complete this card before the break. However, during this time we the following:

1. Built out the documentation
2. Redeployed the staging versions of egress + tock and confirmed the behavior works for restricting egress
3. Experimented (unsuccessfully) with removing the public-egress application security group from tock staging and identified a new area for investigation in the setup for New Relic

cantsin:
I think we can finally close this issue :) If anything comes up with egress in the future, I'll make a separate ticket. Many thanks to everyone involved!!

PR #1734, opened 2024-03-18
Tell New Relic to use our proxy

neilmb:
## Description

This is one step towards #1695. New Relic doesn't appear to respect the `https_proxy` environment variable, so we specify it explicitly when we call `newrelic-admin`.

## Additional information

This appears to fix New Relic's stopping of deploy when we remove the public egress security group, as long as the proxy is correctly configured.


cantsin (approved):
LGTM

PR #1736, opened 2024-03-18
add egress_proxy configuration files

cantsin:
## Description

Adding egress configuration files for easier reference/deployment.


neilmb (approved):
LGTM. If the deploy works, we could just merge this as-is and fix it if we learn more.

PR #1739, opened 2024-03-20
update jest-puppeteer and fix integration tests

cantsin:
## Description

Puppeteer dependency updates. This is quite the bump so we had to fix some minor breakage in the integration tests.


neilmb (approved):
LGTM. I hand-checked each SHA and they all look correct to me. ;)

Issue #1763, opened 2024-06-05
Add 75% as an option for weekly billing percentages


lauranash18f:
We would like to add 75% as an option for weekly billing allocations. Please let me know what detailed information you need to implement this change.

@Erica-Perkins requested this change, but @lauranash18f can also help answer implementation questions.

PR #1764, opened 2024-06-05
weekly billing: add 75% as an option

cantsin:
## Description

Addresses https://github.com/18F/tock/issues/1763


edwintorres (approved):
finally!! Thanks!

PR #1797, opened 2024-09-19
Added NoCacheMiddleware

edwintorres:
## Description

Added NoCacheMiddleware in response of a pentest finding.

### DIFF
--- a/tock/tock/middleware.py
+++ b/tock/tock/middleware.py
@@ -42,5 +42,6 @@ class NoCacheMiddleware:
         self.get_response = get_response
 
     def __call__(self, request):
+        response = self.get_response(request)
         add_never_cache_headers(response)
         return response
--- a/tock/tock/middleware.py
+++ b/tock/tock/middleware.py
@@ -34,3 +34,15 @@ class AutoLogout(object):
             datetime.now().strftime(fmt)
 
         return self.get_response(request)
+
+
+class NoCacheMiddleware:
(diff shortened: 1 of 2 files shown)


tock/tock/middleware.py
```diff
@@ -34,3 +34,15 @@ def __call__(self, request):
...
+
+class NoCacheMiddleware:
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        response = self.get_response(request)
+        response['Cache-Control'] = 'no-store'
+        response['Pragma'] = 'no-cache'
+        response['Expires'] = '0'
```
cantsin, line 47: Can we use `from django.views.decorators.cache import never_cache` instead?

https://docs.djangoproject.com/en/5.1/topics/http/decorators/#django.views.decorators.cache.never_cache

tock/tock/middleware.py
```diff
@@ -34,3 +34,15 @@ def __call__(self, request):
...
+        return response
```
cantsin, line 48:
```suggestion
        add_never_cache_headers(response)
        return response
```

tock/tock/middleware.py
cantsin: or even better, `from django.utils.cache import add_never_cache_headers`

cantsin (approved):

Issue #1792, opened 2024-08-23
Validate New Relic in ingesting data from Tock


edwintorres:
## Description:

During the Tock Weekly Log Audit, we observed that no alerts have been generated since March of this year. This issue aims to:

<img width="1040" alt="image" src="https://github.com/user-attachments/assets/b94ccc75-ace7-4ef3-a9ad-c1b2523e40e9">

Confirm the absence of alerts since March.
Ensure that New Relic is correctly ingesting data from Tock.

## Tasks:

- [ ] Validate the absence of alerts since March.
- [ ] Verify that New Relic is properly ingesting data from Tock.

nateborr:
In New Relic, via a [manual query](https://onenr.io/0VwgJLnLoQJ) I've confirmed the last date that we recorded metric data from any of the Tock apps:

- staging: March 16, 2024
- production: April 2, 2024

We are still recording Deployment events for both applications: https://onenr.io/0VjYrayr2Q0 If the deployments are being reported on the application's behalf by its build pipeline, that would explain why we have that data without the APM data we'd expect to get from the application instances.

I don't see any alert data in the system at all at this point, although the data from this spring may be past its data retention period. I can't directly confirm the data retention policies as a basic user.

I'll keep digging in later this week. The existence of the deployment data likely rules out issues in the account configuration or subscription on the New Relic side. Since there are a couple of weeks between the last-reported dates for the Tock staging and production applications, it looks like a change that was rolled out to the lower environment first, although license key issues could be another culprit.

nateborr:
This looks like it's related to the egress proxy configuration for #1695. That was released to production on April 2, 2024, which is the day we stopped receiving APM data in New Relic.

Per [New Relic's documentation](https://docs.newrelic.com/docs/new-relic-solutions/get-started/networks/#new-relic-endpoints), their agents send data to endpoints at `collector.newrelic.com`. If we add that domain to our egress proxy allow list I think we'll be in good shape.

We are continuing to receive Deployment events because Tock's [`run.sh` script](https://github.com/18F/tock/blob/main/bin/run.sh) sends them via the [newrelic_admin script](https://docs.newrelic.com/docs/apm/agents/python-agent/installation/python-agent-admin-script-advanced-usage/#record-deploy). That admin script creates deployments via the New Relic REST API and in #1734 we allowed our egress proxy to send that traffic to `api.newrelic.com`.

nateborr:
[...]

PR #1810, opened 2024-10-09
Set NEW_RELIC_PROXY_HOST env var during app initialization

nateborr:
## Description

Set the `NEW_RELIC_PROXY_HOST` environment variable to the egress proxy URL during app initialization.

Previously we set that only for the `newrelic-admin record-deploy` script command on startup, but the New Relic agent likely needs that configuration as well.

Compare to the notifications-api configuration, which is successfully running the New Relic Python agent against their FedRAMP-compliant collector through an egress proxy: https://github.com/GSA/notifications-api/blob/main/.profile#L8

## Testing

Merge the change and deploy it to staging; I'll shake out that container manually once it's there.

### DIFF
--- a/.profile
+++ b/.profile
@@ -4,3 +4,4 @@
 ##
 export http_proxy=$egress_proxy
 export https_proxy=$egress_proxy
+export NEW_RELIC_PROXY_HOST=$egress_proxy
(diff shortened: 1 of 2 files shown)


cantsin (approved):

nateborr:
This relates to #1792.

PR #1811, opened 2024-10-09
Update deployment docs and lower New Relic log level

nateborr:
## Description

Update the developer documentation for egress proxy setup and Tock deployment. This documents the egress proxy updates we've just applied in staging to restore New Relic traffic.

Also reduce the New Relic agent log level back to `info`.

Addresses #1792 .

## Testing

Once deployed, SSH into the tock-staging container and confirm that the New Relic agent is only logging at the info level and above:

```bash
tail /tmp/newrelic-python-agent.log
```

## Production deployment

With this update, we are clear to deploy the changes that have accumulated in staging over several weeks, including these latest changes for New Relic data.

That will consist of three steps:

1. Manually update and re-push the production egress proxy, following the [updated docs](https://github.com/18F/tock/compare/tock-1792-cleanup-and-docs?expand=1#diff-d4026c90b30e33741a829571bb9aa7944837415d6b3dbd4dff0b6242af3b1039), to add the `gov-collector.newrelic.com` domain the the egress proxy's allow list. This reflects the egress config that I've validated in staging.
2. Push up a tag for a release and auto-deploy the updated main branch to production.
3. Delete the production network policy for `production-egress` on port 8080.

Step 3 is a clean-up step that I've applied in staging; it closes an unused path for external network traffic and simplifies the configuration for future maintenance.

### Avoiding downtime

Our unexpected downtime in staging when we applied the New Relic egress related changes was due to a few factors:

1. Pushing an outdated version of the Caddy proxy application, which broke all external network traffic
2. A pre-existing bug in the `run.sh` script that caused application start-up to fail if the New Relic admin tool failed to record the deployment
3. Doing the code deployment first, followed by the egress proxy update without validation in between

Issue 2 was resolved by #1807 (although that fix won't be in effect until the release is deployed) and the other two issues will be mitigated by following the deployment steps in the order above with a pair / co-pilot, and with manual smoke-testing of Tock between each step.

### DIFF
--- a/docs/egress.md
+++ b/docs/egress.md
@@ -10,7 +10,8 @@ Accordingly, we have configured [a Caddy proxy](https://github.com/GSA-TTS/cg-eg
 
 - `uaa.fr.cloud.gov`: The [cloud.gov UAA server](https://cloud.gov/docs/management/leveraging-authentication/) which in turn uses GSA SecureAuth for authentication.
 - `google-analytics.com`: [DAP](https://digital.gov/guides/dap/), for web app analytics
-- `api.newrelic.com`: [New Relic endpoints](https://docs.newrelic.com/docs/apis/rest-api-v2/get-started/introduction-new-relic-rest-api-v2/) which is used for the `newrelic-admin` tool
+- `api.newrelic.com`: The [New Relic REST API endpoint](https://docs.newrelic.com/docs/apis/rest-api-v2/get-started/introduction-new-relic-rest-api-v2/) which is used by the `newrelic-admin` tool to record deployments
+- `gov-collector.newrelic.com`: The FedRAMP-compliant [New Relic APM collector endpoint](https://docs.newrelic.com/docs/security/security-privacy/compliance/fedramp-compliant-endpoints/#agents), used by the New Relic Python agent
 
 ## A note about cloud.gov egress and spaces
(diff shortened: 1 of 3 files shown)


docs/egress.md
```diff
@@ -10,39 +10,94 @@ Accordingly, we have configured [a Caddy proxy](https://github.com/GSA-TTS/cg-eg
 
 - `uaa.fr.cloud.gov`: The [cloud.gov UAA server](https://cloud.gov/docs/management/leveraging-authentication/) which in turn uses GSA SecureAuth for authentication.
 - `google-analytics.com`: [DAP](https://digital.gov/guides/dap/), for web app analytics
```
cantsin, line 12: Let's remove `google-analytics`?

cantsin (approved):
LGTM! love the docs update. one lingering question about GA -- do we really need it there?

docs/egress.md
nateborr: I can try that change out in staging. I found a mention in Slack from April 2023 stating that we need it, but I don't know what that was based on.

I would like to make that change separately from these ones as a follow-up. It'd be convenient to make all the egress config changes together, but I don't want to add to the surface area of this update in case we do run into additional issues with the prod deploy.

PR #1816, opened 2024-10-17
Require cg-django-uaa 2.1.5 and Django 4.2

nateborr:
## Description

Bump the `cg-django-uaa` library version to 2.1.5 since that version is [required](https://github.com/cloud-gov/django-uaa/blob/main/CHANGELOG.md#215---2024-01-26) for compatibility with Django 4.x. This should remediate installation errors occurring in staging since #1809 was merged.

Additionally update the Pipfile to require Django 4.x at or above 4.2, since 3.x will no longer be compatible with `cg-django-uaa`. (This change doesn't update the `django` version in the lockfile.)

## Additional information

I've shaken this out locally without issue. I'd like to confirm the PR build and test suite succeeds, get it into staging, and confirm a successful build there.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -4,7 +4,7 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-django = ">=3.2.19"
+django = "~=4.2"
 dj-database-url = "*"
 gevent = "*"
 gunicorn = "*"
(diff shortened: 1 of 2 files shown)


edwintorres (approved):
LGTM

nateborr:
I'm working on identifying what I'm doing wrong to update the `Pipfile.lock` file here.

PR #1829, opened 2024-11-07
Show both 18F and CoE projects on timecard entry page

neilmb:
## Description

https://github.com/18F/TLC-crew/issues/1002 asks for a solution for 18F and CoE staff to tock to the same project. As discussed there, the only fix needed is to include the other orgs projects on the timecard entry page. This least-effort solution special-cases that filtering logic for just those two organizations so that they will see each other's projects.

There was a test case for this and I've updated it to reflect the new special-case behavior.


cantsin (approved):
LGTM -- thank you good sir!

